PR #60, opened 2018-01-12
MODINVSTOR-52: Replace DatabaseExceptionUtils by RMB's PgExceptionUtil

julianladisch:


julianladisch:
There isn't a 1:1 relationship of Postgres and Mysql/MariaDB error codes: https://mariadb.com/kb/en/library/mariadb-error-codes/ Therefore a generic DatabaseExceptionUtils is much more than an abstract class.

PR #63, opened 2018-01-18
MODINVSTOR-26: Extend Instances to include MARC payload.

julianladisch:


julianladisch:
We decided not to merge this code because we may come up with a better solution in https://issues.folio.org/browse/MODINVSTOR-26

PR #74, opened 2018-03-08
MODINVSTOR-62: Increase pageable's maximum limit and offset

julianladisch:
by updating ramls/raml-util to latest raml submodule


marcjohnson-kint:
@julianladisch I've merge in master and updated the implementation and interface versions to reflect this change.

PR #86, opened 2018-03-23
MODINVSTOR-71: Added metadata to holdings, items, loan types and material types

doytch:
[STCOM-228](https://issues.folio.org/browse/STCOM-228), [UIIN-98](https://issues.folio.org/browse/UIIN-98) and [UIIN-99](https://issues.folio.org/browse/UIIN-98) need update `metadata` added to the payloads they fetch. There will need to be one more PR against `mod-inventory` to add metadata to the `/inventory/items` endpoint.

I'm using PR #53 as my guide with this implementation but let me know if there's anything I could be doing better here, please. 

I also bumped minor version numbers on the APIs since we're adding functionality (but mostly bc @marc-johnson did it in that PR 😝).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ## 8.2.3 Unreleased
 * Removing SQ warnings and improving test coverage in the new locations and location-units (MODINVSTOR-89)
+* Adds metadata generation (dates and update user) to item, holding, material type and loan type records (MODINVSTOR-71)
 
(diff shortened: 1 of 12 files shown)


marcjohnson-kint:
@nielserik Indeed, I have changed the implementation version to 8.3.0 to reflect the additional behaviour. And I've changed the versions in the module descriptor template to reflect the changes in the interface definitions.

PR #96, opened 2018-04-10
Replace invalid V4 UUIDs in reference and sample data.

wafschneider:
Towards CIRC-97.


wafschneider:
Apologies, I didn't notice that @julianladisch had already submitted a more comprehesive PR.

PR #99, opened 2018-04-10
Modinvstor 92 deletechecks

heikkilevanto:
Second take on the delete checks. Foreign keys for locations in items and holdings, SQ warnings, refactored some of the helpers into a helper class.


heikkilevanto:
Fixed the SQ warnings and pushed.

PR #134, opened 2018-07-26
MODINVSTOR-142: JSON Schema for MARC 'fields' array

julianladisch:


julianladisch:
This fails because RMB creates an empty Field class that may fail, see discussion on https://issues.folio.org/issues/MODINVSTOR-142

PR #142, opened 2018-08-09
MODINVSTOR-144: PUT marc-json should also update existing marc-json

julianladisch:


src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
```diff
@@ -597,18 +597,15 @@ public void putInstanceStorageInstancesByInstanceIdSourceRecordMarcJson(
...
+    postgresClient.upsert(INSTANCE_SOURCE_MARC_TABLE, instanceId, entity, reply -> {
       if (reply.succeeded()) {
         asyncResultHandler.handle(Future.succeededFuture(
             PutInstanceStorageInstancesByInstanceIdSourceRecordMarcJsonResponse.withNoContent()));
         return;
       }
       Map<Object, String> fields = PgExceptionUtil.getBadRequestFields(reply.cause());
-      if (fields != null && "23503".equals(fields.get('C'))) {  // foreign key constraint violation
+      if (fields != null && "23503".equals(fields.get('C'))  // foreign key constraint violation
+          && INSTANCE_SOURCE_MARC_TABLE.equals(fields.get('t'))) {
```
marcjohnson-kint: Is this the same check for all foreign key violations or specific to the foreign key from MARC source records to instances?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -562,9 +577,52 @@ public void canCreateInstanceSourceRecord() throws Exception {
...
+
+    put(id, toMarcJson("/101073931X.mrcjson"));
+
+    Response getResponse = getMarcJson(id);
+    assertThat(getResponse.getStatusCode(), is(200));
+    JsonArray fields = getResponse.getJson().getJsonArray("fields");
+    assertThat(fields.getJsonObject(0).getString("001"), is("101073931X"));
+  }
+
+  @Test  // https://issues.folio.org/browse/MODINVSTOR-143?focusedCommentId=33618#comment-33618
```
marcjohnson-kint: This comment suggests that the change to use the PotgresClient upsert also resolved the single quote character error for one of the samples. Is that the case?

marcjohnson-kint: I've verified this locally and have included the last sample file in the set of sample records to import

PR #159, opened 2018-09-14
MODINVSTOR-172: UUID syntax check in InstanceStorageAPI, HoldingsStorageAPI

julianladisch:


julianladisch:
Draft to be discussed at https://issues.folio.org/browse/MODINVSTOR-172

NEWS.md
```diff
@@ -1,3 +1,5 @@
+* Enable UUID syntax check for POST instance and POST holding (MODINVSTOR-172)
```
marcjohnson-kint: Please add a version header (12.8.2 Unreleased, in my opinion) to this news entry

julianladisch: I've added that version header, but I don't know whether the next release will be 12.8.2, 12.9.0 or 13.0.0.

marcjohnson-kint: That is ok, if a subsequent change is merged which means the next release should be 12.9.0, that merge should change the version header.

The policies for when version numbers are going to be allocated and the increment following a release are both being evaluated at the moment, after Tuesday's meeting (see https://issues.folio.org/browse/FOLIO-1482)

PR #160, opened 2018-09-15
MODINVSTOR-174: handleCQL: fix tableName initialisation, code cleanup

julianladisch:


julianladisch:
Good catch, the code contains a thread safety / concurrency issue with tableName.

Using `synchronized` doesn't help because the result handlers get called asynchronously for example in putInstanceStorageInstancesByInstanceId.

Currently a new instance of InstanceStorageAPI is created for each request but this may change.

Using an inner class that contains both tableName and CQLWrapper is a good solution. However we do not really need it because that code will be removed by the other pull request: https://github.com/folio-org/mod-inventory-storage/pull/163

julianladisch:
I approve the commit "Added inner-class to help with thread safety 40a5a87" (cannot use github's approve gui because i created this pull request).

PR #167, opened 2018-09-27
Modinvstor 179 document json schema properties

nielserik:


adamdickmeiss:
Squash and merge this one so we don't get history with things that don't pass

PR #215, opened 2019-01-02
Modinvstor 216 load reference data

adamdickmeiss:


adamdickmeiss:
Since there are bad commits in this, it should be "Squash and merge".

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -0,0 +1,199 @@
...
+import org.folio.rest.jaxrs.model.Parameter;
+import org.folio.rest.jaxrs.model.TenantAttributes;
+
+public class TenantRefAPI extends TenantAPI {
+
+  private static final Logger log = LoggerFactory.getLogger(TenantRefAPI.class);
+
+  private HttpClient httpClient;
+
+  private static List<InputStream> getStramsfromClassPathDir(String directoryName) throws URISyntaxException, IOException {
```
craigmcnally: You have a typo here... "Strams"

PR #224, opened 2019-01-18
MODINVSTOR-215: Use sort or search index depending on the result size.

julianladisch:


julianladisch:
@adamdickmeiss You wrote:
> re-parses of SQL. Why not have an SQL query object
Can you elaborate?

adamdickmeiss:
Thx. I can see in this case it's broken parsing of CQL. Using regular expressions lookup or string lookup's such as query.contains("holdingsRecords.") or try to inspect sorting is just plain broken and will lead to all kinds of bad things when some day people do use those names as terms.. Example: https://github.com/folio-org/mod-codex-mux/blob/master/src/main/java/org/folio/codex/Multiplexer.java#L347

julianladisch:
> query.contains("holdingsRecords.")

This is not part of this pull request.

adamdickmeiss:
> > query.contains("holdingsRecords.")
> 
> This is not part of this pull request.

Indeed. But it is the style that is used, so I understand why that way of "accessing" CQL is used.

PR #225, opened 2019-01-18
MODINVSTOR-254: Performance: PUT /instance-storage/instances/{instId}

julianladisch:
* Speed up PUT /instance-storage/instances/{instanceId} by using PgUtil.put.
* Fix 404 for PUT /instance-storage/instances/{instanceId}.
* Remove duplicated code for deleting source record by using PgUtil.deleteById.
* Update RMB to v23.4.0, needed for PgUtil.put.
* Deprecate StorageHelper methods that have been moved to RMB PgUtil.


julianladisch:
The only reason why sonarlint checks fail is that I deprecated some methods. This should not stop the merge.

julianladisch:
The only reason why sonarlint checks fail is that I deprecated some methods. These warnings should no be suppressed until that code gets removed.

PR #227, opened 2019-01-22
Modinvstor 248 remove item material view

evaluk:
I have run the performance jmeter script and the mod-inventory-storage api postman scripts against this module after the update and they succeeded.  I am seeing an increase in performance and complete removal of the mt_items_view queries related to query by ID on the high usage query list.


pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-inventory-storage</artifactId>
   <groupId>org.folio</groupId>
-  <version>14.1.0-SNAPSHOT</version>
+  <version>15.0.0-SNAPSHOT</version>
```
julianladisch: We increment the version in a separate pull request when doing the release: https://dev.folio.org/guidelines/release-procedures/
Incrementing the version in a feature branch often results in a merge conflicts.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -89,12 +54,13 @@ public void getItemStorageItems(
...
           String[] fieldList = {"*"};
+          
+          CQL2PgJSON cql2pgJson = new CQL2PgJSON("item.jsonb");
+          CQLWrapper cql = new CQLWrapper(cql2pgJson, query)
+            .setLimit(new Limit(limit))
+            .setOffset(new Offset(offset));
 
-          CQLWrapper cql = getCQL(query,limit,offset);
-
-          log.info(String.format("SQL generated by CQL query (%s): %s", query, cql.toString()));
```
julianladisch: Please log the SQL so that we can see if the generated SQL is ok.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -89,12 +54,13 @@ public void getItemStorageItems(
           PostgresClient postgresClient = StorageHelper.postgresClient(vertxContext, okapiHeaders);
 
           String[] fieldList = {"*"};
+
```
julianladisch: Can you configure your editor to trim trailing whitespace like these spaces in the empty line?
https://dev.folio.org/guides/developer-setup/#coding-style

PR #228, opened 2019-01-23
Modinvstor 251

mtraneis:
Created a new time-period.json that's stored under /ramls. 
Added holdShelfExpiryPeriod  to service-points.raml 
Added holdShelfExpiryPeriod and validation logic with respect to pickupLocation. 
Added new and modified existing tests.


ramls/time-period.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
```
mreno-EBSCO: I think raml-cop will complain if there is not a description here. Yup, just looked at the raml-cop comment in the PR and saw this: `ERROR: lint-raml-cop: time-period.json: Missing top-level "description".`

src/main/java/org/folio/rest/impl/ServicePointAPI.java
```diff
@@ -34,6 +35,8 @@
   public static final String SERVICE_POINT_TABLE = "service_point";
   public static final String LOCATION_PREFIX = "/service-points/";
   public static final String ID_FIELD = "'id'";
+  public static final String SERVICE_POINT_CREATE_ERR_MSG_WITHOUT_HOLD_EXPIRY = "Hold shelf expiry period must be specified when service point can be used for pickup.";
+  public static final String SERVICE_POINT_CREATE_ERR_MSG_WITHOUT_BEING_PICKUP_LOC= "Hold shelf expiry period cannot be specified when service point cannot be used for pickup";
```
mreno-EBSCO: Should these be translatable? I see we use `Messages` from RMB in other APIs. I also see we don't use it in some APIs, so it isn't clear what our i18n policy is here. Also, there is a space missing here: "SERVICE...LOC="

src/main/java/org/folio/rest/impl/ServicePointAPI.java
```diff
@@ -326,6 +349,23 @@ public void putServicePointsByServicepointId(String servicepointId,
     });
   }
 
+  public void putServicePoints(Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext){
```
mreno-EBSCO: Uh... What's this all about? :) Can we remove this? I don't see how the RAML would generate a PUT for this resource. If it does, we should probably fix it so it doesn't. I'm not sure what PUT without an ID means in REST.

src/main/java/org/folio/rest/impl/ServicePointAPI.java
```diff
@@ -156,6 +159,16 @@ public void postServicePoints(String lang, Servicepoint entity,
           Context vertxContext) {
     vertxContext.runOnContext(v -> {
       try {
+
+        String validateSvcptResult = validateServicePoint(entity) ;
+        if (validateSvcptResult != null && !validateSvcptResult.isEmpty()){
```
julianladisch: `&& !validateSvcptResult.isEmpty()` is not needed and can be removed.

src/main/java/org/folio/rest/impl/ServicePointAPI.java
```diff
@@ -156,6 +159,16 @@ public void postServicePoints(String lang, Servicepoint entity,
...
+          asyncResultHandler.handle(Future.succeededFuture(
+            PostServicePointsResponse.respond422WithApplicationJson(
+              ValidationHelper.createValidationErrorMessage("name",
+                entity.getName(), validateSvcptResult))));
```
julianladisch: Can you move `ValidationHelper.createValidationErrorMessage("name", entity.getName(), validateSvcptResult)` into validateServicePoint?

src/main/java/org/folio/rest/impl/ServicePointAPI.java
mtraneis: Removed

src/main/java/org/folio/rest/impl/ServicePointAPI.java
mtraneis: The only gain by doing this is hiding away a few lines of codes for reusing purposes. However, would it be clearer to leave them there to be consistent with the overall pattern in the class?

src/main/java/org/folio/rest/impl/ServicePointAPI.java
marcjohnson-kint: We do not currently support translations in the backend. There was some early experimental support in RAML Module Builder, however it is rarely used now.

There are a few examples of validation messages with `code` properties that the UI could use for translations. However, I'm deferring specifically introducing those until we have a decision that this is the agreed approach, see https://issues.folio.org/browse/FOLIO-1716

PR #233, opened 2019-01-29
Update reference data when upgrading MODINVSTOR-263

adamdickmeiss:


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -152,6 +153,18 @@ private void loadRef(Map<String, String> headers, Iterator<String> it, Handler<A
     }
   }
 
+  private void applyHeaders(HttpClientRequest req, Map<String, String> headers, String json) {
```
julianladisch: This does not add all headers but only x headers.
And it invokes end.
I would have named it `endWithXHeaders`.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -181,22 +194,29 @@ private void loadRef(Map<String, String> headers, String endPoint, Handler<Async
...
+            } else {
+              f.handle(Future.failedFuture("POST " + endPointUrl + " returned status " + resPost.statusCode()));
+            }
+          });
+          applyHeaders(reqPost, headers, json);
+        } else if (resPut.statusCode() == 200) {
           f.handle(Future.succeededFuture());
         } else {
-          f.handle(Future.failedFuture("POST " + endPointUrl + " returned status " + x.statusCode()));
+          f.handle(Future.failedFuture("PUT " + endPointUrl + "/" + id + " returned status " + resPut.statusCode()));
```
julianladisch: Using guard clauses can reduce the nesting making it more easy to read and understand:

        if (resPut.statusCode() == 200) {
          f.handle(Future.succeededFuture());
          return;
        }
        if (resPut.statusCode() != 404) {
          f.handle(Future.failedFuture("PUT " + endPointUrl + "/" + id + " returned status " + resPut.statusCode()));
          return;
        }
        // resPut.statusCode() == 404
        HttpClientRequest reqPost = httpClient.postAbs(endPointUrl, resPost -> {
          if (resPost.statusCode() == 201) {
            f.handle(Future.succeededFuture());
          } else {
            f.handle(Future.failedFuture("POST " + endPointUrl + " returned status " + resPost.statusCode()));
          }
        });
        applyHeaders(reqPost, headers, json);

adamdickmeiss: I see, but it would break.. The f.handle stuff is collected and a return would therefore not handle "res" as it should. Your comment revealed a mistake that I did however :-)

PR #234, opened 2019-01-29
MODINVSTOR-215: Fix sorting of canGetAllInstances

julianladisch:
canGetAllInstances unit test may fail because the database returns the records unsorted.


julianladisch:
Marc did the code review on slack and accepts the pull request.

PR #254, opened 2019-03-06
MODINVSTOR-269: barcode==123 does not perform

hjiebsco:
Adjust barcode index so SQL generated by {{query=(barcode==123)}} can leverage. Remove lower/accent should be fine since we have full text index (and gin index) on barcode field as well. Performance test (50 requests in 50 seconds) shows that the change can improve {{query-by-barcode}} (~8s to ~100ms) and related parent APIs ({{checkin/out-by-barcode}} for example, ~25s to ~1s) significantly.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -810,8 +810,8 @@
         {
           "fieldName": "barcode",
           "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
+          "caseSensitive": true,
+          "removeAccents": false
         },
         {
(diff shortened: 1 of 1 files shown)


julianladisch:
There also exists the "fullTextIndex" "barcode" (line 837). A fullTextIndex always has these implicit properties:

    "caseSensitive": false, 
    "removeAccents": true

This is inconsistent with the proposed `"caseSensitive": true` for "index" "barcode" (line 793) in this pull request .

hjiebsco:
As discussed on slack, making suggested gin index change to be consistent. As for keeping gin or ft, that will be a different issue.

PR #255, opened 2019-03-07
MODINVSTOR-269: barcode search: drop fulltext index, use item.json schema.

julianladisch:
For barcode==123 (exact match) and barcode==123* (right truncation) we do
not need the fulltext index, the standard index (b-tree) is sufficient.

Using the item.json schema prevents trying to use a numeric match for barcode.

### DIFF
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -7,7 +7,6 @@ import io.vertx.core.Handler;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


julianladisch:
https://github.com/folio-org/raml-module-builder/blob/v23.9.0/README.md#the-post-tenant-api 
says that the defaults caseSensitive=false and removeAccents=true cannot be changed: 

* the caseSensitive allows you to create case insensitive indexes (boolean true / false), if you have a string field that may have different casings and you want the value to be unique no matter the case. Default: false should not be changed, temporarily all indexes are created with a lower() function wrapper. Index will not be used if this is changed (important on large tables)
* removeAccents - normalize accents or leave accented chars as is. Default: true should not be changed, temporarily all indexes are created with a f_unaccent() function wrapper. Index will not be used if this is changed (important on large tables)

marcjohnson-kint:
@julianladisch Is this still relevant, or can it be closed?

julianladisch:
See https://github.com/folio-org/mod-inventory-storage/pull/254

PR #267, opened 2019-04-25
Modinvstor 278 update to rmb 24

adamdickmeiss:


src/main/java/org/folio/rest/impl/LocationAPI.java
```diff
@@ -177,7 +176,7 @@ public void getLocationsById(
     Context vertxContext) {
 
     String tenantId = getTenant(okapiHeaders);
-    Criterion criterion = idCriterion(id, LOCATION_SCHEMA_PATH, asyncResultHandler);
+    Criterion criterion = idCriterion(id, asyncResultHandler);
```
marcjohnson-kint: What is the impact of not using the schema, does it change any validation that is done?

src/main/java/org/folio/rest/impl/StorageHelper.java
```diff
@@ -420,9 +420,9 @@ protected static String getTenant(Map<String, String> headers) {
     return TenantTool.calculateTenantId(headers.get(RestVerticle.OKAPI_HEADER_TENANT));
   }
 
-  protected static Criterion idCriterion(String id, String schemaPath, Handler<AsyncResult<Response>> asyncResultHandler) {
+  protected static Criterion idCriterion(String id, Handler<AsyncResult<Response>> asyncResultHandler) {
     try {
-      Criteria criteria = new Criteria(schemaPath);
```
marcjohnson-kint: What is the impact of not using the schema? Does it change any validation that is done? Or is it only accommodating a change in the underlying framework that doesn't make a difference in this circumstance?

julianladisch: Doesn't make a difference because mod-inventory-storage doesn't have numeric fields. https://issues.folio.org/browse/CQLPG-47

PR #268, opened 2019-04-30
Modinvstor 279

mrdpopov:
Add an electronic location & service point to the reference data


reference-data/service-points/cd3.json
```diff
@@ -0,0 +1,8 @@
+{
```
marcjohnson-kint: I think the file should be named after the name of the service point, e.g. `online.json` rather than `cd3.json`

mrdpopov: I was thinking of that too, I thought it would be better for consistency to  follow the naming pattern of other two files in the service-points directory: cd1 and cd2, Circ Desk 1 and Circ Desk 2, respectively.

mrdpopov: File has been renamed.

marcjohnson-kint: I would prefer more descriptive file names over less. 

I agree that we want consistency. I think we should rename the existing files to something more descriptive e.g. `circulation-desk-1`

PR #277, opened 2019-06-07
MODINVSTOR-285: Upgrade to RMB 26.0.0

julianladisch:


src/main/java/org/folio/rest/impl/InstanceNoteTypeAPI.java
```diff
@@ -46,10 +45,9 @@
   private static final String LOCATION_PREFIX = "/instance-note-types/";
   private static final Logger log             = LoggerFactory.getLogger(InstanceNoteTypeAPI.class);
   private final Messages messages             = Messages.getInstance();
-  private String idFieldName                  = "_id";
 
   public InstanceNoteTypeAPI(Vertx vertx, String tenantId) {
-    PostgresClient.getInstance(vertx, tenantId).setIdField(idFieldName);
+    PostgresClient.getInstance(vertx, tenantId);
   }
```
julianladisch: The constructor should be completely removed, there is no need to create a throw-away PostgresClient instance.

adamdickmeiss: OK.. I'll make a PR that change. Wonder if there others like that..

PR #279, opened 2019-06-24
Remove null from reference and sample data

hjiebsco:
While investigating mod-inventory-storage API test failures, it was noticed that 
- two sample data records use "null" values that break JSON schema validation. 
-- null statisticalCodeId. See https://github.com/folio-org/mod-inventory-storage/blob/v15.5.1/sample-data/holdingsrecords/aba-holdingsrecord3.json#L29
-- null notes. See https://github.com/folio-org/mod-inventory-storage/blob/v15.5.1/sample-data/instances/american-journal-of-medicine.json#L44
- one reference data uses "null" as instance status code.
-- null instance status code. See https://github.com/folio-org/mod-inventory-storage/blob/v15.5.1/reference-data/instance-statuses/notyetassigned.json#L3

This PR **removed** null values from sample data and **changed "null" to "none"** in reference data.

### DIFF
--- a/reference-data/instance-statuses/notyetassigned.json
+++ b/reference-data/instance-statuses/notyetassigned.json
@@ -1,6 +1,6 @@
 {
   "id": "f5cc2ab6-bb92-4cab-b83f-5a3d09261a41",
-  "code": "null",
+  "code": "none",
   "name": "Not yet assigned",
   "source": "folio"
 }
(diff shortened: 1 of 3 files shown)


marcjohnson-kint:
> While investigating mod-inventory-storage API test failures, it was noticed that two sample data records use "null" values that break JSON schema validation.

Where was it observed that these are invalid according to the schema?

hjiebsco:
The JSON schema validation is reported in folio-api-tests https://github.com/folio-org/folio-api-tests

PR #281, opened 2019-07-03
Define initial set of code owners

marcjohnson-kint:
To help broaden the collective ownership of this module and expand the set of folks who can review changes to this module we are introducing an initial set of code owners.

This initial set is made up of the active contributors to this module over the last year and the Tech Leads from teams actively working in this area.


marcjohnson-kint:
> Can the format be multi-line? Git and I work on lines - rather than words in a line ..

@adamdickmeiss I would prefer line per user as well. The [documentation|https://help.github.com/en/articles/about-code-owners] only includes examples with a single line, and a quick search didn't present any evidence that this was allowed.

There is no easy way to test this, so I have been cautious about experimenting

PR #287, opened 2019-07-08
Array array modifiers for contributors and identifiers MODINVSTOR-311

adamdickmeiss:


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -941,6 +941,31 @@ public void canSearchForInstancesUsingSimilarQueryToUILookAheadSearch() {
...
+  }
+
+  @Test
+  public void arrayModifierfsIdentifiers2() {
+    canSort("identifiers = /@identifierTypeId = isbn 9781447294146", "Uprooted");
+  }
+
+  @Test
+  public void arrayModifierfsIdentifiers3() {
+    canSort("identifiers = @identifierTypeId asin", "Nod");
```
hjiebsco: No slash before @ like other test cases?

hjiebsco: Also there is no "=" before "asin". Does the new syntax support that?

adamdickmeiss: The omitted slash was a mistake.. But it shows that full text search in many cases.. (key+values in sequence gets a hit)

adamdickmeiss: It's simply a search identifiers = "@identifierTypeId asin" .. @ most likely will be ŕemoved.. ANd "identitifierTypeId asin" is left with two tokens in order .. (phrase).

PR #289, opened 2019-07-09
[MODINVSTOR-310] Add date and source fields to the circulation notes

opochepam:
## Purpose
Resolves: MODINVSTOR-310
Add date and source fields to the circulation notes
## Links
https://issues.folio.org/browse/MODINVSTOR-310
https://github.com/folio-org/mod-inventory/pull/125


ramls/item.json
```diff
@@ -128,6 +132,34 @@
             "type": "string",
             "description": "Text to display"
           },
+          "source": {
+            "type": "object",
+            "description": "The user who added the note",
```
marcjohnson-kint: Is this always the user who added the note, or does it change when the note is updated (to the user who last updated the note)?

ramls/item.json
```diff
@@ -128,6 +132,34 @@
...
+            "properties": {
```
marcjohnson-kint: As these properties are stored copies of information from the user (presumably to support sorting), they should be marked read only, as this behaviour is controlled by the server.

marcjohnson-kint: As these properties are intended to be owned by the server, I think it could be useful to reflect that in the description

opochepam: Updated descriptions for the user and the date.

ramls/item.json
```diff
@@ -128,6 +132,34 @@
             "type": "string",
             "description": "Text to display"
           },
+          "source": {
+            "type": "object",
+            "description": "The user who added/updated the note. The property is generated by the server and needed to support sorting.",
+            "properties": {
+              "id": {
+                "type": "string",
+                "description": "The id of the user who added/updated the note"
```
julianladisch: Does this point to /users/{id}, if yes, then it should be mentioned in the description.

ramls/item.json
```diff
@@ -128,6 +132,34 @@
             "type": "string",
             "description": "Text to display"
           },
+          "source": {
+            "type": "object",
+            "description": "The user who added/updated the note. The property is generated by the server and needed to support sorting.",
+            "properties": {
+              "id": {
```
julianladisch: This is the 3rd "id" field in this file. Can you add a description to the "id" field in line 6 to explain the difference?

ramls/item.json
opochepam: Yes it does. Updated the description.

ramls/item.json
opochepam: Added the description.

ramls/item.json
julianladisch: Thanks for adding it to the description of "source" in line 138, can you also add it to the "id" description in line 142?

opochepam: done

PR #295, opened 2019-07-12
MODINVSTOR-291 Created endpoint for batch save Instances

KaterynaSenchenko:
Batch POST endpoint for saving Instances in inventory-storage was implemented to improve [data-import performance](https://wiki.folio.org/display/FOLIJET/data-import+performance). It will be refactored to comply with standard design for batch endpoints when it is defined ([FOLIO-2050](https://issues.folio.org/browse/FOLIO-2050)).

_**Design**_
_Method_: POST
_Resource_: /instance-storage/batch/instances (should be a separate interface "id": "instance-storage-batch")
_Body_: collection of Instances and total number of Instances
_Returns_: collection of successfully created Instances, error messages for failure Instances, total number of created Instances:

- If at least one Instance is successfully saved - returns 201 response with saved instances ("instances" section), "errorMessages" array contains errors for failed Instances (empty if all Instances are successfully saved).

- If save for all Instances failed - returns 500 response with "errorMessages" array, explaining a reason of why Instances were processed with failures (one error message per one Instance). Instances array is empty.


OleksiiKuzminov:
Plz, update readme and news. It would be nice to add a description of how partial batch save works

KaterynaSenchenko:
Hi @julianladisch, thank you for your review. I'm still working on the necessary changes and just want to clarify a few things:
1. Regarding partial success response status - I do agree that 500 is not the right solution, it was a temporary measure to indicate that not all of the records were saved that we applied in SRS. However, partial success implies success and we have a [PoC for 207 status](https://github.com/folio-org/mod-source-record-storage/pull/71) (ongoing discussion is in [FOLIO-2050](https://issues.folio.org/browse/FOLIO-2050)). We will definitely change this implementation once the design for batch endpoints is defined. At this time, do you think it would be more clear to use
201 status for partial success, not error status?
2. As for the number of connections to the db - is there a way to get the actual max_connections parameter (is it specified in postgres.config somewhere)? Or should we just assume that it's value is set to 100?

KaterynaSenchenko:
Hi @julianladisch, I think I addressed all your concerns, could you please take another look at the PR? Also, we decided to use 201 response status for partial success for now. Thanks.

KaterynaSenchenko:
Just learned that Julian is on vacation, @marcjohnson-kint, could you please review this PR?

KaterynaSenchenko:
Hi @kostyantyn-kh, @mrdpopov, @nielserik, @piotr-kalashuk, could you please review the PR?

KaterynaSenchenko:
Since it is an "experimental" API (version 0.1) and we plan to change it anyway, we probably can move forward with this PR. @julianladisch, if you find that you still have some concerns we can create a separate JIRA issue and address them in another PR. @OleksiiKuzminov, please merge this PR if you consider it is ready.

PR #300, opened 2019-07-26
Changed to mod-inv-store for new view query constructor

evaluk:


evaluk:
My changes no longer exercise all pathways through the PreparedCQL class init.  This class is used in other endpoints that dont seem to have unit or api level tests.  This is a completely new endpoint test which is out of scope of my user story.   Made this user story to address this.  
https://issues.folio.org/browse/MODINVSTOR-329

PR #301, opened 2019-07-28
MODINV-135 - Inventory App: Assign tags to instance record

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODINV-135 we would like to introduce tags for instance entities.  

## Approach
* updated instance model with tags

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -66,7 +66,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.0",
+      "version": "7.1",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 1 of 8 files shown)


natali-zaitseva:
Hi @marcjohnson-kint. I used old reference, currently, the issue is already moved to mod-inventory-storage --> https://issues.folio.org/browse/MODINVSTOR-323. I will update the branch to correspond to the MODINVSTOR-323. The changes are related to both mod-inventory-storage and mod-inventory modules. Should I create an additional issue for changes in another module? Thanks!

marcjohnson-kint:
> The changes are related to both mod-inventory-storage and mod-inventory modules. Should I create an additional issue for changes in another module?

Yes, (almost) all changes in a module need to be associated with an issue for that module.

For changes that span modules, an issue is needed for each module, in the relevant project.

PR #302, opened 2019-07-29
MODINVSTOR-323 - Inventory App: Assign tags to instance record

natali-zaitseva:
## Purpose
Due to https://issues.folio.org/browse/MODINV-135 we would like to introduce tags for instance entities.  

## Approach
* updated instance model with tags
* updated instance interfaces versions

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -1,8 +1,14 @@
 package org.folio.rest.api;
 
+import static java.net.HttpURLConnection.HTTP_OK;
+import static java.net.URLEncoder.encode;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.concurrent.TimeUnit.SECONDS;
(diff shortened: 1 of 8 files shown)


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -505,20 +512,20 @@ public void canProvideLargePageOffsetAndLimit()
...
 
     JsonObject page = pageResponse.getJson();
 
-    JsonArray instances = page.getJsonArray("instances");
+    JsonArray instances = page.getJsonArray(INSTANCES_KEY);
 
     assertThat(instances.size(), is(0));
     // Reports 0, not sure if this is to due with record count approximation
-    //assertThat(page.getInteger("totalRecords"), is(5));
+    //assertThat(page.getInteger(TOTAL_RECORDS_KEY), is(5));
```
OleksiiKuzminov: delete commented code or fix plz

natali-zaitseva:
@marcjohnson-kint there is only one test has been added with the name shouldReturnInstanceWhenFilterByTags().

PR #304, opened 2019-07-31
MODINVSTOR-322 Inventory App: Assign tags to item record

stanislav-mureiko:
## Purpose
Due to https://issues.folio.org/browse/MODINVSTOR-322 we would like to introduce tags for item entities.

## Approach
* updated item model with tags
* updated item interfaces versions

### DIFF
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -51,7 +51,7 @@ import io.vertx.core.json.JsonObject;
 
 public class ItemStorageTest extends TestBaseWithInventoryUtil {
 
-  public static final String TAG_VALUE = "test-tag";
+  private static final String TAG_VALUE = "test-tag";
 
   private static String journalMaterialTypeID;
   private static String bookMaterialTypeID;
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -50,6 +50,9 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 
 public class ItemStorageTest extends TestBaseWithInventoryUtil {
+
+  public static final String TAG_VALUE = "test-tag";
+
   private static String journalMaterialTypeID;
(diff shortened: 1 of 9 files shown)


stanislav-mureiko:
This PR should be merged after https://github.com/folio-org/mod-inventory#302

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -777,6 +782,47 @@ public void canSearchForItemsByBarcode()
...
+    assertThat(searchResponse.getStatusCode(), is(200));
+
+    JsonObject searchBody = searchResponse.getJson();
+
+    JsonArray foundItems = searchBody.getJsonArray("items");
+
+    LinkedHashMap item = (LinkedHashMap) foundItems.getList().get(0);
+    LinkedHashMap<String, ArrayList<String>> itemTags = (LinkedHashMap<String, ArrayList<String>>) item.get("tags");
+
+    assertTrue(searchResponse.getBody().contains(tagValue));
```
kostyantyn-kh: Could you move asserts for _foundItems_, _searchResponse_ and _searchBody_ under item? to check, for example, items size before you retrieve first element from list

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -777,6 +782,47 @@ public void canSearchForItemsByBarcode()
       is("673274826203"));
   }
 
+  @Test
+  public void canSearchForItemsByTags()
```
marcjohnson-kint: Personally, I would also update the tests for creating and updating items to include tags as well.

(I think that is also relevant for the previous changes for instances, apologies that I didn't manage to provide that feedback during those pull requests)

src/test/java/org/folio/rest/api/ItemStorageTest.java
stanislav-mureiko: Thanks, I'll fix it.

src/test/java/org/folio/rest/api/ItemStorageTest.java
stanislav-mureiko: Thanks it really makes sense.

PR #305, opened 2019-08-01
Modinvstor 286

Anatolii-Starkov:


ramls/examples/itemdamagedstatus.json
```diff
@@ -0,0 +1,5 @@
+{
```
marcjohnson-kint: Personally, I might separate the words in the file name to make it easier to read

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -46,4 +68,83 @@ public static void testBaseAfterClass()
...
+   * It validates a response with expected status.
+   * If response has body in application/json format then it returns a JsonObject, otherwise null.
+   *
+   * @param url                an URL of endpoint
+   * @param method             a http method
+   * @param content            a body content (can be null)
+   * @param expectedStatusCode an expected status
+   * @return a response from an endpoint
+   */
+  protected JsonObject send(
```
marcjohnson-kint: There are already methods in the HttpClient class that make HTTP requests, that the existing tests use.

Why have these been introduced? Is there a capability missing that was needed?

src/main/java/org/folio/rest/impl/ItemDamagedStatusAPI.java
```diff
@@ -0,0 +1,298 @@
...
+  public void getItemDamagedStatusesById(
+    String id,
+    String lang,
+    Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
+
+    vertxContext.runOnContext(v -> {
+      try {
+        getItemDamagedStatus(id, okapiHeaders, vertxContext)
```
marcjohnson-kint: This approach is rather different to much of the existing code in the module, some of which has been updated to use the PgUtil class from RAML Module Builder.

Can you help me understand why this approach was chosen? Did it come from your work on another team?

Anatolii-Starkov: PgUtil uses the reflection API that is slow that's why I don't want to use it.
I don't think that this approach is completely different to existing code. I just move DB work out of controller to a method. 
Is that a problem? Shall I rewrite it using single method style?

src/test/java/org/folio/rest/api/TestBase.java
Anatolii-Starkov: HttpClient use callbacks and this method returns some result without callback. I found it in some test and move it to the TestBase. Honestly, there are a lot of such peaces of code ...

ramls/examples/itemdamagedstatus.json
Anatolii-Starkov: it's weird, I agree. But many of those files have the same style, so I decided that's a rule.

marcjohnson-kint: I don't think it is a rule.

I agree that it is valuable to be consistent. Which means we either
* start a new precedent during this pull request, and then issue another soon after to rename all of the other files
* leave it as it is, including these new files, during this pull request and change to a different convention in a separate pull request
* leave it as is

What do you think?

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: > HttpClient use callbacks and this method returns some result without callback. 

Yeah, that was likely written by me back when callbacks were the standard way to use vert.x in FOLIO. I have long since moved on to often use CompletableFutures (other folks use vert.x futures).

I'd be more than happy if you'd like to add future based variants of the methods in that class.

> I found it in some test and move it to the TestBase.

I don't think the TestBase should be responsible for knowing how to make HTTP requests to the module.

If there is a need for these new methods, I'd prefer they are extracted out of the base class.

(I'm generally conflicted by the base classes we use, because they have become places for putting lots of different responsibilities out of convenience, yet they are essential for supporting running the tests on their own)

> Honestly, there are a lot of such peaces of code ...

Yeah, there likely are, as different folks have made changes and written tests in different ways.

I don't like the `send` methods because I think they obscure the intent of the caller and they have more complex implementations, as they typically have branching to deal with the variety of ways they can be invoked.

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
```diff
@@ -1,5 +1,6 @@
 package org.folio.rest.support;
 
+//TODO remove it and use io.netty.handler.codec.http.HttpResponseStatus.UNPROCESSABLE_ENTITY
```
marcjohnson-kint: I agree, this constant should possibly be removed. 

The RAML Module Builder defines it's own set of [HTTP status codes](https://github.com/folio-org/raml-module-builder/blob/a058d1c5be1d70d014fb404cff7deff9dffda5e4/util/src/main/java/org/folio/HttpStatus.java)

Why are you suggesting that we use the netty defined HTTP status code constants instead?

src/main/java/org/folio/rest/impl/ItemDamagedStatusAPI.java
marcjohnson-kint: > PgUtil uses the reflection API that is slow that's why I don't want to use it.

I can understand those concerns. I imagine this is something you or colleagues discovered on another team. Have you or the other team, provided this feedback to the core platform team?

Where possible, I think it is better if we can improve the shared tooling (admittedly, mod-circulation is the biggest outlier in the project in this regard)

(Aside: A long time ago I tried to write a set of classes for making storage endpoints relatively easy to write. However, at the time I didn't feel I could achieve it whilst sticking with the idioms of RAML Module Builder. 

I subsequently abandoned my attempts as I worked more on business logic modules and the PgUtil was introduced.)

> I don't think that this approach is completely different to existing code.

I may well have missed that, I'm not too familiar with the storage modules anymore. Can you refer me to examples of existing code like this?

> Is that a problem? Shall I rewrite it using single method style?

At this point, my goal is to understand. It likely isn't my call as to whether to do that or not.

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
Anatolii-Starkov: Because 422 http status is called Unprocessable Entity but in RAML Module Builder set it's HTTP_VALIDATION_ERROR (with status code 422)

src/main/java/org/folio/rest/impl/ItemDamagedStatusAPI.java
Anatolii-Starkov: > I can understand those concerns. I imagine this is something you or colleagues discovered on another team. Have you or the other team, provided this feedback to the core platform team?

Do you ask me to say core platform guys that PgUtil is slow and they need to create a ticket to rewrite it without reflection API?

> Can you refer me to examples of existing code like this?

Here is similar code [DataImportImpl.java](https://github.com/folio-org/mod-data-import/blob/2d579b9b054fbbc0bc97f3e8c6a0b2b2e633490d/src/main/java/org/folio/rest/impl/DataImportImpl.java#L94)

src/test/java/org/folio/rest/api/TestBase.java
Anatolii-Starkov: > I'd be more than happy if you'd like to add future based variants of the methods in that class.

You got it. 

One more thing. Now code coverage is 72% and I can not cover my code more because I need to simulate exceptions of DB. Do you have any objections of using PowerMock for mocking of static methods?

ramls/examples/itemdamagedstatus.json
Anatolii-Starkov: I would crate a separate ticket in which scope will create a pull request
(Probably add something else TODO etc.)

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
Anatolii-Starkov: > Why are you suggesting that we use the netty defined HTTP status code constants instead?

[in  HTTP status codes](https://github.com/folio-org/raml-module-builder/blob/a058d1c5be1d70d014fb404cff7deff9dffda5e4/util/src/main/java/org/folio/HttpStatus.java#L84) it's called HTTP_VALIDATION_ERROR(422) but in the world It's well known as 422 Unprocessable Entity
It was confused me

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: > Now code coverage is 72% and I can not cover my code more because I need to simulate exceptions of DB. Do you have any objections of using PowerMock for mocking of static methods?

Testing exceptions thrown by methods provided by RAML Module Builder is a challenge, particularly within the storage module.

I'm not aware of a standard approach for this, and it is a general platform challenge. Given that, I'm not going to object to the use of PowerMock. Has this been used elsewhere in FOLIO?

ramls/examples/itemdamagedstatus.json
marcjohnson-kint: Ok, we shall leave it for this pull request

src/main/java/org/folio/rest/impl/ItemDamagedStatusAPI.java
marcjohnson-kint: > > I can understand those concerns. I imagine this is something you or colleagues discovered on another team. Have you or the other team, provided this feedback to the core platform team?
> 
> Do you ask me to say core platform guys that PgUtil is slow and they need to create a ticket to rewrite it without reflection API?
> 

I'm suggesting that if you, and others, are rejecting the use of PgUtil, then I think it could be worth giving feedback as to why to the core platform team.

Otherwise, we likely end up with another significantly different approach within the storage modules.

> > Can you refer me to examples of existing code like this?
> 
> Here is similar code [DataImportImpl.java](https://github.com/folio-org/mod-data-import/blob/2d579b9b054fbbc0bc97f3e8c6a0b2b2e633490d/src/main/java/org/folio/rest/impl/DataImportImpl.java#L94)

Thank you for the example. 

Do you know if that team chose to do this for the same reason?

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
marcjohnson-kint: > > Why are you suggesting that we use the netty defined HTTP status code constants instead?
> 
> [in HTTP status codes](https://github.com/folio-org/raml-module-builder/blob/a058d1c5be1d70d014fb404cff7deff9dffda5e4/util/src/main/java/org/folio/HttpStatus.java#L84) it's called HTTP_VALIDATION_ERROR(422) but in the world It's well known as 422 Unprocessable Entity
> It was confused me

Could it be preferably to raise a change for the RAML Module Builder to rename it, rather than create a need to introduce a new top level dependency purely for a single constant (that presumably is differently modelled).

src/main/java/org/folio/rest/impl/ItemDamagedStatusAPI.java
marcjohnson-kint: As discussed on Slack, this is a broader topic and should be discussed with a broader audience on a JIRA issue.

Given that, I'm shall leave it to other reviewers if they want to continue this topic within this pull request.

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(400, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void putItemDamagedStatusesByIdShouldInternalServerError(TestContext testContext) {
```
marcjohnson-kint: Under what circumstances should a PUT result in a server error?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(400, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void putItemDamagedStatusesByIdShouldReturnBadRequest(TestContext testContext) {
```
marcjohnson-kint: When would a PUT respond with a bad request?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(500, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void deleteItemDamagedStatusesByIdShouldReturnBadRequest(TestContext testContext) {
```
marcjohnson-kint: Under what circumstances does a delete respond with a bad request?

src/test/java/org/folio/rest/support/db/ErrorFactory.java
```diff
@@ -0,0 +1,240 @@
...
+import static org.folio.rest.support.db.ErrorConstants.CHILD_TABLE_NAME;
+import static org.folio.rest.support.db.ErrorConstants.SCHEMA_NAME;
+
+import scala.collection.immutable.Map;
+
+public class ErrorFactory {
+
+  private ErrorFactory() {}
+
+ public  static Map<Object, String> getForeignKeyErrorMap(){
```
marcjohnson-kint: Please put a space between the method parameters and the opening brace to be consistent with the code standards.

Other methods in this class are also affected, please correct those as well.

```suggestion
 public  static Map<Object, String> getForeignKeyErrorMap() {
```

src/test/java/org/folio/rest/support/db/ErrorBuilder.java
```diff
@@ -0,0 +1,104 @@
+package org.folio.rest.support.db;
+
+import scala.Tuple2;
+import scala.collection.immutable.Map;
+import scala.collection.immutable.Map$;
+
+public class ErrorBuilder {
```
marcjohnson-kint: What kind of error does this build?

Has this and the other general classes below been copied from another project?

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
Anatolii-Starkov: Here is the ticket https://issues.folio.org/browse/RMB-445

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: a developer mistake

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: wrong UUID

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: wrong UUID

src/test/java/org/folio/rest/support/db/ErrorFactory.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/support/db/ErrorBuilder.java
Anatolii-Starkov: You are right. I have copied then from another project

src/test/java/org/folio/rest/api/TestBase.java
Anatolii-Starkov: I have just extended the HttpClient with methods those return CompletableFuture and returned send method as was

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
marcjohnson-kint: In the meantime, please either remove or update the TODO to reflect the RAML Module Builder issue

src/test/java/org/folio/rest/support/HttpClient.java
```diff
@@ -1,46 +1,55 @@
...
-    if (tenantId != null) {
+  private void initDefaultHeaders(HttpClientRequest request, URL url, String tenantId) {
+    if (isNotBlank(tenantId)) {
       request.headers().add(TENANT_HEADER, tenantId);
     }
     if (url != null) {
-      request.headers().add("X-Okapi-Url", url.getProtocol() + "://" + url.getHost() + ":" + url.getPort());
-      request.headers().add("X-Okapi-Url-to", url.getProtocol() + "://" + url.getHost() + ":" + url.getPort());
+      String baseUrl = format("%s://%s", url.getProtocol(), url.getAuthority());
+      request.headers().add(X_OKAPI_URL, baseUrl);
```
marcjohnson-kint: What is the value of a constant when it is used in only a single place?

src/test/java/org/folio/rest/support/HttpClient.java
```diff
@@ -1,46 +1,55 @@
...
     } catch (MalformedURLException ex) {
-      log.warn("Malformed url:" + url + " message: " + ex.getMessage());
+      LOG.info(format("Malformed url: %s, message: %s", url, ex.getMessage()));
     }
-    stdHeaders(request, tmp, tenantId);
   }
 
-  private void stdHeaders(HttpClientRequest request, URL url, String tenantId) {
-    if (tenantId != null) {
+  private void initDefaultHeaders(HttpClientRequest request, URL url, String tenantId) {
```
marcjohnson-kint: Personally I prefer not to use abbreviations in methods names if possible.

```suggestion
  private void initializeDefaultHeaders(HttpClientRequest request, URL url, String tenantId) {
```

```suggestion
  private void addDefaultHeaders(HttpClientRequest request, URL url, String tenantId) {
```

src/test/java/org/folio/rest/support/HttpClient.java
```diff
@@ -50,17 +59,22 @@ public void post(
...
     HttpClientRequest request = client.postAbs(url.toString(), responseHandler);
-    request.headers().add("Content-type","application/json");
-    stdHeaders(request, url, tenantId);
+    request.putHeader(CONTENT_TYPE, APPLICATION_JSON);
+    initDefaultHeaders(request, url, tenantId);
 
     if (body == null) {
       request.end();
-      return;
+    } else {
```
marcjohnson-kint: To me, the null check is effectively a guard clause. Why does this second aspect needs to be wrapped in an else statement?

src/test/java/org/folio/rest/support/db/ErrorBuilder.java
marcjohnson-kint: Which project was it copied from?

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -49,16 +52,23 @@
...
   private static Vertx vertx;
   private static int port;
 
-  public static URL storageUrl(String path) throws MalformedURLException {
-    return new URL("http", "localhost", port, path);
+  public static URL storageUrl(String path) {
+    try {
+      return new URL("http", "localhost", port, path);
+    } catch (MalformedURLException e) {
+      throw new RuntimeException(e);
```
marcjohnson-kint: Why is this being converted to a runtime exception? Just so the checked exception can be avoided in calling methods?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
marcjohnson-kint: What kind of developer mistake, one that causes a specific kind of exception to occur?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
marcjohnson-kint: By wrong do you mean a UUID of a record that does not already exist? Or an invalid UUID?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
marcjohnson-kint: By wrong do you mean a UUID of a record that does not already exist? Or an invalid UUID?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(500, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void deleteItemDamagedStatusesByIdShouldReturnInternalServerError(TestContext testContext) {
```
marcjohnson-kint: Under what circumstances should it return an internal server error, when a specific exception is thrown?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(500, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void getItemDamagedStatusesByIdShouldReturnInternalServerError(TestContext testContext) {
```
marcjohnson-kint: Under what circumstances should it return an internal server error, when a specific exception is thrown?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(400, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void postItemDamagedStatusesShouldReturnInternalServerError(TestContext testContext) {
```
marcjohnson-kint: The test name does not describe why an internal server error should be returned. Under what circumstances should it return an internal server error, when a specific exception is thrown?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+      rule.vertx().getOrCreateContext()
+    );
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(500, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void postItemDamagedStatusesShouldReturnBadRequest(TestContext testContext) {
```
marcjohnson-kint: The test name does not describe why a bad request should be returned. Under what circumstances should a bad request response be returned?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+    );
+
+    responseFuture.setHandler(it -> {
+      testContext.assertEquals(400, it.result().getStatus());
+      async.complete();
+    });
+  }
+
+  @Test
+  public void getItemDamagedStatusesShouldReturnInternalServerError(TestContext testContext) {
```
marcjohnson-kint: The test name does not describe why an internal server error should be returned. Under what circumstances should it return an internal server error, when a specific exception is thrown?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
```diff
@@ -0,0 +1,333 @@
...
+
+
+  @Before
+  public void setUp(TestContext testContext) throws Exception {
+    MockitoAnnotations.initMocks(this);
+    when(pgClientFactory.getInstance(any(Context.class), any(Map.class))).thenReturn(postgresClient);
+  }
+
+  @Test
+  public void getItemDamagedStatusesShouldReturnBadRequest(TestContext testContext) {
```
marcjohnson-kint: The test name does not describe why a bad request should be returned. Under what circumstances should a bad request response be returned?

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+      })
+      .whenComplete((response, throwable) -> {
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_NO_CONTENT.toInt()));
+      });
+  }
+
+  @Test
+  public void cannotDeleteItemDamagedStatusThatDoesNotExist() {
+
```
marcjohnson-kint: I think this blank line can be removed, other new tests don't have a leading blank line

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+      })
+      .thenCompose(it -> client.get(itemDamagedStatusesUrl(EMPTY), TEST_TENANT))
+      .whenComplete((response, throwable) -> {
+        int size = response.getJson().getJsonArray("itemDamageStatuses").size();
+        assertThat(size, is(2));
+      });
+  }
+
+  @Test
+  public void shouldFindItemDamagedStatusesByName() {
```
marcjohnson-kint:
```suggestion
  public void shouldFindAnItemDamagedStatusByName() {
```

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+      })
+      .whenComplete((response, throwable) -> {
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_CREATED.toInt()));
+      })
+      .thenCompose(it -> {
+        URL url = itemDamagedStatusesUrl("?query=name=itemDamagedStatusName");
+        return client.get(url, TEST_TENANT);
+      })
+      .whenComplete((response, throwable) -> {
```
marcjohnson-kint: Is this chaining style common in the tests of other projects?

Personally I find it harder to read than sequential blocking of asynchronous requests.

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_BAD_REQUEST.toInt()));
+      });
+  }
+
+  @Test
+  public void canGetItemDamagedStatus() {
+
```
marcjohnson-kint: The other tests do not have a leading blank line, I believe this can be removed

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+    client.post(itemDamagedStatusesUrl(EMPTY), status, TEST_TENANT)
+      .whenComplete((response, throwable) -> {
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_BAD_REQUEST.toInt()));
+      });
+  }
+
+  @Test
+  public void shouldNotCreateItemDamagedStatusWithTheSameId() {
+
```
marcjohnson-kint: The other tests do not have a leading blank line, I believe this can be removed

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+    client.post(itemDamagedStatusesUrl(EMPTY), status, TEST_TENANT)
+      .whenComplete((response, throwable) -> {
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_BAD_REQUEST.toInt()));
+      });
+  }
+
+  @Test
+  public void shouldNotCreateItemDamagedStatusWithTheSameId() {
```
marcjohnson-kint:
```suggestion
  public void shouldNotBeAbleToCreateItemDamagedStatusWithTheSameId() {
```

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+    JsonObject object = new JsonObject().put("foo", "boo");
+    client.post(itemDamagedStatusesUrl(EMPTY), object, TEST_TENANT)
+      .whenComplete((response, throwable) -> {
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(422));
+      });
+  }
+
+  @Test
+  public void shouldNotCreateItemDamagedStatusWithTheSameName()
```
marcjohnson-kint:
```suggestion
  public void shouldNotBeAbleToCreateItemDamagedStatusWithTheSameName()
```

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+        assertThat(response, notNullValue());
+        assertThat(response.getStatusCode(), is(HttpStatus.HTTP_CREATED.toInt()));
+        JsonObject statusJson = response.getJson();
+        assertThat(statusJson.getString("id"), notNullValue());
+        assertThat(statusJson.getString("name"), is(status.getName()));
+      });
+  }
+
+  @Test
+  public void shouldNotCreateItemDamagedStatusWhenUnprocessableEntity() {
```
marcjohnson-kint: What kind of unprocessable entity?

The body of the test appears to describe when unexpected properties are included in the request, maybe the test could be renamed?
```suggestion
  public void shouldNotCreateItemDamagedStatusWhenUnexpectedPropertiesAreProvided() {
```

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+  public static final String TEST_TENANT = "test_tenant";
+
+  @Before
+  public void beforeEach() {
+    StorageTestSuite.deleteAll(TEST_TENANT, ItemDamagedStatusAPI.REFERENCE_TABLE);
+  }
+
+  @Test
+  public void canCreateItemDamagedStatus() {
+
```
marcjohnson-kint: The other tests do not have a leading blank line, I believe this can be removed

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
```diff
@@ -0,0 +1,247 @@
...
+  public static final String TEST_TENANT = "test_tenant";
+
+  @Before
+  public void beforeEach() {
+    StorageTestSuite.deleteAll(TEST_TENANT, ItemDamagedStatusAPI.REFERENCE_TABLE);
+  }
+
+  @Test
+  public void canCreateItemDamagedStatus() {
```
marcjohnson-kint:
```suggestion
  public void canCreateAnItemDamagedStatus() {
```

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: I don't think that it's common style. Shall rewrite it ?

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
marcjohnson-kint: I think it would be preferable for it to closer resemble the existing tests in the module, especially as it seems unlikely they will all be converted to use this style.

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: if you say so. The third time I will rewrite these test. I think it will be the last one

marcjohnson-kint: I'm not trying to force you to rewrite things multiple times. I'm trying to share my thoughts, ask questions and understand decisions.

What I don't understand with these tests is why you chose to undertake a different style to the existing tests? Do you find this style easier to read, you've used it elsewhere?

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: Thank you

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: the first call returns a future as you asked me to do for the HttpClient to avoid callbacks.
A future has appropriate methods to handle it
The main reason is to avid async method call on TestContext object.
All productive ready code has chain-style approach. Why don't it use in tests? 
I find it well-structured, easy read, async styled approach.

marcjohnson-kint: > the first call returns a future as you asked me to do for the HttpClient to avoid callbacks.

Sure, and I'm glad for that change. In retrospect, I would've preferred to use completable futures from the beginning (I was trying to stick to FOLIO's decision to use vert.x).

> A future has appropriate methods to handle it
> The main reason is to avid async method call on TestContext object.

The TestContext that is provided by vert.x testing tools?

If so, I agree with you, I find the vert.x testing tools to be especially difficult to follow.

> All productive ready code has chain-style approach. Why don't it use in tests?
> I find it well-structured, easy read, async styled approach.

I'll admit I have different beliefs about how test code and production code read.

It may well be that I've given an unfair comparison to the other code in mod-inventory-storage.

I want tests to always produce an output. Asynchronously chained code can easily not do so. For example, if the future in that chain does not complete, what happens during the test?

I would hope that most of the production code that uses this style (there is lots in FOLIO that does not) uses method references or single line lambda statements for the chained steps.

I find the multi-line nested lambda statements difficult to read, with the content obscured by the noise of the callback-esque indentation and brackets.

Does any of that make sense?

marcjohnson-kint: Would it be useful to get feedback from other folks, given that I'm the only one who has interacted with this pull request?

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: Renamed

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: Renamed

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: Renamed

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: renamed

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: renamed

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: renamed

src/test/java/org/folio/rest/api/StorageTestSuite.java
Anatolii-Starkov: Yes it is, In case It's used only in tests and an URL is not created by User(any typos we can avoid) I don't think it's necessary to keep this checked exception because it brings much worries for developers

src/test/java/org/folio/rest/support/HttpClient.java
Anatolii-Starkov: The main reason is to avoid return statement, because it's redundant in a void method

src/test/java/org/folio/rest/support/HttpClient.java
Anatolii-Starkov: Renamed

src/test/java/org/folio/rest/support/db/ErrorBuilder.java
Anatolii-Starkov: folio-service-tools

src/test/java/org/folio/rest/support/HttpClient.java
Anatolii-Starkov: Because I use standard headers from javax.ws.rs.core.HttpHeaders and decided to extract custom headers to the constant for better readability

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: I think it doesn't matter

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: I think it doesn't matter

src/test/java/org/folio/rest/unit/ItemDamagedStatusAPIUnitTest.java
Anatolii-Starkov: It's standard catch if something goes wrong, during refactoring or something else an exception will be wrapped in 500 response

src/test/java/org/folio/rest/support/AdditionalHttpStatusCodes.java
Anatolii-Starkov: Done

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
Anatolii-Starkov: > I want tests to always produce an output. Asynchronously chained code can easily not do so. For example, if the future in that chain does not complete, what happens during the test?

This is a good point. To prevent this we can use it at the end of a chain.
```future.get(5, TimeUnit.SECONDS);```

src/test/java/org/folio/rest/support/db/ErrorBuilder.java
marcjohnson-kint: Why did you choose to copy this code instead of introduce a dependency on the folio-service-tools library?

src/test/java/org/folio/rest/api/ItemDamagedStatusAPITest.java
marcjohnson-kint: Please give that a try, and check that the test finishes with a timeout if the implementation does not respond.

Anatolii-Starkov:
> - the variety of testing styles (vert.x TestContext vs. chained futures vs. fetched futures)

Fixed

> - the quantity of code copied and pasted from the folio-service-tools (rather than introducing a 
reference)

It can't be done. The classes are in test package and they are not present in the folio-service-tools jar

PR #308, opened 2019-08-08
MODINVSTOR-330 Limited number of parallel connections of batch post Instances

KaterynaSenchenko:
Per @julianladisch: "The current implementation of the batch post instances API limits the number of concurrent connections to the postgresql database to 50:
https://github.com/folio-org/mod-inventory-storage/pull/295/commits/e3d60f30da817baa4dc4856ae1991c2425457e9f#diff-c10bf6e19900e0ec47ca321980f8fad7R34

The default maximum number of concurrent connections the database server allows is 100, see max_connections in https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-MAX-CONNECTIONS

It takes 60 seconds until an unused connections is released:
https://github.com/vert-x3/vertx-mysql-postgresql-client/compare/3.8...folio-org:release-connection-pool-3-5-1-folio-1#diff-902e4ca9c95725812a0d6c1b3c6ebe3eR52

If this API is invoked twice at the same time or if two modules each take 50 connections then all read access and all write access to the database is blocked for all modules and all tenants that need a new database connection. It takes at least 60 seconds until it gets unblocked.

Please limit the number of connections to a reasonable size, for example 4."

To set different number of concurrent db connections one needs to specify it as "inventory.storage.parallel.db.connections.limit" program argument.


src/main/java/org/folio/rest/impl/InstanceStorageBatchAPI.java
```diff
@@ -31,7 +31,7 @@
 
   private static final String INSTANCE_TABLE = "instance";
   private static final String BATCH_SIZE_KEY = "inventory.storage.max.batch.size";
-  private static final int MAX_BATCH_SIZE = Integer.parseInt(MODULE_SPECIFIC_ARGS.getOrDefault(BATCH_SIZE_KEY, "50"));
+  private static final int MAX_BATCH_SIZE = Integer.parseInt(MODULE_SPECIFIC_ARGS.getOrDefault(BATCH_SIZE_KEY, "4"));
```
marcjohnson-kint: Where do `MODULE_SPECIFIC_ARGS` come from?

marcjohnson-kint:
Is there documentation for how to change this configuration setting in the modules readme?

src/main/java/org/folio/rest/impl/InstanceStorageBatchAPI.java
KaterynaSenchenko: It's an argument that can be specified in command that runs jar file for module on deployment. I will update the README with an example.

OleksiiKuzminov:
plz update documentation and I'm ok

PR #311, opened 2019-08-13
MODINVSTOR-327, UIIN-470: Add natureOfContentTermIds field to Instance.

bohdan-suprun:
- Add natureOfContentTermIds to Instance domain
- Update sample data
- Create test-case

### DIFF
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -346,6 +346,15 @@
       "folio:linkFromField": "id",
       "folio:linkToField": "instanceId",
       "folio:includedElement": "holdingsRecords"
+    },
+    "natureOfContentTermIds": {
+      "type": "array",
+      "description": "Array of UUID for the Instance nature of content (e.g. bibliography, biography, exhibition catalogue, festschrift, newspaper, proceedings, research report, thesis or website)",
+      "items": {
+          "type": "string",
+          "description": "Single UUID for the Instance nature of content",
+          "$ref": "uuid.json"
+        }
     }
   },
   "additionalProperties": false,
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -350,11 +350,12 @@
     "natureOfContentTermIds": {
       "type": "array",
       "description": "Array of UUID for the Instance nature of content (e.g. bibliography, biography, exhibition catalogue, festschrift, newspaper, proceedings, research report, thesis or website)",
+      "uniqueItems": true,
       "items": {
-          "type": "string",
-          "description": "Single UUID for the Instance nature of content",
-          "$ref": "uuid.json"
-        }
+        "type": "string",
+        "description": "Single UUID for the Instance nature of content",
+        "$ref": "uuid.json"
+      }
     }
   },
   "additionalProperties": false,
--- a/sample-data/instances/girl-on-the-train.json
+++ b/sample-data/instances/girl-on-the-train.json
@@ -25,5 +25,8 @@
       "name": "Creator B"
     }
   ],
-  "instanceTypeId": "6312d172-f0cf-40f6-b27d-9fa8feaf332f"
+  "instanceTypeId": "6312d172-f0cf-40f6-b27d-9fa8feaf332f",
+  "natureOfContentTermIds": [
+    "44cd89f3-2e76-469f-a955-cc57cb9e0395"
+  ]
 }
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -488,6 +488,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "natureOfContentTermIds",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
         }
       ],
       "fullTextIndex": [
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -488,12 +488,6 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
-        },
-        {
-          "fieldName": "natureOfContentTermIds",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
         }
       ],
       "fullTextIndex": [
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -31,20 +31,22 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
+import java.util.stream.Stream;
 
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
 import org.folio.HttpStatus;
 import org.folio.rest.jaxrs.model.MarcJson;
 import org.folio.rest.jaxrs.model.NatureOfContentTerm;
-import org.folio.rest.jaxrs.model.NatureOfContentTerms;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.support.AdditionalHttpStatusCodes;
@@ -80,6 +82,7 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
   private static UUID annexLocationId;
   private static UUID bookMaterialTypeId;
   private static UUID canCirculateLoanTypeId;
+  private Set<String> natureOfContentIdsToRemoveAfterTest = new HashSet<>();
 
   @BeforeClass
   public static void beforeAny() throws Exception {
@@ -115,11 +118,19 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
     StorageTestSuite.deleteAll(itemsStorageUrl(""));
     StorageTestSuite.deleteAll(holdingsStorageUrl(""));
     StorageTestSuite.deleteAll(instancesStorageUrl(""));
+
+    natureOfContentIdsToRemoveAfterTest.clear();
   }
 
   @After
   public void checkIdsAfterEach() {
     StorageTestSuite.checkForMismatchedIDs("instance");
+  }
+
+  @After
+  public void removeGeneratedEntities() {
+    natureOfContentIdsToRemoveAfterTest.forEach(id -> client
+      .delete(natureOfContentTermsUrl("/" + id), TENANT_ID));
   }
 
   @Test
@@ -130,21 +141,15 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
     TimeoutException {
 
     UUID id = UUID.randomUUID();
-    CompletableFuture<Response> natureOfContentGet = new CompletableFuture<>();
-
-    client.get(natureOfContentTermsUrl("/?limit=2"), TENANT_ID, json(natureOfContentGet));
-
-    String[] natureOfContentIds = natureOfContentGet
-      .get(5, SECONDS)
-      .getJson()
-      .mapTo(NatureOfContentTerms.class)
-      .getNatureOfContentTerms()
-      .stream()
+    String[] natureOfContentIds = Stream.of(
+      createNatureOfContentTerm("journal_test"),
+      createNatureOfContentTerm("book_test"))
       .map(NatureOfContentTerm::getId)
       .toArray(String[]::new);
 
     JsonObject instanceToCreate = smallAngryPlanet(id);
-    instanceToCreate.put("natureOfContentTermIds", Arrays.asList(natureOfContentIds));
+    instanceToCreate.put("natureOfContentTermIds", Arrays
+      .asList(natureOfContentIds));
 
     CompletableFuture<Response> createCompleted = new CompletableFuture<>();
 
@@ -1701,5 +1706,25 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
 
     assertThat(String.format("Create item failed: %s", response.getBody()),
       response.getStatusCode(), is(201));
+  }
+
+  private NatureOfContentTerm createNatureOfContentTerm(final String name)
+    throws InterruptedException, ExecutionException, TimeoutException {
+    NatureOfContentTerm natureOfContentTerm = new NatureOfContentTerm()
+      .withId(UUID.randomUUID().toString())
+      .withName(name)
+      .withSource("test");
+
+    CompletableFuture<Response> createNatureOfContent =
+      new CompletableFuture<>();
+
+    client.post(natureOfContentTermsUrl(""), natureOfContentTerm,
+      TENANT_ID, json(createNatureOfContent));
+
+    Response response = createNatureOfContent.get(5, SECONDS);
+    assertThat(response.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    natureOfContentIdsToRemoveAfterTest.add(natureOfContentTerm.getId());
+    return response.getJson().mapTo(NatureOfContentTerm.class);
   }
 }
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -32,6 +32,7 @@ import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Optional;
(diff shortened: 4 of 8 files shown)


ramls/instance.json
```diff
@@ -346,6 +346,15 @@
       "folio:linkFromField": "id",
       "folio:linkToField": "instanceId",
       "folio:includedElement": "holdingsRecords"
+    },
+    "natureOfContentTermIds": {
```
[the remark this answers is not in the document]
bohdan-suprun: Hi @mreno-EBSCO,

Thanks for pointing on this. Yep, you're right, fixed now.

Not sure about case sensitivity, I think UI should provide proper NatureOfContentTerm.ID and we just need to check that no duplicates has been specified + the UUID string has the proper format. 

Is it make sense?

Thanks,
Bohdan

ramls/instance.json
```diff
@@ -346,6 +346,15 @@
...
+      "type": "array",
+      "description": "Array of UUID for the Instance nature of content (e.g. bibliography, biography, exhibition catalogue, festschrift, newspaper, proceedings, research report, thesis or website)",
+      "items": {
+          "type": "string",
+          "description": "Single UUID for the Instance nature of content",
+          "$ref": "uuid.json"
+        }
```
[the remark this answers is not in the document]
bohdan-suprun: Thanks, fixed

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1530,6 +1537,51 @@ public void shouldReturnErrorResponseIfAllInstancesFailed()
...
+
+    Response response = instanceCreate.get(5, SECONDS);
+    assertThat(response.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    // Make sure Nature of content IDs are get saved
+    Instance instanceCreated = response.getJson().mapTo(Instance.class);
+    assertEquals(natureOfContentIds, instanceCreated.getNatureOfContentTermIds());
+    assertEquals(instanceId.toString(), instanceCreated.getId());
+
+    // Re-reed instance by id and make sure nature of content IDs are still here
```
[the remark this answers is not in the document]
bohdan-suprun: Thanks @mreno-EBSCO, fixed now.

marcjohnson-kint:
> Also, we need to consider how this change will affect DB migration, if at all. 

@mreno-EBSCO I agree that database schema and record migration is an important and increasingly urgent topic.

> Will the new index be created automatically on module upgrade?

I think that the RAML Module Builder migration process requires a from version to be specified for that to work.

> We may need to first figure out a process outside this PR, but we cannot lose sight of post Q3.1 DB changes of any kind.

I think we need to more documentation, guidance and a way of testing changes before we can really adopt this into day to day development.

> I think it may, but this needs to be part of the dev process for Q3.2. Chalmers is going live with Q3.1 and we can no longer just blow away data and start over. 

@skoczko Please can you confirm that is the case?

If it is we need to agree when to require this as part of day to day development. And I believe we need to retrofit the relevant migration information to *all* changes made since Q3.1 releases.

I don't know what we need to do for migration information prior to Q3.1 releases, what should go in there?

ramls/instance.json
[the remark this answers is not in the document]
marcjohnson-kint: @mreno-EBSCO Do any of the systems in FOLIO check for unique items in an array, or is it just for informational purposes for the client?

sample-data/instances/girl-on-the-train.json
```diff
@@ -25,5 +25,8 @@
       "name": "Creator B"
     }
   ],
-  "instanceTypeId": "6312d172-f0cf-40f6-b27d-9fa8feaf332f"
+  "instanceTypeId": "6312d172-f0cf-40f6-b27d-9fa8feaf332f",
+  "natureOfContentTermIds": [
+    "44cd89f3-2e76-469f-a955-cc57cb9e0395"
```
marcjohnson-kint: Does a nature of content term already exist for this UUID in the standard reference data?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -488,6 +488,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "natureOfContentTermIds",
```
marcjohnson-kint: Why is an index being added?

@skoczko Is it now standard practice to add indexes for new properties?

Are we intending for this property to be used in filtering, searching or sorting?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1530,6 +1537,54 @@ public void shouldReturnErrorResponseIfAllInstancesFailed()
     assertThat(instances.size(), is(0));
   }
 
+  @Test
+  public void canCreateAnInstanceWithNatureOfContent() throws Exception {
```
marcjohnson-kint: Could the existing test for creating an instance have included this new property instead of a new test?

marcjohnson-kint: What happens if an instance is created with a nature of content term that does not exist?

ramls/instance.json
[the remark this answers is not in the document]
mreno-EBSCO: I believe a Set is generated instead of a List. I don't know if there is any validation done, like a 422 when there are duplicates. I assume they are just coalesced via the Set.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: Hi @marcjohnson-kint, there are no constrains for this, unfortunately.

In order to fix it we will need following:
- Create a separate table which will hold `instance-id` - `nature-of-content` columns (because natureOfContentIds in an Instance is array).
- then create FK for both fields;
- then create trigger for `instance` table which will write those values to the new table

Should we proceed with such a solution? It will increase original estimation of the task I think.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: I've decided to add this index because potentially we may require to search through those fields in the future. Should I remove it for now?

sample-data/instances/girl-on-the-train.json
bohdan-suprun: yep, it is:
```
{
  "id": "44cd89f3-2e76-469f-a955-cc57cb9e0395",
  "name": "textbook",
  "source": "folio"
}
```

and 
```
{
  "id": "0abeee3d-8ad2-4b04-92ff-221b4fce1075",
  "name": "journal",
  "source": "folio"
}
```
for `aba` and `american-journal-of-medicine`

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: Hi @marcjohnson-kint, sure, no prob, will do so.

bohdan-suprun: Hi @marcjohnson-kint, fixed, thanks.

ramls/instance.json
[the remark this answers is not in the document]
marcjohnson-kint: @mreno-EBSCO @bohdan-suprun  Is this important enough that we want to test the behaviour? For example, we submit an instance representation with two references to the same nature of content term and receive back a representation with only a single reference, or does it error?

sample-data/instances/girl-on-the-train.json
marcjohnson-kint: Ok, thanks for checking

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: @bohdan-suprun  

> I've decided to add this index because potentially we may require to search through those fields in the future.

My concern with this approach is that any property is potentially searchable with the way FOLIO uses CQL which means that we could end up creating indexes for every property in a record (with likely significant write performance impact).

That said, deferring index creation has often caused performance surprises later on.

@skoczko @adamdickmeiss @julianladisch What is the guidance around this, should we pre-emptively create indexes?

If so, what kind of index is best suited to an array of UUID strings?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: @bohdan-suprun I think this is worth asking as a question on the issue, including folks like @skoczko @zburke @mkuklis so that we can decide on what to do here?

I don't know off the top of my head what we've done in the past, maybe you could investigate if any of the other array properties have this kind of protection?

If we haven't done it in other places maybe this is outside the scope of this story.

I think there is a simpler approach to providing this check, it does not necessarily need to be done in the database.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -126,8 +130,21 @@ public void canCreateAnInstance()
     TimeoutException {
 
     UUID id = UUID.randomUUID();
+    CompletableFuture<Response> natureOfContentGet = new CompletableFuture<>();
```
marcjohnson-kint: Many of the tests create their test records explicitly in order to improve the independence between tests.

Can you help me understand why this test took the approach of fetching two existing records?

What happens if there aren't any existing nature of content terms? It appears to me that this test would pass without demonstrating the behaviour (it will check that no nature of content terms appear when none were supplied).

This could happen if the nature of content terms are deleted between tests (as is typical, for much of the sample and reference records).

marcjohnson-kint:
Please can a test demonstrate the updating the nature of content terms on an instance. Please use an existing update test if possible.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: @marcjohnson-kint, for me, an integration test should show real business case (e.g. how UI will prepare such a request to create an instance: UI will fetch all available NatureOfContentTerms an allow user to pick only existing one).

However, you're right, it will be an issue when there will be no nature of content available in test time. I'm a new on the project, have not been aware of the approach, fixing it.

src/main/resources/templates/db_scripts/schema.json
skoczko: @bohdan-suprun @marcjohnson-kint Unless it is known that the property will be used for searching, filtering or sorting (e.g there is a feature/user story somewhere in the backlog that will call for this functionality) I would recommend against pre-emptively adding indexes. We now have a "warning" system in RMB which will produce a WARN in the log if a property is used for searching but there is no index.

For UUIDs I would recommend btree (regular "index"), it's the smallest and fastest index type and also allows sorting.

marcjohnson-kint: @skoczko Thanks. @bohdan-suprun I think that is your answer, we should not create an index at this point.

I think that also makes @mreno-EBSCO point about database migrations redundant for this particular change. It is still a very important topic to consider.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: > @marcjohnson-kint, for me, an integration test should show real business case (e.g. how UI will prepare such a request to create an instance: UI will fetch all available NatureOfContentTerms an allow user to pick only existing one).

I agree with you that a test should demonstrate a meaningful situation, and that tests higher up the stack these become more organisationally meaningful.

I think there are trade offs to this, with how succinct tests are, how expressive they are to the most important aspects and how reliable they are at demonstrating the intended behaviour.
 
> However, you're right, it will be an issue when there will be no nature of content available in test time. I'm a new on the project, have not been aware of the approach, fixing it.

Thank you. It isn't a universal approach. This is my personal feedback and approach, rather than an official project stance.

ramls/instance.json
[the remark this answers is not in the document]
mreno-EBSCO: @marcjohnson-kint I think, if this is implemented (I guess it was), then we should have a test for it in case something unexpected happens in a new version of RMB. There have been cases where devs forgot or removed the `@Validate` annotation from the impl... hold on... I just looked, and the annotation is not present on any method in `InstanceStorageImpl.java`. I don't think any of the aspectj based validation is being performed here. The `validate` trait is not present in the raml as well. It is there for `item-storage`. Hmm... ah-ha... found some familiar JIRAs:

This JIRA brings back memories: [MODINVSTOR-24](https://issues.folio.org/browse/MODINVSTOR-24). I created it almost 2 years ago. 🙄
You also created an issue, Marc: [MODINVSTOR-94](https://issues.folio.org/browse/MODINVSTOR-94)

Maybe we should add this to the tech debt list for an upcoming sprint.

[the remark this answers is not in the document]
mreno-EBSCO: Actually, I consider this a bug and not tech debt. We should really schedule it before something bad gets to the DB level. Maybe I am a bit too paranoid about the lack of validation. 😄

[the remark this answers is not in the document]
marcjohnson-kint: > Actually, I consider this a bug and not tech debt. We should really schedule it before something bad gets to the DB level. Maybe I am a bit too paranoid about the lack of validation. 😄

My understanding of FOLIO technical debt is that it can be any issue type. Please feel free to include it in the tech debt options for Cate in the next few weeks.

We need to be aware that it is a breaking implementation compatibility change, and so might cause unexpected issues in the UI.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: @marcjohnson-kint, I've reviewed  mod-inventory-storage module and it seems to me that this is the general problem for Instance entity. 

There are no restrictions for the similar elements, which hold an external ID (e.g. contributor.nameTypeId/typeId/, identifiers.identifierTypeId - I was able to save an instance with ID which does not exist, in unit-test).

But in other modules, like `mod-users` I see, that there is additional java check, that do a query to DB and validates inputs (see `org.folio.rest.impl.UsersAPI#checkAllAddressTypesValid`).

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Hi @skoczko, @marcjohnson-kint, @mreno-EBSCO, I've removed the index from schema.json file, so resolving this discussion.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: @marcjohnson-kint, it has been fixed, thanks.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -111,13 +118,21 @@ public void beforeEach() throws Exception {
...
+    natureOfContentIdsToRemoveAfterTest.clear();
   }
 
   @After
   public void checkIdsAfterEach() {
     StorageTestSuite.checkForMismatchedIDs("instance");
   }
 
+  @After
+  public void removeGeneratedEntities() {
```
marcjohnson-kint: Did you try deleting all of the nature of content records? Did that cause issues?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -126,8 +141,15 @@ public void canCreateAnInstance()
     TimeoutException {
 
     UUID id = UUID.randomUUID();
+    String[] natureOfContentIds = Stream.of(
```
marcjohnson-kint: Whilst this is elegant, I find it harder to read than something like

journalContentTermId = createNatureOfContentTerm("journal_test");
bookContentTermId = createNatureOfContentTerm("book_test");

And then construct the list from those.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: I think this is a requirements question back to Cate and Charlotte, to ask them if the API should check this specifically, for consistency with the UI (given that is a hot topic).

@bohdan-suprun Please ask this question on the issue.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: @marcjohnson-kint I afraid it can cause. Also, we do not have an endpoint to remove all records at once.

bohdan-suprun: Hi @marcjohnson-kint, if we clean-up the whole nature-of-content table than we will get this `org.folio.rest.api.ReferenceTablesTest#natureOfContentTermsLoaded` test broken. Not sure if this is a good idea.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
bohdan-suprun: @marcjohnson-kint, done

bohdan-suprun:
Hi @marcjohnson-kint,

May I ask you to approve the PR? I've resolved all the discussed issues, except the validation of `natureOfContentTermIds`, but let's  discuss it as a separate item, is it fine? 

**Please also note**: https://github.com/folio-org/mod-inventory/pull/141 is now waiting for this changes to be merged.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: Hmm, maybe not. This should probably be left out of this work. 

The reference records loading tests should likely be separate to the rest of the tests.

PR #313, opened 2019-08-16
MODINVSTOR-287, UIIN-593: Add 'itemDamagedStatusDescription' property.

bohdan-suprun:
- Add 'itemDamagedStatusDescription' property
- Update API version
- Add unit test.

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -92,6 +92,10 @@
     "itemDamagedStatusDate": {
       "type": "string"
     },
+    "itemDamagedStatusDescription": {
+      "type": "string",
+      "description": "Holds information about how an item is damaged"
+    },
     "notes": {
       "type": "array",
       "description": "Notes about action, copy, binding etc.",
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -448,6 +448,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
   public void canCreateAnItemWithManyProperties()
     throws MalformedURLException, InterruptedException,
     ExecutionException, TimeoutException {
+    final String damagedDescription = "Somehow damaged";
 
     UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
 
@@ -465,6 +466,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("temporaryLocationId", annexLibraryLocationId.toString());
     itemToCreate.put("tags", new JsonObject().put("tagList", new JsonArray().add(TAG_VALUE)));
+    itemToCreate.put("itemDamagedStatusDescription", damagedDescription);
 
     //TODO: Replace with real service point when validated
     itemToCreate.put("inTransitDestinationServicePointId", inTransitServicePointId);
@@ -496,6 +498,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     assertThat(item.getString("inTransitDestinationServicePointId"), is(inTransitServicePointId));
     assertThat(tags.size(), is(1));
     assertThat(tags, hasItem(TAG_VALUE));
+    assertThat(item.getString("itemDamagedStatusDescription"), is(damagedDescription));
   }
 
   @Test
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "7.5",
+      "version": "7.6",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/item-storage.raml
+++ b/ramls/item-storage.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Item Storage
-version: v7.5
+version: v7.6
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 


ramls/item.json
```diff
@@ -92,6 +92,10 @@
     "itemDamagedStatusDate": {
       "type": "string"
     },
+    "itemDamagedStatusDescription": {
```
bohdan-suprun: Hi @marcjohnson-kint, JIRA says that the new property should be named as `itemDamagedDescription`, but existing properties have  `itemDamagedStatusDate` and `itemDamagedStatusId` names, should the new property follow the name pattern and be named as  `itemDamagedStatusDescription`?

mreno-EBSCO: I'll leave @marcjohnson-kint to be the final word here, but to me, `itemDamagedDescription` seems right (I'd prefer just `damageDescription`, since we already know it is an `item`, but again I'll defer to Marc). The description is of the damage to the item. The date is for the status change from whatever to `Damaged` or whatever the damaged status needs to be. The id also refers to the specific damage status, like `Damaged` or `Damaged by patron` or `Damaged in transit` or pretty much anything the library wants.

However, the description here only applies to the damage of the item, like `Page 5 missing` or `Coffee spilled on cover` or `Currently on fire`. This doesn't describe the status, but the item itself. I suppose the damage status could change several times due to a item with bad luck, but the description would continue to evolve/grow as more damage occurs or is repaired.

I'm actually surprised damage is not repeatable. Wait a second, according to scenario 2 in UIIN-593 (a linked story in the invstor JIRA), the elements are  "Not repeatable/Not required" for viewing, but for editing, scenario 1, the elements are "repeatable, not required". Maybe it is just me, but that is confusing. In UXPROD-1659, Charlotte asks Emma if a decision on whether or not damage is repeatable, but there is no response. UIIN-583 was created before this comment was added to UXPROD-1659. After some digging, it looks like UIIN-562 confirms that this has been decided "damage" to be a non-repeatable field.

I think we need to get the POs to do a better job coordinating and documenting clear requirements. UIIN-562 is about adding damage status types, not the overall feature. OK, enough ranting. My comment still stands about `itemDamagedDescription`. 😄

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -465,6 +466,7 @@ public void canCreateAnItemWithManyProperties()
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("temporaryLocationId", annexLibraryLocationId.toString());
     itemToCreate.put("tags", new JsonObject().put("tagList", new JsonArray().add(TAG_VALUE)));
+    itemToCreate.put("itemDamagedStatusDescription", damagedDescription);
```
mreno-EBSCO: Should we allow a damage description without a damage status being set? How are these properties intended to work together, if at all?  Maybe "damage" should be a sub-object with certain required fields? What do you think @marcjohnson-kint?

bohdan-suprun: Hi @mreno-EBSCO, good point. I agree that those three properties `(itemDamagedStatusId`, `itemDamagedStatusDate`, `itemDamagedDescription`) should be handled only together.

What about following structure?
```javascript
"damage": {
  "type": "object",
  "description": "Item damage information",
  "properties": {
    "statusId": {
      "description": "Damage status ID",
      "type": "string"
    },
    "statusDate": {
      "description": "Damage date",
      "type": "string"
    },
    "description": {
      "type": "string",
      "description": "Holds information about how an item is damaged"
    }
  },
  "additionalProperties": false,
  "required": [
    "statusId",
    "statusDate",
    "description"
  ]
}
```

But then, it will be breaking changes for UI. We should let UI guys know about it. Not aware about process in this case. 

Let's wait response from @marcjohnson-kint.

ramls/item.json
marcjohnson-kint: > I'll leave @marcjohnson-kint to be the final word here, but to me, `itemDamagedDescription` seems right 

@bohdan-suprun I think I pretty much agree with @mreno-EBSCO 

> (I'd prefer just `damageDescription`, since we already know it is an `item`, but again I'll defer to Marc). 

I wasn't involved in the earlier work in this area. I'd hope that if I had been I might have designed it something like:

```
“damage” (or damaged, but that doesn't read as well to me) : {
 “statusId”: ...
 “description”: ...
 “lastChangedDate”: ....
}
```

However that comes with the retrospective coherence based upon the scenarios in this issue, so I might have fallen into similar pitfalls.

> The description is of the damage to the item. The date is for the status change from whatever to `Damaged` or whatever the damaged status needs to be. The id also refers to the specific damage status, like `Damaged` or `Damaged by patron` or `Damaged in transit` or pretty much anything the library wants.

> However, the description here only applies to the damage of the item, like `Page 5 missing` or `Coffee spilled on cover` or `Currently on fire`. This doesn't describe the status, but the item itself. I suppose the damage status could change several times due to a item with bad luck, but the description would continue to evolve/grow as more damage occurs or is repaired.

My interpretations of the requirements is that this is the description of the damage, not a description of the status of the damage.

To be consistent with the existing properties, that would mean naming it `itemDamagedDescription` which doesn't read well to me, but consistency and implicit grouping is likely more important than readability at this point.

Given you are unclear, please feel free to confirm with the PO by asking a question on the issue. If you don't feel comfortable doing that, let me know, and I can ask on your behalf.

> I'm actually surprised damage is not repeatable. Wait a second, according to scenario 2 in UIIN-593 (a linked story in the invstor JIRA), the elements are "Not repeatable/Not required" for viewing, but for editing, scenario 1, the elements are "repeatable, not required". Maybe it is just me, but that is confusing. In UXPROD-1659, Charlotte asks Emma if a decision on whether or not damage is repeatable, but there is no response. UIIN-583 was created before this comment was added to UXPROD-1659. After some digging, it looks like UIIN-562 confirms that this has been decided "damage" to be a non-repeatable field.

I'm both surprised and confused.

> I think we need to get the POs to do a better job coordinating and documenting clear requirements. UIIN-562 is about adding damage status types, not the overall feature. OK, enough ranting. My comment still stands about `itemDamagedDescription`. 😄

I think it's clear from the previous work that there has been some confusion in the requirements for this area.

For example, scenario 3 describes updating the date when the description changes. Unfortunately this property has been named `itemDamagedStatusDate` in the API, which is somewhat misleading now it's broader purpose is understood.

This is probably compatibility breaking technical debt, should we raise an issue for it?

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: @mreno-EBSCO @bohdan-suprun I don't think I have good answers for you, as these questions feel like requirements gaps.



> Should we allow a damage description without a damage status being set? How are these properties intended to work together, if at all?

I think these are good questions for Charlotte as a comment on the issue.

> Maybe "damage" should be a sub-object with certain required fields? What do you think @marcjohnson-kint?

> What about following structure?

As above, I would prefer these properties were grouped together, however they are not and changing that would be compatibility breaking, and I don't really want to do that right now, unless we feel it is necessary?

Also as above, my interpretation of Scenario 3 is that the date is for any changes to damage information, not just the status, so should really be named more generally.

We could describe this challenge on the issue and let Cate and Charlotte decide if we want to feel the pain now or later (where it will be more expensive because we would need to migrate data due to Chalmers being in production).

@skoczko What do you think?

> But then, it will be breaking changes for UI. We should let UI guys know about it. Not aware about process in this case.

The process involves coordinating changes to all dependent modules in the reference system, which for items is probably quite a few, especially as this would likely also mean a breaking change to all dependents of the `inventory` interface too.

marcjohnson-kint:
@bohdan-suprun Has this work been abandoned? If so, please close this pull request. If not, please update the branch.

bohdan-suprun:
> @bohdan-suprun Has this work been abandoned? If so, please close this pull request. If not, please update the branch.

Closing the PR.

PR #314, opened 2019-08-16
Fix Aspect issue (mvn clean; mvn test; mvn test)

adamdickmeiss:
So annoying that I always have to clean to make tests work


marcjohnson-kint:
@julianladisch I don’t have sufficient understanding of AspectJ or RAML Module Builder to review this change, do you?

marcjohnson-kint:
@adamdickmeiss Is there any reason this shouldn't be merged? Now that it has been approved by @julianladisch ?

PR #316, opened 2019-08-16
MODINVSTOR-301 Deletion of a Resource type requires a check to determ…

Anatolii-Starkov:
-added the foreign key from instance table to the instance_type table
-added test

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -66,7 +66,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.2",
+      "version": "7.3",
       "handlers": [
         {
           "methods": ["GET"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -66,7 +66,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.1",
+      "version": "7.2",
       "handlers": [
         {
           "methods": ["GET"],
@@ -133,7 +133,7 @@
     },
     {
       "id": "instance-storage-batch",
-      "version": "0.2",
+      "version": "0.3",
       "handlers": [
         {
           "methods": ["POST"],
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -66,7 +66,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.3",
+      "version": "7.2",
       "handlers": [
         {
           "methods": ["GET"],
@@ -133,7 +133,7 @@
     },
     {
       "id": "instance-storage-batch",
-      "version": "0.3",
+      "version": "0.2",
       "handlers": [
         {
           "methods": ["POST"],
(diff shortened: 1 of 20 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -66,7 +66,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.1",
+      "version": "8.0",
```
marcjohnson-kint: Why is this a breaking compatibility change?

Anatolii-Starkov: Probably yes

marcjohnson-kint: > Probably yes

Yes, to causing confusion?

How could I have described that differently? Are you aware of the difference in FOLIO between interface and implementation compatibility?

Anatolii-Starkov: Sorry Marc I didn't catch it at once. I will fix it. It should be 7.2 for the instance-storage and 0.3 for the instance-storage-batch

ramls/instance-storage-batch.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory Storage Batch API
-version: v0.2
+version: v1.0
```
Igor-Gorchakov: Why the Interface version has increment if batch-API has not changed ?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: Why does the interface version need to change in any way? As far as I can tell, the interface hasn't been changed in any way.

What has changed is that the implementation is choosing to reject instances where the resource type reference does not refer to a resource type reference record?

Anatolii-Starkov: Is it fine now? If no please provide your input how it should look like

Anatolii-Starkov:
Thanks Marc for you explanation. I have refreshed my knowledge about versions on FOLIO 
(https://dev.folio.org/guidelines/contributing/#version-numbers)
So, I reverted versions of API 
instance-storage from 7.3 to 7.2 
instance-storage-batch from 0.3 to 0.2
and increased version of implementation from 16.1.0-SNAPSHOT to 17.0.0-SNAPSHOT

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -574,36 +569,7 @@ private JsonObject uprooted(UUID id) {
     tags.add("test-tag");
 
     return createInstanceRequest(id, "TEST", "Uprooted",
-      identifiers, contributors, UUID_TEXT, tags);
-  }
-
-  private JsonObject temeraire(UUID id) {
```
marcjohnson-kint: Why has this been removed?

Anatolii-Starkov: It isn't used

marcjohnson-kint: Ok, thanks. In future, please try to minimize the amount of extraneous changes within a pull request. 

Please feel free to issue preparatory pull requests that perform this kind of cleanup only.

PR #317, opened 2019-08-23
[MODINVSTOR-318] Respond 422 on post holdings type with duplicate name

opochepam:
## Purpose
Resolves: MODINVSTOR-318

Controlled Vocab Duplicate Issues (Holdings Type and Call Number type) Change server response for errors during loan type creation.
## Links
https://issues.folio.org/browse/MODINVSTOR-318

### DIFF
--- a/src/main/java/org/folio/rest/impl/HoldingsTypeAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsTypeAPI.java
@@ -1,5 +1,8 @@
 package org.folio.rest.impl;
 
+import static java.util.Collections.singletonList;
+import static org.folio.rest.persist.PgUtil.postgresClient;
+
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


src/main/java/org/folio/rest/impl/HoldingsTypeAPI.java
```diff
@@ -229,5 +210,31 @@ private void internalServerErrorDuringPut(Throwable e, String lang, Handler<Asyn
...
+    Future<String> future = Future.future();
+    pgClient.save(REFERENCE_TABLE, entity.getId(), entity, future);
+    return future;
+  }
 
+  private PostHoldingsTypesResponse handleSaveHoldingsTypeException(Throwable t) {
+    if (PgExceptionUtil.isUniqueViolation(t)) {
+      Error error = new Error()
+        .withCode("name.duplicate")
+        .withMessage("Cannot create entity with not unique name")
```
craigmcnally: Consider changing "not unique" -> "nonunique" or rephrasing e.g. "Cannot create entity; name is not unique"

craigmcnally: Also consider making this a static constant so it can be referenced in the unit tests and you don't have to worry about keeping the string literals synchronized

PR #318, opened 2019-08-26
MODINVSTOR-342: Update to RMB 27

julianladisch:
* Update to RMB 27, use targetPath foreign key syntax
* add sample data unit test

Before merging change RMB version to 27!


julianladisch:
The sample data test is not needed.
If the test existed all these issues were caught before release: https://issues.folio.org/browse/FOLIO-2102

PR #320, opened 2019-08-29
Remove UUID constraints for Instance TypeIds

Igor-Gorchakov:
Instance creation fails (Unprocessable Entity) during Record-to-Instance mapping, because of many types in Instance schema are limited to receive only UUID. But some types can be not only UUID, also plain text. 

We need to skip UUID constraints out of Instance schema for the next fields: 
* identifiers.identifierTypeId
* contributors.contributorTypeId,  contributorNameTypeId
* classifications.classificationTypeId
* notes.instanceNoteTypeId


Igor-Gorchakov:
Discussed record-to-instance mapping logic, no need to in changes here.

mreno-EBSCO:
Makes sense. I was going to suggest a mapping of some sort instead but did not know the context well enough to know if that was feasible. Thanks!

PR #322, opened 2019-09-04
MODINVSTOR-343: Add preceding-succeeding relationship. Add sample data.

bohdan-suprun:
* Add preceding-succeeding relationship type
* Add tests 
* Add sample data

### DIFF
--- a/src/test/java/org/folio/rest/api/ReferenceTablesTest.java
+++ b/src/test/java/org/folio/rest/api/ReferenceTablesTest.java
@@ -13,6 +13,7 @@ import static org.folio.rest.support.http.InterfaceUrls.identifierTypesUrl;
 import static org.folio.rest.support.http.InterfaceUrls.illPoliciesUrl;
 import static org.folio.rest.support.http.InterfaceUrls.instanceFormatsUrl;
 import static org.folio.rest.support.http.InterfaceUrls.instanceNoteTypesUrl;
+import static org.folio.rest.support.http.InterfaceUrls.instanceRelationshipTypesUrl;
 import static org.folio.rest.support.http.InterfaceUrls.instanceStatusesUrl;
 import static org.folio.rest.support.http.InterfaceUrls.instanceTypesUrl;
 import static org.folio.rest.support.http.InterfaceUrls.itemNoteTypesUrl;
@@ -31,6 +32,7 @@ import java.net.URL;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
+import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
@@ -49,6 +51,7 @@ import org.folio.rest.api.entities.IllPolicy;
 import org.folio.rest.api.entities.Instance;
 import org.folio.rest.api.entities.InstanceFormat;
 import org.folio.rest.api.entities.InstanceNoteType;
+import org.folio.rest.api.entities.InstanceRelationshipType;
 import org.folio.rest.api.entities.InstanceStatus;
 import org.folio.rest.api.entities.InstanceType;
 import org.folio.rest.api.entities.ItemNoteType;
@@ -701,6 +704,28 @@ public class ReferenceTablesTest extends TestBase {
 
     testGetPutDeletePost(statisticalCodeTypesPath, entityUUIDCodeType, statisticalCodeType, updatePropertyCodeType);
 
+  }
+
+  @Test
+  public void instanceRelationshipTypesLoaded() throws Exception {
+    Response searchResponse = getReferenceRecords(instanceRelationshipTypesUrl(""));
+    validateNumberOfReferenceRecords("instance-relationship types",
+      searchResponse, 4, 200);
+  }
+
+  @Test
+  public void instanceRelationshipTypesCrud() throws Exception {
+    final String entityPath = "/instance-relationship-types";
+    final String instanceRelationshipId = UUID.randomUUID().toString();
+
+    InstanceRelationshipType instanceRelationshipType =
+      new InstanceRelationshipType(instanceRelationshipId, "Test Type");
+    Response createResponse = createReferenceRecord(entityPath, instanceRelationshipType);
+
+    assertThat(createResponse.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    testGetPutDeletePost(entityPath, instanceRelationshipId, instanceRelationshipType,
+      InstanceRelationshipType.NAME_KEY);
   }
 
   private Response getReferenceRecords(URL baseUrl)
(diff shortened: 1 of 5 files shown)


bohdan-suprun:
> It looks fine to me. I'm not well versed with this relationship domain so I'm not sure if the superInstance and subInstance sample data Ids are in the correct place (they may as well be). What determines which title should be super- or sub-instance?

Hi @mtraneis, preceding title is a child (sub) instance and succeeding is a parent (super) instance. See this comment in the ticket https://issues.folio.org/browse/MODINVSTOR-343?focusedCommentId=59016&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-59016

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
```diff
@@ -703,6 +706,28 @@ public void statisticalCodesAndCodeTypesBasicCrud()
...
 
+  @Test
+  public void instanceRelationshipTypesLoaded() throws Exception {
+    Response searchResponse = getReferenceRecords(instanceRelationshipTypesUrl(""));
+    validateNumberOfReferenceRecords("instance-relationship types",
+      searchResponse, 4, 200);
+  }
+
+  @Test
+  public void instanceRelationshipTypesCrud() throws Exception {
```
marcjohnson-kint: What does this test check?

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
```diff
@@ -703,6 +706,28 @@ public void statisticalCodesAndCodeTypesBasicCrud()
 
   }
 
+  @Test
+  public void instanceRelationshipTypesLoaded() throws Exception {
+    Response searchResponse = getReferenceRecords(instanceRelationshipTypesUrl(""));
+    validateNumberOfReferenceRecords("instance-relationship types",
+      searchResponse, 4, 200);
```
marcjohnson-kint: If we have a test which checks for the expected count of relationship types, why does this check for a range?

Is there a reason why we aren't sure how many should be loaded?

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
bohdan-suprun: @marcjohnson-kint, it is basic CRUD operations for `/instance-relationship-types`  endpoint, for some reason there was no test present for this endpoint.

marcjohnson-kint: Is this how the other tests for reference records are structured?

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
bohdan-suprun: @marcjohnson-kint, It was a question for me as well. We use such checks for all the reference data tests. 

My guess, that maven test runner do not run tests in the same order every time and because of this there might be test cases that creates additional elements, so we just need some delta to not depend on the order.

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
bohdan-suprun: @marc-johnson yes, commonly we have one test to check that reference data were loaded to DB and one test to check CRUD operations.

marcjohnson-kint: Ok, if these are the established patterns for this area, then any feedback I might have is likely not relevant for this pull request.

(I think that records created during tests should be deleted prior to the next test, in order to improve the reliability and specificity of tests)

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
marcjohnson-kint: Ok, if these are the established patterns for this area, then any feedback I might have is likely not relevant for this pull request.

(I think that records created during tests should be deleted prior to the next test, in order to improve the reliability and specificity of tests)

mreno-EBSCO: It is best practice that each test should be self contained and execution order should not matter. If order needs to matter, of course, there are ways to do it. But testing ranges instead of exact results yields a fragile test, IMHO. Tests that create things should also clean up those things. Are we not doing there here?

bohdan-suprun:
Marked with WIP, because we have to merge this PR together with https://github.com/folio-org/mod-inventory/pull/146 to not introduce breaking change.

marcjohnson-kint:
> Marked with WIP, because we have to merge this PR together with [folio-org/mod-inventory#146](https://github.com/folio-org/mod-inventory/pull/146) to not introduce breaking change.

I don't understand. How are these changes breaking?

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
marcjohnson-kint: @mreno-EBSCO I agree and I'd prefer it that way. 

There are quite a few places where that is not done, quite often with regards to the built in sample or reference records.

This pull request doesn't seem to me to be the place to resolve those more broader issues.

bohdan-suprun:
> > Marked with WIP, because we have to merge this PR together with [folio-org/mod-inventory#146](https://github.com/folio-org/mod-inventory/pull/146) to not introduce breaking change.
> 
> I don't understand. How are these changes breaking?

This PR is the 1st part of the preceding-succeeding title feature, here I just added new relationship type.

But in the 2nd part of the feature (https://github.com/folio-org/mod-inventory/pull/146) I've added new preceding/succeedingTitles properties to instance (similar to child/parentInstances), so these properties will be used only for this particular relationship type (preceedingTitles == subInstances, succeedingTitles == superInstances).

And it is possible, that some API consumers/users can create instance relationship with such type and expect to see them in child/parentInstances array. But, when we merge the 2nd part, those relationships will be present in preceding/succeedingTitles array instead of parent/childInstances.

It seems like breaking change for me, is it?

marcjohnson-kint:
> > > Marked with WIP, because we have to merge this PR together with [folio-org/mod-inventory#146](https://github.com/folio-org/mod-inventory/pull/146) to not introduce breaking change.
> > 
> > 
> > I don't understand. How are these changes breaking?
> 
> This PR is the 1st part of the preceding-succeeding title feature, here I just added new relationship type.
> 
> But in the 2nd part of the feature ([folio-org/mod-inventory#146](https://github.com/folio-org/mod-inventory/pull/146)) I've added new preceding/succeedingTitles properties to instance (similar to child/parentInstances), so these properties will be used only for this particular relationship type (preceedingTitles == subInstances, succeedingTitles == superInstances).
> 
> And it is possible, that some API consumers/users can create instance relationship with such type and expect to see them in child/parentInstances array. But, when we merge the 2nd part, those relationships will be present in preceding/succeedingTitles array instead of parent/childInstances.
> 
> It seems like breaking change for me, is it?

I don't think I follow. How can adding a relationship type be a breaking change?

I can understand how separating these out into different properties could be considered a breaking behavioural change. Are you suggesting that it wouldn't be, if it is added at the same time as the relationship type?

PR #327, opened 2019-09-13
MODINVSTOR-348: Calculate and persist effective location for item.

bohdan-suprun:
Add set of triggers (for item and holdings_record tables) to persist effectiveLocation for item.

**Logic to calculate item effectiveLocation:**
First not null:
1. Item temporaryLocation;
2. Item permanentLocation;
3. Holding record temporaryLocation;
4. Holding record permanentLocation;

**Holding records trigger implementation details**:
* Updates related items only if following properties have changed: 
  * temporaryLocationId; 
  * permanentLocationId;
* Only those items are updated which has no value for their's permanent or temporary locations because otherwise the update won't take any affect.

**Item trigger implementation details:**
* Item effective location is re-calculated only if following properties changed: 
  * temporaryLocationId; 
  * permanentLocationId;
  * holdingsRecordId.

Related PR https://github.com/folio-org/mod-inventory/pull/150

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -202,7 +202,7 @@
     },
     "effectiveLocationId": {
       "type": "string",
-      "description": "The current home location for the item. Read only, it is automatically updated to be the first defined value from item.temporaryLocationId, item.permanentLocationId, holdingsRecord.temporaryLocationId, holdingsRecord.permanentLocationId.",
+      "description": "Read only current home location for the item.",
       "$ref": "uuid.json",
       "readonly": true
     },
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -200,6 +200,11 @@
     "temporaryLocationId": {
       "type": "string"
     },
+    "effectiveLocationId": {
+      "type": "string",
+      "description": "The effective location is used by FOLIO and other integrated systems to know the current home location for the item (read only, derived from locations on HoldingsRecord and Item)",
+      "readonly": true
+    },
     "electronicAccess": {
       "type": "array",
       "items": {
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -202,7 +202,7 @@
     },
     "effectiveLocationId": {
       "type": "string",
-      "description": "The effective location is used by FOLIO and other integrated systems to know the current home location for the item (read only, derived from locations on HoldingsRecord and Item)",
+      "description": "The current home location for the item. Read only, it is automatically updated to be the first defined value from item.temporaryLocationId, item.permanentLocationId, holdingsRecord.temporaryLocationId, holdingsRecord.permanentLocationId.",
       "$ref": "uuid.json",
       "readonly": true
     },
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -203,6 +203,7 @@
     "effectiveLocationId": {
       "type": "string",
       "description": "The effective location is used by FOLIO and other integrated systems to know the current home location for the item (read only, derived from locations on HoldingsRecord and Item)",
+      "$ref": "uuid.json",
       "readonly": true
     },
     "electronicAccess": {
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -1,13 +1,11 @@
 package org.folio.rest.impl;
 
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Context;
-import io.vertx.core.Future;
-import io.vertx.core.Handler;
-import io.vertx.core.logging.Logger;
-import io.vertx.core.logging.LoggerFactory;
+import static org.folio.rest.support.ResponseUtil.copyResponseWithNewEntity;
+import static org.folio.rest.support.ResponseUtil.hasCreatedStatus;
+import static org.apache.commons.lang3.ObjectUtils.firstNonNull;
 
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 
 import javax.ws.rs.core.Response;
 
@@ -18,8 +16,14 @@ import org.folio.rest.jaxrs.model.Status;
 import org.folio.rest.jaxrs.resource.ItemStorage;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
-import org.folio.rest.support.ResponseUtil;
 import org.folio.rest.tools.utils.TenantTool;
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
 
 /**
  * CRUD for Item.
@@ -58,12 +62,16 @@ public class ItemStorageAPI implements ItemStorage {
     }
 
     PgUtil.post(ITEM_TABLE, entity, okapiHeaders, vertxContext,
-      PostItemStorageItemsResponse.class, response -> {
+      PostItemStorageItemsResponse.class, postResponse -> {
         // Have to re-read item to get calculated fields like effectiveLocationId
-        if (ResponseUtil.hasCreatedStatus(response.result())) {
-          rereadCreatedItemAndRespond(entity, okapiHeaders, asyncResultHandler, vertxContext);
+        if (hasCreatedStatus(postResponse.result())) {
+          readItemById(entity.getId(), okapiHeaders, vertxContext)
+            // copy original response to save all headers etc. and set
+            // the retrieved item or set the original entity in case item is null
+            .thenApply(item -> copyResponseWithNewEntity(postResponse.result(), firstNonNull(item, entity)))
+            .thenAccept(respToSend -> asyncResultHandler.handle(Future.succeededFuture(respToSend)));
         } else {
-          asyncResultHandler.handle(Future.succeededFuture(response.result()));
+          asyncResultHandler.handle(postResponse);
         }
       });
   }
@@ -125,27 +133,15 @@ public class ItemStorageAPI implements ItemStorage {
         DeleteItemStorageItemsByItemIdResponse.class, asyncResultHandler);
   }
 
-  private void rereadCreatedItemAndRespond(
-    Item entity, Map<String, String> okapiHeaders,
-    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+  private CompletableFuture<Item> readItemById(
+    String itemId, Map<String, String> okapiHeaders, Context vertxContext) {
+    final CompletableFuture<Item> readItemFuture = new CompletableFuture<>();
 
     PgUtil.postgresClient(vertxContext, okapiHeaders)
-      .getById(ITEM_TABLE, entity.getId(), Item.class, response -> {
-        if (response.succeeded() && response.result() != null) {
-          asyncResultHandler.handle(respond201Created(response.result()));
-        } else {
-          log.warn("Unable to re-read item [{}] returning old item",
-            entity.getId());
-          asyncResultHandler.handle(respond201Created(entity));
-        }
-      });
-  }
+      .getById(ITEM_TABLE, itemId, Item.class,
+        response -> readItemFuture.complete(response.result())
+      );
 
-  private Future<Response> respond201Created(Item item) {
-    return Future.succeededFuture(PostItemStorageItemsResponse
-      .respond201WithApplicationJson(
-        item,
-        PostItemStorageItemsResponse.headersFor201()
-      ));
+    return readItemFuture;
   }
 }
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -18,6 +18,7 @@ import org.folio.rest.jaxrs.model.Status;
 import org.folio.rest.jaxrs.resource.ItemStorage;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
+import org.folio.rest.support.ResponseUtil;
 import org.folio.rest.tools.utils.TenantTool;
 
 /**
@@ -55,8 +56,16 @@ public class ItemStorageAPI implements ItemStorage {
     if (entity.getStatus() == null) {
       entity.setStatus(new Status().withName(DEFAULT_STATUS_NAME));
     }
+
     PgUtil.post(ITEM_TABLE, entity, okapiHeaders, vertxContext,
-        PostItemStorageItemsResponse.class, asyncResultHandler);
+      PostItemStorageItemsResponse.class, response -> {
+        // Have to re-read item to get calculated fields like effectiveLocationId
+        if (ResponseUtil.hasCreatedStatus(response.result())) {
+          rereadCreatedItemAndRespond(entity, okapiHeaders, asyncResultHandler, vertxContext);
+        } else {
+          asyncResultHandler.handle(Future.succeededFuture(response.result()));
+        }
+      });
   }
 
   @Validate
@@ -114,5 +123,29 @@ public class ItemStorageAPI implements ItemStorage {
 
     PgUtil.deleteById(ITEM_TABLE, itemId, okapiHeaders, vertxContext,
         DeleteItemStorageItemsByItemIdResponse.class, asyncResultHandler);
+  }
+
+  private void rereadCreatedItemAndRespond(
+    Item entity, Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+
+    PgUtil.postgresClient(vertxContext, okapiHeaders)
+      .getById(ITEM_TABLE, entity.getId(), Item.class, response -> {
+        if (response.succeeded() && response.result() != null) {
+          asyncResultHandler.handle(respond201Created(response.result()));
+        } else {
+          log.warn("Unable to re-read item [{}] returning old item",
+            entity.getId());
+          asyncResultHandler.handle(respond201Created(entity));
+        }
+      });
+  }
+
+  private Future<Response> respond201Created(Item item) {
+    return Future.succeededFuture(PostItemStorageItemsResponse
+      .respond201WithApplicationJson(
+        item,
+        PostItemStorageItemsResponse.headersFor201()
+      ));
   }
 }
--- a/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
+++ b/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
@@ -26,7 +26,7 @@ AS $$
   END;
   $$ LANGUAGE 'plpgsql';
 
--- Set Item effective location in case insert or update of temporary/permanentLocationId or holdingsRecordId properties
+-- Set Item effective location when inserting or updating an Item.
 CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_effective_location_on_item_update() RETURNS trigger
 AS $$
   declare
--- a/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
+++ b/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
@@ -1,0 +1,85 @@
+-- Returns item effectiveLocationId.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.calculate_item_effective_location(hr_perm_location text,
+           hr_temp_location text, itm_perm_location text, itm_temp_location text) RETURNS jsonb
+AS $$
+  begin
+   RETURN ('"' || coalesce(itm_temp_location, itm_perm_location, hr_temp_location, hr_perm_location) || '"')::jsonb;
+  END;
+  $$ LANGUAGE 'plpgsql';
+
+-- Updates item effective location property in case holding record permanent or temp location has changed.
+-- only items that do not have permanent or temp location set will be updated.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_effective_location_on_holding_update() RETURNS trigger
+AS $$
+  begin
+	  -- null-safe comparison of OLD vs NEW location
+	  -- have to update effectiveLocation only if related fields have been changed
+	  if (OLD.jsonb->>'permanentLocationId' is distinct from NEW.jsonb->>'permanentLocationId'
+	  	  OR OLD.jsonb->>'temporaryLocationId' is distinct from NEW.jsonb->>'temporaryLocationId') THEN
+	  	  UPDATE ${myuniversity}_${mymodule}.item
+	  	  SET jsonb = jsonb_set(jsonb,
+		  					  '{effectiveLocationId}',
+		  					   ${myuniversity}_${mymodule}.calculate_item_effective_location(
+	  							   new.jsonb->>'permanentLocationId',
+	  							   new.jsonb->>'temporaryLocationId',
+	  							   jsonb->>'permanentLocationId',
+	  							   jsonb->>'temporaryLocationId')
+	  			      )
+	  	  WHERE holdingsrecordid = new.id
+	  	  -- if item's either temp or perm location is not null then we don't need to update it's effective location
+	  	  -- as item's location has more priority than holding's one.
+	  	  AND jsonb->>'permanentLocationId' is null AND jsonb->>'temporaryLocationId' is null;
+	  end if;
+   RETURN NEW;
+  END;
+  $$ LANGUAGE 'plpgsql';
+
+-- Set Item effective location in case insert or update of temporary/permanentLocationId or holdingsRecordId properties
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_effective_location_on_item_update() RETURNS trigger
+AS $$
+	declare
+		hr_perm_location text;
+		hr_temp_location text;
+		effective_location jsonb;
+	begin
+
+	  -- skip the trigger in case item's sensitive fields were not updated.
+	  if (TG_OP = 'UPDATE' and OLD.jsonb->>'holdingsRecordId' IS NOT DISTINCT FROM NEW.jsonb->>'holdingsRecordId'
+	      AND OLD.jsonb->>'temporaryLocationId' IS NOT DISTINCT FROM NEW.jsonb->>'temporaryLocationId'
+	      AND OLD.jsonb->>'permanentLocationId' IS NOT DISTINCT FROM NEW.jsonb->>'permanentLocationId') then
+	     return new;
+	  end if;
+
+		-- we need to retrieve holding_record temp and perm location only if item location fields are empty
+		-- otherwise it does not make any affect as item location takes precedence over holding location
+		if(NEW.jsonb ->> 'temporaryLocationId' is null and NEW.jsonb ->> 'permanentLocationId' is null) then
+			select jsonb->>'temporaryLocationId', jsonb->>'permanentLocationId'
+			 into hr_temp_location, hr_perm_location
+			from ${myuniversity}_${mymodule}.holdings_record
+			 where id::text = new.jsonb->>'holdingsRecordId' limit 1;
+		else
+		  -- initialize to nulls as they won't change final result
+			hr_perm_location = null;
+			hr_temp_location = null;
+		end if;
+		new.jsonb = jsonb_set(new.jsonb,
+		                      '{effectiveLocationId}',
+		                      ${myuniversity}_${mymodule}.calculate_item_effective_location(
+									            hr_perm_location,
+									            hr_temp_location,
+									            new.jsonb->>'permanentLocationId',
+									            new.jsonb->>'temporaryLocationId')
+									        );
+	  return new;
+	END;
+$$ LANGUAGE 'plpgsql';
+
+DROP TRIGGER IF EXISTS update_effective_location_for_items ON ${myuniversity}_${mymodule}.holdings_record;
+-- should be after update trigger, will allow item trigger to fetch up-to-date locations from holding
+create trigger update_effective_location_for_items after update
+on ${myuniversity}_${mymodule}.holdings_record for each row execute procedure ${myuniversity}_${mymodule}.update_effective_location_on_holding_update();
+
+
+DROP TRIGGER IF EXISTS update_effective_location ON ${myuniversity}_${mymodule}.item;
+create trigger update_effective_location before insert or update
+on ${myuniversity}_${mymodule}.item for each row execute procedure ${myuniversity}_${mymodule}.update_effective_location_on_item_update();
--- a/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
+++ b/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
@@ -9,21 +9,19 @@ AS $$
   BEGIN
     old_effective_location_id = coalesce(OLD.jsonb->'temporaryLocationId', OLD.jsonb->'permanentLocationId');
     new_effective_location_id = coalesce(NEW.jsonb->'temporaryLocationId', NEW.jsonb->'permanentLocationId');
-    -- null-safe comparison
+
+    -- null-safe comparison, do nothing if location is not changed.
     if (new_effective_location_id IS NOT DISTINCT FROM old_effective_location_id) THEN
       RETURN NEW;
     end if;
-    if (new_effective_location_id IS NULL) then
-      -- empty string signals the item trigger to remove the effective location without running a holdings query
-      new_effective_location_id = to_jsonb(''::text);
-    end if;
+
     UPDATE ${myuniversity}_${mymodule}.item
       SET jsonb = jsonb_set(jsonb, '{effectiveLocationId}', new_effective_location_id)
       WHERE holdingsRecordId = new.id
+            -- if item's either temp or perm location is not null then we don't need to update it's effective location
+            -- as item's location has more priority than holding's one.
             AND permanentLocationId IS NULL
             AND temporaryLocationId IS NULL;
-    -- if item's either temp or perm location is not null then we don't need to update it's effective location
-    -- as item's location has more priority than holding's one.
     RETURN NEW;
   END;
   $$ LANGUAGE 'plpgsql';
@@ -34,22 +32,23 @@ AS $$
   declare
     effective_location_id jsonb;
   begin
+    -- If location attributes set on item - use tem as they have higher priority
     effective_location_id = coalesce(NEW.jsonb->'temporaryLocationId', NEW.jsonb->'permanentLocationId');
     if (effective_location_id IS NOT NULL) then
       NEW.jsonb = jsonb_set(NEW.jsonb, '{effectiveLocationId}', effective_location_id);
       return NEW;
     end if;
+
+    -- If 'effectiveLocationId' is present than it was set by holdings trigger,
+    -- so we're skipping the main logic.
     effective_location_id = NEW.jsonb->'effectiveLocationId';
     if (effective_location_id IS NOT NULL) then
       -- Field exists so we know it was set by the holdings trigger
       -- because RMB always removes this read-only field.
-      if (effective_location_id ? '') then
-        -- Holdings trigger uses empty string to signal removal, so there is no need to run a holdings query.
-        -- "effectiveLocationId" must be a valid UUID or removed at all, null or empty string is not allowed.
-        NEW.jsonb = NEW.jsonb - 'effectiveLocationId';
-      end if;
       return NEW;
     end if;
+
+    -- If sensitive fields (holdingsRecordId, permanent/temporaryLocationId) have not been changed
     -- No change in item's sensitive fields?
     if (TG_OP = 'UPDATE'
         AND (OLD.jsonb->'holdingsRecordId' IS NOT DISTINCT FROM NEW.jsonb->'holdingsRecordId')
@@ -63,14 +62,16 @@ AS $$
       end if;
       return NEW;
     end if;
+
+    -- If item has holdingsRecordId then use location from holdings_record
     SELECT coalesce(jsonb->'temporaryLocationId', jsonb->'permanentLocationId')
       INTO effective_location_id
       FROM ${myuniversity}_${mymodule}.holdings_record
       WHERE id = (NEW.jsonb->>'holdingsRecordId')::uuid
       LIMIT 1;
-    if (effective_location_id IS NOT NULL) then
-      NEW.jsonb = jsonb_set(NEW.jsonb, '{effectiveLocationId}', effective_location_id);
-    end if;
+
+    NEW.jsonb = jsonb_set(NEW.jsonb, '{effectiveLocationId}', effective_location_id);
+
     return NEW;
   END;
 $$ LANGUAGE 'plpgsql';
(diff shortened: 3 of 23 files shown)


ramls/item.json
```diff
@@ -200,6 +200,11 @@
     "temporaryLocationId": {
       "type": "string"
     },
+    "effectiveLocationId": {
```
marcjohnson-kint: Given this is expected to be a UUID, even though it is read only, it could be useful to clients to be explicit about this by including the UUID pattern for this property.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -55,8 +60,20 @@ public void postItemStorageItems(
     if (entity.getStatus() == null) {
       entity.setStatus(new Status().withName(DEFAULT_STATUS_NAME));
     }
+
     PgUtil.post(ITEM_TABLE, entity, okapiHeaders, vertxContext,
-        PostItemStorageItemsResponse.class, asyncResultHandler);
+      PostItemStorageItemsResponse.class, postResponse -> {
+        // Have to re-read item to get calculated fields like effectiveLocationId
+        if (hasCreatedStatus(postResponse.result())) {
+          readItemById(entity.getId(), okapiHeaders, vertxContext)
```
marcjohnson-kint: How do you know there is a need to copy the headers?

Is this code based upon a similar example elsewhere? 

Should we request that the RAML Module Builder pgUtil responds with the record from the database rather than what was provided?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -810,6 +810,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "effectiveLocationId",
```
marcjohnson-kint: As of the beginning of Q4, RAML Module Builder based modules are now expected to provide from versions for database changes.

Please refer to [the docs](https://github.com/folio-org/raml-module-builder/blob/master/DB-schema-migration.md) for more guidance.

marcjohnson-kint: Oh, it appears that is already present in the scripts section. Is there a reason this wasn't included here?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
bohdan-suprun: @marcjohnson-kint,

> How do you know there is a need to copy the headers?
 PgUtil.post method sets Location header internally, just checked code of this method.

> Should we request that the RAML Module Builder pgUtil responds with the record from the database rather than what was provided?
It would be great I think.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: @marcjohnson-kint, 

As far as I see this property is not applicable for indexes.

ramls/item.json
bohdan-suprun: @marcjohnson-kint, thanks, fixed.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -810,6 +810,12 @@
...
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
```
julianladisch: There is no need to remove accents from an UUID. "removeAccents" should be false.

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
```diff
@@ -27,6 +27,11 @@
 
 
   protected static UUID createInstanceAndHolding(UUID holdingsPermanentLocationId)
+    throws ExecutionException, InterruptedException, MalformedURLException, TimeoutException {
+    return createInstanceAndHolding(holdingsPermanentLocationId, null);
+  }
+
+  protected static UUID createInstanceAndHolding(UUID holdingsPermanentLocationId, UUID holdingTempLocation)
```
julianladisch: For consistency: Can you change `holdingTempLocation` to `holdingTempLocationId`?

src/test/java/org/folio/rest/support/ResponseUtilTest.java
```diff
@@ -0,0 +1,75 @@
...
+
+  @Test
+  public void canReturnFalseIfNotCreatedStatus() {
+    Response response = Response.status(Response.Status.ACCEPTED).build();
+
+    assertFalse(ResponseUtil.hasCreatedStatus(response));
+  }
+
+  @Test
+  public void canNotInstantiateClass() throws Exception {
```
julianladisch: Can this unit test been extended and replaced by UtilityClassTester.assertUtilityClass? See StorageHelperTest for an example.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,376 @@
...
+  @Before
+  public void beforeEach() {
+    StorageTestSuite.deleteAll(itemsStorageUrl(""));
+    StorageTestSuite.deleteAll(holdingsStorageUrl(""));
+    StorageTestSuite.deleteAll(instancesStorageUrl(""));
+  }
+
+  @After
+  public void checkIdsAfterEach() {
+    StorageTestSuite.checkForMismatchedIDs("item");
```
julianladisch: We also have a holdings trigger, therefore we should also check for mismatchedIDs in "holdings_record".

bohdan-suprun: thanks @julianladisch, fixed.

src/test/java/org/folio/rest/support/ResponseUtilTest.java
bohdan-suprun: @julianladisch, thanks, it really useful.

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
bohdan-suprun: done

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: changed.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,377 @@
...
+        StorageTestSuite.TENANT_ID, createCompleted::complete);
+
+    HttpClientResponse response = createCompleted.get(5, TimeUnit.SECONDS);
+
+    assertThat(response.statusCode(), is(201));
+    assertThat(response.getHeader("location"), not(isEmptyString()));
+  }
+
+  private Item buildItem(UUID holdingsRecordId,
+                         UUID tempLocation, UUID permLocation) {
```
julianladisch: buildItem uses "temp, perm" as the parameter order whereas createInstanceAndHolding uses "perm, temp".
This is inconsistent and makes it difficult to understand the unit tests.
I suggest changing buildItem to also use "perm, temp".

ramls/item.json
```diff
@@ -200,6 +200,12 @@
...
+      "type": "string",
+      "description": "The effective location is used by FOLIO and other integrated systems to know the current home location for the item (read only, derived from locations on HoldingsRecord and Item)",
```
julianladisch: The beginning of the description is redundant.
The calculation should be added.
My suggestion:

    "The current home location for the item. Read only, it is automatically updated to be the first defined value from item.temporaryLocationId, item.permanentLocationId, holdingsRecord.temporaryLocationId, holdingsRecord.permanentLocationId."

bohdan-suprun: done

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: fixed, thanks.

julianladisch:
Important test cases are missing.
There are too many SELECTS in the triggers that should be avoided.
I will post more details later.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Hi @marcjohnson-kint, I finally realized how to use this property. We have to put this `fromModuleVersion` property  on table level and update it every time we make some changes to the table. 

So added this property to the item table.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -220,72 +182,135 @@ public void canCalculateEffectiveLocationOnHoldingUpdateWhenSomeItemsHasLocation
...
+    } else {
+      json.put("temporaryLocationId", location[temp].toString());
+    }
+  }
 
-    Item itemFetched = getItem(item.getId());
-    assertEquals(itemFetched.getEffectiveLocationId(), annexLibraryLocationId.toString());
+  private String effectiveLocation(int holdingPerm, int holdingTemp, int itemPerm, int itemTemp) {
+    int loc = 0;
+    if (itemTemp != 0) {
```
marcjohnson-kint: Is this determining the effective location within the tests?

I prefer that tests do not use conditional parts as it makes them harder to follow and understand, and hence less effective at describing the intended behaviour.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: @bohdan-suprun Thank you for investigating and working this out.

Is this described in the documentation? If not, please raise an issue for the documentation to be improved around this area.

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
```diff
@@ -0,0 +1,86 @@
...
+$$ LANGUAGE 'plpgsql';
+
+DROP TRIGGER IF EXISTS update_effective_location_for_items ON ${myuniversity}_${mymodule}.holdings_record;
+-- should be after update trigger, will allow item trigger to fetch up-to-date locations from holding
+create trigger update_effective_location_for_items after update
+on ${myuniversity}_${mymodule}.holdings_record for each row execute procedure ${myuniversity}_${mymodule}.update_effective_location_on_holding_update();
+
+
+DROP TRIGGER IF EXISTS update_effective_location ON ${myuniversity}_${mymodule}.item;
+create trigger update_effective_location before insert or update
```
marcjohnson-kint: Why is this defined as before insert or update and the holdings trigger defined as after update?

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
```diff
@@ -0,0 +1,86 @@
...
+      if (effective_location_id ? '') then
+        -- Holdings trigger uses empty string to signal removal, so there is no need to run a holdings query.
+        -- "effectiveLocationId" must be a valid UUID or removed at all, null or empty string is not allowed.
+        NEW.jsonb = NEW.jsonb - 'effectiveLocationId';
+      end if;
+      return NEW;
+    end if;
+    -- No change in item's sensitive fields?
+    if (TG_OP = 'UPDATE'
+        AND (OLD.jsonb->'holdingsRecordId' IS NOT DISTINCT FROM NEW.jsonb->'holdingsRecordId')
```
marcjohnson-kint: What does this check do? Does it mean that if an item moves holding this trigger won't be run?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: @marcjohnson-kint, yes this is described in the documentation: `
fromModuleVersion - this field indicates the version which the table was created/updated in. `

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
bohdan-suprun: We need to calculate the effectiveLocationId for the newly created items + update the property when one (or several) of the sensitive fields have been changed. 

And for holdings we have to recalculate item's effective location on update only. I've declared it as after update in order to allow item's trigger to read new values from holdings table (but I'm not sure if this still required after @julianladisch rework).

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
bohdan-suprun: @marcjohnson-kint the `IS NOT DISTINCT FROM` instructions is the same as `=` but it also null safe. 

It means when we use `=` operator to compare `NULL` with something we will always get `NULL` as the operation result and if statement won't trigger.

But  `IS NOT DISTINCT FROM` allow us to compare NULLs safely, so `NULL IS NOT DISTINCT FROM 'Hello'` will return False vs `NULL` when we use `=`.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -220,72 +182,135 @@ public void canCalculateEffectiveLocationOnHoldingUpdateWhenSomeItemsHasLocation
...
+      // { "effectiveLocationId": null } is not allowed, the property must have been removed
+      assertThat(item.containsKey("effectiveLocationId"), is(false));
+      return;
+    }
+    assertThat(item.getString("effectiveLocationId"), is(effectiveLocation));
+  }
 
-    assertEquals(getItem(item.getId()).getEffectiveLocationId(), onlineLocationId.toString());
+  @SuppressWarnings("unused")  // is actually used by @Parameters(method = "parameters")
+  private Object parameters() throws Exception {
```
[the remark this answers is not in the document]
bohdan-suprun: @marcjohnson-kint, will refactore it a bit if @julianladisch don't mind

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -220,72 +182,135 @@ public void canCalculateEffectiveLocationOnHoldingUpdateWhenSomeItemsHasLocation
...
+    JsonObject holding2 = holdingsClient.getById(holdingsRecordId).getJson();
+    setPermTempLocation(holding2, newHoldingPerm, newHoldingTemp);
+    holdingsClient.replace(holdingsRecordId, holding2);
+    JsonObject item3 = itemsClient.getById(itemId).getJson();
+    assertEffectiveLocation(item3, newHoldingPerm, newHoldingTemp, itemPerm, itemTemp);
   }
 
   @Test
   public void canSearchItemByEffectiveLocation() throws Exception {
+    StorageTestSuite.deleteAll(itemsStorageUrl(""));
```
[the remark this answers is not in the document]
julianladisch: This is the only test that requires all holdings and items been deleted.
All other tests don't.
There are two methods that run 112 times because they take the 112 data sets from the parameters method. In some former version there were many more data sets and they took 30 seconds.
Having deleteAll only in this method saves 2*112=224 delete requests per table.

ramls/item.json
```diff
@@ -200,6 +200,12 @@
     "temporaryLocationId": {
       "type": "string"
     },
+    "effectiveLocationId": {
+      "type": "string",
+      "description": "The current home location for the item. Read only, it is automatically updated to be the first defined value from item.temporaryLocationId, item.permanentLocationId, holdingsRecord.temporaryLocationId, holdingsRecord.permanentLocationId.",
```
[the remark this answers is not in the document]
bohdan-suprun: @marcjohnson-kint, you're right, changed it.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
bohdan-suprun: done

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: outdated

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,468 @@
...
+    if (tempLocationId != null) {
+      json.put(temporaryLocationIdFieldName, tempLocationId.toString());
+    }
+  }
+
+  private void assertEffectiveLocation(JsonObject item, UUID holdingPerm, UUID holdingTemp, UUID itemPerm, UUID itemTemp) {
+    UUID expectedEffectiveLocation = ObjectUtils
+      .firstNonNull(itemTemp, itemPerm, holdingTemp, holdingPerm);
+
+    if (expectedEffectiveLocation == null) {
```
marcjohnson-kint: Under what circumstances might this happen? Could it be better represented by explicitly expecting this during those tests?

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,468 @@
...
+    return locationsClient.getAll().stream()
+      .filter(json -> json.getString("name").equals(name))
+      .map(json -> UUID.fromString(json.getString("id")))
+      .findAny()
+      .orElse(null);
+  }
+
+  private void setPermanentTemporaryLocation(
+    JsonObject json, UUID permanentLocationId, UUID tempLocationId) {
+    String permanentLocationIdFieldName = "permanentLocationId";
```
marcjohnson-kint: Are these constants? If so, they should be marked final.

Personally, I might find it easier to read if there was whitespace between the signature and this first line.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,468 @@
...
+        .map(Item::getId)
+        .collect(Collectors.toList()),
+      hasItems(itemWithTempLocation.getId(), itemWithAllLocation.getId()));
+
+    assertEquals(1, secondFloorLibraryItems.getTotalRecords().intValue());
+    assertThat(secondFloorLibraryItems.getItems().get(0).getId(), is(itemWithPermLocation.getId()));
+  }
+
+  @Test
+  public void responseContainsAllRequiredHeaders() throws Exception {
```
marcjohnson-kint: Why do we need this test?

bohdan-suprun: @marcjohnson-kint, this test were created to cover post and then re-read item logic but since we are making changes to RMB, then it is redundant.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
julianladisch: Thanks for the suggestion, I've replaced it with PermTemp object that holds the the permanent location UUID and temporary location UUID of either a holding or an item.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,459 @@
...
+
+  private static String journalMaterialTypeID;
+  private static String canCirculateLoanTypeID;
+  private static UUID instanceId;
+  private static UUID mainLibraryLocationId;
+  private static UUID annexLibraryLocationId;
+  private static UUID onlineLocationId;
+  private static UUID secondFloorLocationId;
+
+  private static UUID loc1 = UUID.fromString("11111111-1111-4111-8111-111111111111");
```
bohdan-suprun: @julianladisch, why we can not use the locations above but creating this dummies instead?

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: @marcjohnson-kint, I'm not sure if this is a real case. UI says that `permanetLocationId` for a holding record is a required field, however API does not.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
bohdan-suprun: @julianladisch, please let me know in the future if you're working on something in the PR. So we won't make the changes in the same area. Thanks.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
julianladisch: Yes, we calculate the expected effective location in the test code.
It does not make sense to put a hardcoded list of all location string combinations and the expected effective location into the test class.
We want to test the triggers. The triggers have complicated code to avoid as many additional SELECT queries for performance reasons. We want to make sure that this complicated code comes to the correct result.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
marcjohnson-kint: I can understand wanting to reduce the amount of record clean up needed. Maybe this test could be moved to a test class better suited to that, away from the tests that need considerable record set up?

ramls/item.json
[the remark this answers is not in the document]
julianladisch: I strongly disagree with the removal of the rule description.
The temporary and permanent location of holding and item have exactly the meaning as described in the effectiveLocationId description: "first defined value from item.temporaryLocationId, item.permanentLocationId, holdingsRecord.temporaryLocationId, holdingsRecord.permanentLocationId."
It it very, very unlikely that the meaning of any of these five fields or the effective location rule change.
If they change, we for sure need an interface version bump.
Please re-add the rule description how the effective location is calculated. This is exactly the information that any user of the API needs to know.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,468 @@
...
+    holdingsClient.replace(holdingsRecordId, holding);
+
+    for (Item item : itemsToCreate) {
+      Item fetchedItem = getItem(item.getId());
+      assertEquals(fetchedItem.getEffectiveLocationId(), secondFloorLocationId.toString());
+    }
+  }
+
+  @Test
+  public void canCalculateEffectiveLocationOnHoldingRemoveTempLocation() throws Exception {
```
marcjohnson-kint: Why does this test need to involve multiple items?

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
julianladisch: The holdings trigger should be defined as an "after" trigger because it doesn't change the holdings record. It can wait until all checks and holdings "before" triggers have run and did not cancel the update.
The item trigger should be defined as an "before" trigger because it changes the item record by setting the effective location id. This should be done before the item is saved so that all item "after" triggers have access to this new effective value.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
marcjohnson-kint: Can these test names describe what effective location is being calculated, e.g. that it should be the permanent location from the holding.

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
```diff
@@ -0,0 +1,86 @@
...
+      RETURN NEW;
+    end if;
+    if (new_effective_location_id IS NULL) then
+      -- empty string signals the item trigger to remove the effective location without running a holdings query
+      new_effective_location_id = to_jsonb(''::text);
+    end if;
+    UPDATE ${myuniversity}_${mymodule}.item
+      SET jsonb = jsonb_set(jsonb, '{effectiveLocationId}', new_effective_location_id)
+      WHERE holdingsRecordId = new.id
+            AND permanentLocationId IS NULL
```
marcjohnson-kint: Why are checks for whether the item locations are null needed here?

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
julianladisch: If both the item's perm and temp locations are null before and after the update then the holding record determines the effective location. If the holdings record doesn't change then the effective location doesn't change and we can copy it from the old item record. This is what this if clause does.
If the item moves to a different holding then we need to query the new holding record and must skip this if clause.

marcjohnson-kint: Why is this section protected by a check for if the holdings record ID has not changed?

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
marcjohnson-kint: Thanks for the explanation @julianladisch

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
marcjohnson-kint: I think I'm failing to follow this logic.

Does this check say, only re-calculate the effective location if the holdings record ID has not changed and the item temporary and permanent location used to be null?

How does that match the intent expressed in the comment, that no relevant item properties have changed?

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
julianladisch: If all four locations (permanent and temporary of item and holding) are not defined then there is no effective location. This may happen because they are optional. An item that is ordered but not received (and therefore cannot been loaned) does not need any location.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
julianladisch: Fixed.

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
marcjohnson-kint: @julianladisch 

> If both the item's perm and temp locations are null before and after the update

Where in this check does it check that the new item permanent or temporary location is null?

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
marcjohnson-kint: @bohdan-suprun Thanks for the explanation

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,459 @@
...
+    assertEquals(itemNoLocationFetched.getEffectiveLocationId(), secondFloorLocationId.toString());
+
+    // Assert that itemWithPermLocationFetched was not updated
+    assertEquals(itemWithTempLocationFetched.getEffectiveLocationId(), annexLibraryLocationId.toString());
+  }
+
+  @Test
+  @Parameters(method = "parameters")
+  public void canCalculateEffecticeLocationOnItemUpdate(
+      PermTemp holdingLoc, PermTemp itemStartLoc, PermTemp itemEndLoc) throws Exception {
```
marcjohnson-kint: Is the `itemEndLoc` the expected effective location?

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
marcjohnson-kint: Out of interest, why do we need 224 tests to check this behaviour?

src/main/resources/templates/db_scripts/itemEffectiveLocation.sql
julianladisch: In line 38 it checks whether the new item permanent or temporary location is null. If it isn't null it will return in line 40.

julianladisch: The `if (TG_OP = 'UPDATE'` clause checks if there were no changes in the relevant item fields. If there were no changes then it uses the old effective location, no need to do any calculation.
Otherwise the execution continues after the if clause and runs a select query to fetch the holdings record and then calculates the effective location.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: @marcjohnson-kint, I think this is permanent and temp location to set on update either item or holding.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
julianladisch: These are not dummies.
`beforeAny()` will create fully-fledged locations using that location id.
`PermTemp.toString()` uses the first digit of the uuid therefore I use a predefined value and not a random like the other ones. These predefined numbers improve readability in `parameters()` and in IDE's JUnit view.
Note that JUnit runs `parameters()` *before* `beforeAny()`.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
julianladisch: Yes. This test method tests the item trigger by changing the item's permanent and temporary location to the values from `itemEndLoc`.
The expected effective location is calculated inside of `assertEffectiveLocation()`.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: done

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
bohdan-suprun: @julianladisch, I actually meant why we have separate locations for those two parametrized tests and can not reuse existing locations. I've already pushed some refactoring, so no prob on this.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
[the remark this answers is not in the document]
bohdan-suprun: @julianladisch are you going to move it to a separate test class?

src/main/resources/templates/db_scripts/populateEffectiveLocationForExistingItems.sql
```diff
@@ -0,0 +1,15 @@
+UPDATE ${myuniversity}_${mymodule}.item AS it
+SET jsonb = jsonb_set(jsonb, '{effectiveLocationId}',
+	CASE
+	  -- Check item's locations first
+		WHEN jsonb->'temporaryLocationId' IS NOT NULL THEN jsonb->'temporaryLocationId'
+		WHEN jsonb->'permanentLocationId' IS NOT NULL THEN jsonb->'permanentLocationId'
+		-- If no item's locations present - check holding's locations, and return an empty string by default
+		ELSE (
+			SELECT COALESCE(hr.jsonb->'temporaryLocationId', hr.jsonb->'permanentLocationId', to_jsonb(''::text))
```
mreno-EBSCO: Does this actually set `effectiveLocationId` to an empty string in the jsonb or will COALESCE interpret the empty string as "do not include this field" in the jsonb? If it is an empty string, I don't see how that can be validated with the `uuid` schema by a client. Either it should be a valid UUID or not set at all.

What is actually returned by the API in the default case? Perhaps RMB handles this, which, if it does, seems wrong to me.

julianladisch:
I'm busy today and can continue working on this on Friday.

src/main/resources/templates/db_scripts/populateEffectiveLocationForExistingItems.sql
bohdan-suprun: Hi @mreno-EBSCO, yes, you're completely right. It is fixed now.

src/main/resources/templates/db_scripts/populateEffectiveLocationForExistingItems.sql
```diff
@@ -0,0 +1,30 @@
...
+SET jsonb =
+	CASE
+		WHEN it.jsonb->'temporaryLocationId' IS NOT NULL
+			THEN jsonb_set(it.jsonb, '{effectiveLocationId}', it.jsonb->'temporaryLocationId')
+		WHEN it.jsonb->'permanentLocationId' IS NOT NULL
+			THEN jsonb_set(it.jsonb, '{effectiveLocationId}', it.jsonb->'permanentLocationId')
+		-- do nothing if all locations are nulls
+		ELSE it.jsonb
+	END
+WHERE it.jsonb->'effectiveLocationId' IS NULL AND it.holdingsrecordid IS NULL;
```
julianladisch: holdingsRecordId cannot be NULL because it is a required field: https://github.com/folio-org/mod-inventory-storage/blob/v17.0.0/ramls/item.json#L278
Therefore the "2nd step - update 'orphaned' items" should be removed.

julianladisch:
This pull request spends much effort in handling the case when the effective location is null. This cannot happen because holdingsRecord.permantLocationId should be a required field.
This is enforced in the front-end.
There is a pull request to also enforce this on the back-end:
https://github.com/folio-org/mod-inventory-storage/pull/338
I suggest to block this pull request until the other pull request has been merged. That will make this pull request much simpler.

marcjohnson-kint:
Separate to the question about whether the breaking change should go in before this. 

Is this change waiting for an update to RAML Module Builder in order to avoid having to re-read the record in order to provide a response that includes the derived effective location property?

bohdan-suprun:
> Separate to the question about whether the breaking change should go in before this.
> 
> Is this change waiting for an update to RAML Module Builder in order to avoid having to re-read the record in order to provide a response that includes the derived effective location property?

@marcjohnson-kint, Yep, @julianladisch has created RMB-476 however it is still in open status.

bohdan-suprun:
Hi @marcjohnson-kint, @julianladisch, can I clarify the next steps on this PR?

I think we're waiting for following:
* RMB-476 - to return entity from DB:
   * Then I will have to clean-up code and remove unnecessary logic;
* PR #338 so the `holdingsRecordId` property becomes mandatory for an item:
   * Clean-up triggers and migration script (probably check parametrized tests as well @julianladisch to confirm).

Have I missed something?

Please confirm.

marcjohnson-kint:
> Hi @marcjohnson-kint, @julianladisch, can I clarify the next steps on this PR?
> 
> I think we're waiting for following:
> 
> * RMB-476 - to return entity from DB:
>   
>   * Then I will have to clean-up code and remove unnecessary logic;
> * PR #338 so the `holdingsRecordId` property becomes mandatory for an item:
>   
>   * Clean-up triggers and migration script (probably check parametrized tests as well @julianladisch to confirm).
> 
> Have I missed something?
> 
> Please confirm.

I don't know if we are blocked on those. I think we need @julianladisch opinion. I haven't received any comments from my questions on the issue.

src/test/java/org/folio/rest/api/ItemEffectiveLocationTest.java
```diff
@@ -0,0 +1,352 @@
...
+    /**
+     * @param perm permanent location UUID
+     * @param temp temporary location UUID
+     */
+    PermTemp(UUID perm, UUID temp) {
+      this.perm = perm;
+      this.temp = temp;
+    }
+    /**
+     * The first character of the perm UUID and of the temp UUID; use _ for null.
```
julianladisch: The code has been changed to use locationIdToNameMap. Please update the first javadoc sentence accordingly.

julianladisch:
RMB-476 is not a blocker for this issue; the current insert+select can is very simple code that can be merged as is. When a new raml-module-builder version is available that contains RMB-476 then mod-inventory-storage can replace the insert+select code by the RMB-476 PgUtil code.

For me, PR #338 = https://issues.folio.org/browse/MODINVSTOR-364 (making holdingsRecord.permanentLocationId a required field) is a blocker for this issue because it will reduce the code and complexity by more than 50%. The code and logic of this issue #327 is very complex and there is at least one border case that is not handled correctly, but I don't want to spend time and effort in reviews and fixes for code that will be removed/simplified shortly. We already know that the performance of holdings and items bulk loading is an issue, therefore the triggers implemented by this issue must be both fast and correct.

bohdan-suprun:
Hi @julianladisch, @marcjohnson-kint, https://github.com/folio-org/mod-inventory-storage/pull/338 is merged now, so I've removed logic in the triggers that deals with nulls. Please review.

src/main/resources/templates/db_scripts/populateEffectiveLocationForExistingItems.sql
```diff
@@ -0,0 +1,9 @@
+UPDATE ${myuniversity}_${mymodule}.item AS it
+-- Since holdings_record.permanentLocationId and item.holdingsRecordId are required there can not be a NULL value
+SET	jsonb = JSONB_SET(it.jsonb,
+			'{effectiveLocationId}',
+			COALESCE(it.jsonb->'temporaryLocationId', it.jsonb->'permanentLocationId',
+			hr.jsonb->'temporaryLocationId', hr.jsonb->'permanentLocationId')
+)
+FROM ${myuniversity}_${mymodule}.holdings_record AS hr
+WHERE it.jsonb->'effectiveLocationId' IS NULL AND hr.id = it.holdingsrecordid;
```
julianladisch: I've added unit tests for populateEffectiveLocationForExistingItems.sql. Two of them fail. Please fix.

bohdan-suprun: Hi @julianladisch, thanks! I don't think that both tests are correct. 

The migration script will be executed only when effectiveLocationId is not set for an item, otherwise we assume that the item already got handled by triggers, so nothing to do with it.

Am I missing something?

julianladisch:
Bulk load scripts may temporarly disable triggers.
Changing from 17.1.0 back to 17.0.0, changing data, and then again switch to 17.1.0.

bohdan-suprun:
> Bulk load scripts may temporarly disable triggers.
Changing from 17.1.0 back to 17.0.0, changing data, and then again switch to 17.1.0.

Does it means that the migration script have to be executed on each item does not matter whether it already has `effectiveLocationId` specified or not?

marcjohnson-kint:
> > Bulk load scripts may temporarly disable triggers.
> > Changing from 17.1.0 back to 17.0.0, changing data, and then again switch to 17.1.0.
> 
> @julianladisch, Does it means that the migration script have to be executed on each item ant no matter whether it already has `effectiveLocationId` specified or not?

@skoczko Are we intending FOLIO's initial migration scripts to support the kinds of activities that @julianladisch describes above e.g. downgrades and compensating for migration / bulk updates being made directly in the database or with configuration changed?

julianladisch:
> Does it means that the migration script have to be executed on each item and no matter whether it already has `effectiveLocationId` specified or not?

Yes. We cannot assume that a record from a version before 17.1.0 has the correct value. Therefore "WHERE it.jsonb->'effectiveLocationId' IS NULL" should be removed from populateEffectiveLocationForExistingItems.sql to let the script overwrite a wrong value with the correct one.

julianladisch:
@marcjohnson-kint  If there really is the requirement that we want to keep a wrong `effectiveLocationId` value and the migration script should not overwrite the wrong value with the correct value by skipping them then there should be a big warning in the documentation.

bohdan-suprun:
> Yes. We cannot assume that a record from a version before 17.1.0 has the correct value. Therefore "WHERE it.jsonb->'effectiveLocationId' IS NULL" should be removed from populateEffectiveLocationForExistingItems.sql to let the script overwrite a wrong value with the correct one.

Well, there was no  `effectiveLocationId` property before `17.1.0`.  It is fully developed from the scratch in this PR. Does it change things?

marcjohnson-kint:
> @marcjohnson-kint If there really is the requirement that we want to keep a wrong `effectiveLocationId` value and the migration script should not overwrite the wrong value with the correct value by skipping them then there should be a big warning in the documentation.

@julianladisch I’m not suggesting it isn’t important to have the correct value. I’m more interested in the general policy of supporting upgrading then downgrading, then upgrading again.

Removing the null check is fine by me because it will be the same for the regular upgrade (as it will be null anyway).

At some point, as I think you mentioned previously, we need to think about the performance of latency of this process.

bohdan-suprun:
@julianladisch, @marcjohnson-kint, the `WHERE it.jsonb->'effectiveLocationId' IS NULL` is removed now, unit tests are fixed.

PR #330, opened 2019-09-17
Add missing classification types UIIN-505, MODINVSTOR-351

nielserik:


nielserik:
@bohdan-suprun:  the unit tests for reference tables generally sets a minimum number lower than the current number of reference records in each table. This was intended to allow for removing reference records -- within reasonably limits -- without having to change the tests.

bohdan-suprun:
> @bohdan-suprun: the unit tests for reference tables generally set a minimum number lower than the current number of reference records in each table. This was intended to allow for removing reference records -- within reasonable limits -- without having to change the tests.

@nielserik, ok, that's fine then.

PR #332, opened 2019-09-20
MODINVSTOR-203: Add some item.json descriptions

hemmefelix:


marcjohnson-kint:
Where did these descriptions come from? Are they from the appropriate SIG?

PR #334, opened 2019-09-20
MODINVSTOR-203 Add description to uuid.json.

julianladisch:


ramls/uuid.json
```diff
@@ -1,4 +1,5 @@
 {
+  "description": "A universally unique identifier (UUID), this is a 128-bit number used to identify a record and is shown in hex with dashes, for example 6312d172-f0cf-40f6-b27d-9fa8feaf332f; the UUID version must be from 1-5; see https://dev.folio.org/guides/uuids/",
```
marcjohnson-kint: Given that this is used as a fragment to be referred to from a property definition, what is the impact of this description?

If the property does not define a (context specific) description, does it use this? And hence, would stop the detection of properties that do not have a description?

julianladisch: A description in the $ref referenced file does not overwrite the description in the referencing file in the API foliodocs. Example in servicepoint.json:
"holdShelfExpiryPeriod" with "$ref": "time-period.json":
https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/service-point.html#service_points__servicepointid__put

When deleting the line

    "description": "expiration period for items on the hold shelf at the service point"

lint_raml_cop.py reports

    ERROR: lint-raml-schema: servicepoint.json: Missing "description" for: holdShelfExpiryPeriod

because the description in time-period.json is not sufficient.

marcjohnson-kint: @julianladisch Thanks for demonstrating that for me

PR #336, opened 2019-09-24
MODINVSTOR-203: Add descriptions for servicepointsuser.json.

julianladisch:


marcjohnson-kint:
Where did these descriptions come from? Are they from the appropriate SIG?

PR #339, opened 2019-09-30
MODINVSTOR-271: Unit tests can fail on ref constraint violations.

julianladisch:
Reusing the completed Future for running the second post tenant
starts the second post tenant that inserts all the sample records
but because of the completed Future it doesn't wait until they
are inserted but immediately returns to start running the other tests.

The other tests fail because records are inserted in parallel.

Use a new (uncompleted) Future to wait until the second post tenant
has completed before starting the other unit tests.

### DIFF
--- a/src/test/java/org/folio/rest/api/StorageTestSuite.java
+++ b/src/test/java/org/folio/rest/api/StorageTestSuite.java
@@ -268,6 +268,7 @@ public class StorageTestSuite {
     assertThat(failureMessage,
       response.getStatusCode(), is(201));
 
+    tenantPrepared = new CompletableFuture<>();
     jo.put("module_to", "mod-inventory-storage-1.1.0");
     jo.put("module_from", "mod-inventory-storage-1.0.0");
     client.post(storageUrl("/_/tenant"), jo, tenantId,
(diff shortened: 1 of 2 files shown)


src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -268,6 +268,7 @@ static void prepareTenant(String tenantId, boolean loadSample)
     assertThat(failureMessage,
       response.getStatusCode(), is(201));
 
+    tenantPrepared = new CompletableFuture<>();
```
marcjohnson-kint: Thank you for spotting this, it might answer the issues that @nielserik was having a while ago with inconsistencies in loading records.

Rather than declare a new future for the same variable, could we instead introduce a new variable for the second step, maybe called {{tenantUpgraded}}, as that seems to reflect the intent of what is happening?

marcjohnson-kint: @julianladisch What do you think about this proposal?

marcjohnson-kint:
@julianladisch This change appears to have broken the tests (or revealed a bug that was hidden), please investigate and fix prior to merging this change.

PR #343, opened 2019-10-11
[MODINVSTOR 357] Store effective call number on item record

jsavell:
Adds effectiveCallNumber property to Item and logic to derive its value on CREATE and PUT.

effectiveCallNumber will be set to:

- itemLevelCallNumber, if it is present on the Item

- If itemLevelCallNumber is not present and the Item has a holdingsRecordId, the matching holdingsRecord will be retrieved and effectiveCallNumber will be set to its value.

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -51,10 +51,15 @@
       "type": "string",
       "description": "Identifies the source of the call number, e.g., LCC, Dewey, NLM, etc."
     },
-    "effectiveCallNumber": {
-      "type": "string",
-      "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item.",
-      "readonly": true
+    "effectiveCallNumberComponents": {
+      "type": "object",
+      "description": "Elements of a full call number generated from the item or holding",
+      "properties": {
+        "callNumber": {
+          "type": "string",
+          "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+        }
+      }
     },
     "volume": {
       "type": "string",
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -53,7 +53,8 @@
     },
     "effectiveCallNumber": {
       "type": "string",
-      "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+      "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item.",
+      "readonly": true
     },
     "volume": {
       "type": "string",
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -51,6 +51,10 @@
       "type": "string",
       "description": "Identifies the source of the call number, e.g., LCC, Dewey, NLM, etc."
     },
+    "effectiveCallNumber": {
+      "type": "string",
+      "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+    },
     "volume": {
       "type": "string",
       "description": "Volume is intended for monographs when a multipart monograph (e.g. a biography of George Bernard Shaw in three volumes)."
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -4,7 +4,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicInteger;
 
 import javax.ws.rs.core.Response;
 
@@ -443,7 +442,6 @@ public class HoldingsStorageAPI implements HoldingsStorage {
   private CompletableFuture<Void> updateEffectiveCallNumbers(Items items, HoldingsRecord holdingsRecord, Map<String, String> okapiHeaders,
       Context vertexContext) {
     CompletableFuture<Void> setEffectiveCallNumberFuture = new CompletableFuture<>();
-    AtomicInteger itemCount = new AtomicInteger();
     items.getItems().forEach(item -> {
       String updatedCallNumner = null;
       if (StringUtils.isNotBlank(item.getItemLevelCallNumber())) {
@@ -457,10 +455,8 @@ public class HoldingsStorageAPI implements HoldingsStorage {
         PgUtil.put(ITEM_TABLE, item, item.getId(), okapiHeaders, vertexContext,
           PutItemStorageItemsByItemIdResponse.class, response -> {});
       }
-      if (itemCount.incrementAndGet() == items.getItems().size()) {
-        setEffectiveCallNumberFuture.complete(null);
-      }
     });
+    setEffectiveCallNumberFuture.complete(null);
     return setEffectiveCallNumberFuture;
   }
 }
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -1,10 +1,8 @@
 package org.folio.rest.impl;
 
-import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -30,6 +28,7 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.Promise;
 import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import io.vertx.ext.sql.SQLConnection;
@@ -335,36 +334,45 @@ public class HoldingsStorageAPI implements HoldingsStorage {
                 if (holdingsList.size() == 1) {
                   try {
                     postgresClient.startTx(connection -> {
-                      updateItemEffectiveCallNumbersByHoldings(connection, postgresClient, entity).thenAccept(voidResult -> {
-                        postgresClient.update(connection, HOLDINGS_RECORD_TABLE, entity, null, false,
-                          update -> {
-                            try {
-                              if (update.succeeded()) {
-                                postgresClient.endTx(connection, done -> {
-                                  asyncResultHandler.handle(
-                                    Future.succeededFuture(
-                                      PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
-                                        .respond204()));
-                                });
-                              }
-                              else {
+                      updateItemEffectiveCallNumbersByHoldings(connection, postgresClient, entity).setHandler(updateResult -> {
+                        if (updateResult.succeeded()) {
+                          postgresClient.update(connection, HOLDINGS_RECORD_TABLE, entity, null, false,
+                            update -> {
+                              try {
+                                if (update.succeeded()) {
+                                  postgresClient.endTx(connection, done -> {
+                                    asyncResultHandler.handle(
+                                      Future.succeededFuture(
+                                        PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
+                                          .respond204()));
+                                  });
+                                }
+                                else {
+                                  postgresClient.rollbackTx(connection, rollback -> {
+                                    asyncResultHandler.handle(
+                                      Future.succeededFuture(
+                                        PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
+                                          .respond500WithTextPlain(
+                                            update.cause().getMessage())));
+                                  });
+                                }
+                              } catch (Exception e) {
                                 postgresClient.rollbackTx(connection, rollback -> {
                                   asyncResultHandler.handle(
                                     Future.succeededFuture(
                                       PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
-                                        .respond500WithTextPlain(
-                                          update.cause().getMessage())));
+                                        .respond500WithTextPlain(e.getMessage())));
                                 });
                               }
-                            } catch (Exception e) {
-                              postgresClient.rollbackTx(connection, rollback -> {
-                                asyncResultHandler.handle(
-                                  Future.succeededFuture(
-                                    PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
-                                      .respond500WithTextPlain(e.getMessage())));
-                              });
-                            }
+                            });
+                        } else {
+                          postgresClient.rollbackTx(connection, rollback -> {
+                            Future.succeededFuture(
+                              PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
+                                .respond500WithTextPlain(
+                                  updateResult.cause().getMessage()));
                           });
+                        }
                       });
                     });
                   } catch (Exception e) {
@@ -433,21 +441,19 @@ public class HoldingsStorageAPI implements HoldingsStorage {
     }
   }
 
-  private CompletableFuture<Void> updateItemEffectiveCallNumbersByHoldings(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, HoldingsRecord holdingsRecord) {
-    CompletableFuture<Void> future = new CompletableFuture<>();
+  private Future<UpdateResult> updateItemEffectiveCallNumbersByHoldings(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, HoldingsRecord holdingsRecord) {
+    Future<UpdateResult> future = Future.future();
     Criterion criterion = new Criterion(
       new Criteria().addField("holdingsRecordId")
         .setJSONB(false).setOperation("=").setVal(holdingsRecord.getId()));
     postgresClient.get(ITEM_TABLE, Item.class, criterion, false, false, response -> {
-      updateEffectiveCallNumbers(connection, postgresClient, response.result().getResults(), holdingsRecord).thenAccept(voidResult -> {
-        future.complete(null);
-      });
+      updateEffectiveCallNumbers(connection, postgresClient, response.result().getResults(), holdingsRecord).future().setHandler(future);
     });
     return future;
   }
 
-  private CompletableFuture<Void> updateEffectiveCallNumbers(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, List<Item> items, HoldingsRecord holdingsRecord) {
-    CompletableFuture<Void> allItemsUpdated = new CompletableFuture<>();
+  private Promise<UpdateResult> updateEffectiveCallNumbers(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, List<Item> items, HoldingsRecord holdingsRecord) {
+    Promise<UpdateResult> allItemsUpdated = Promise.promise();
     List<Function<SQLConnection, Future<UpdateResult>>> batchFactories = items
       .stream()
       .map(item -> updateItemEffectiveCallNumber(item, holdingsRecord))
@@ -460,13 +466,7 @@ public class HoldingsStorageAPI implements HoldingsStorage {
       lastUpdate = lastUpdate.compose(prev -> factory.apply(connectionResult));
     }
 
-    lastUpdate.setHandler(updateResult -> {
-      if (updateResult.succeeded()) {
-        allItemsUpdated.complete(null);
-      } else {
-        allItemsUpdated.completeExceptionally(updateResult.cause());
-      }
-    });
+    lastUpdate.setHandler(allItemsUpdated);
     return allItemsUpdated;
   }
 
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -427,7 +427,7 @@ public class HoldingsStorageAPI implements HoldingsStorage {
   }
 
   private CompletableFuture<Void> updateItemEffectiveCallNumbersByHoldings(HoldingsRecord holdingsRecord, Map<String, String> okapiHeaders, Context vertexContext) {
-    CompletableFuture<Void> future = new CompletableFuture<Void>();
+    CompletableFuture<Void> future = new CompletableFuture<>();
     String query = String.format("holdingsRecordId==%s", holdingsRecord.getId());
     PgUtil.get(ITEM_TABLE, Item.class, Items.class, query, NO_OFFSET, NO_LIMIT, okapiHeaders, vertexContext, GetItemStorageItemsResponse.class, response -> {
       if (response.succeeded() && response.result() != null && response.result().getStatus() == 200) {
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -4,6 +4,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import javax.ws.rs.core.Response;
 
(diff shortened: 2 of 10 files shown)


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -269,6 +269,165 @@ public void canCreateAnItemWithoutProvidingID()
...
+    JsonObject itemToCreate = new JsonObject();
+    String itemId = UUID.randomUUID().toString();
+    itemToCreate.put("id", itemId);
+    itemToCreate.put("holdingsRecordId", holdingsRecordId.toString());
+    itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
+    itemToCreate.put("materialTypeId", bookMaterialTypeID);
+    createItem(itemToCreate);
+
+    CompletableFuture<Response> completed = new CompletableFuture<>();
+    client.put(itemsStorageUrl("/" + itemId), itemToCreate, StorageTestSuite.TENANT_ID,
```
marcjohnson-kint: This appears to update the item straight after it is created with no changes to either the item or the holding.

Can you help me understand the scenario this is trying to demonstrate?

marcjohnson-kint:
This change appears to only address deriving the effective call number when the item is created or changed, what about when the holdings record is changed?

src/test/java/org/folio/rest/api/ItemStorageTest.java
jsavell: This is a test to show that when we create an item and that item does not have its own itemLevelCallNumber, that the holding's callNumber will be set as the item's effectiveCallNumber during item creation.

Line 381 creates a holding with callNumber 'testCallNumber'.
Line 385 associates the holding with the soon to be created item.
Line 400 verifies that 'testCallNumber' has been set as the item's effectiveCallNumber.

marcjohnson-kint: The name of this test, canUpdateAnItemWithHoldingCallNumberAsEffectiveCallNumber, suggests it has something to do with the derivation of the effective call number after the item is updated.

Your description suggests it is purely about item creation.

These seem like different situations to me, have I misunderstood?

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -323,9 +328,10 @@ public void putHoldingsStorageHoldingsByHoldingsRecordId(
           postgresClient.get(HOLDINGS_RECORD_TABLE, HoldingsRecord.class, fieldList, cql, true, false,
             reply -> {
               if(reply.succeeded()) {
-                List<HoldingsRecord> itemList = reply.result().getResults();
+                List<HoldingsRecord> holdingsList = reply.result().getResults();
 
-                if (itemList.size() == 1) {
+                if (holdingsList.size() == 1) {
+                  updateItemEffectiveCallNumbersByHoldings(entity, okapiHeaders, vertxContext);
```
marcjohnson-kint: Should the update of the items happen before the holdings record update has been attempted?

Are these performed in a database transaction?

ramls/item.json
```diff
@@ -51,6 +51,10 @@
       "type": "string",
       "description": "Identifies the source of the call number, e.g., LCC, Dewey, NLM, etc."
     },
+    "effectiveCallNumber": {
+      "type": "string",
+      "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+    },
```
marcjohnson-kint: As this is a derived property it should be marked as read only in the JSON schema, please update this definition

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -577,6 +614,82 @@ public void tenantIsRequiredForGettingAllHoldings()
   }
 
   @Test
+  public void updatingHoldingsUpdatesItemEffectiveCallNumber()
```
marcjohnson-kint: I think it could be useful to have another test that demonstrates what happens when the call number is removed from the holdings record.

I think it could also be worth having a test that demonstrates that multiple items get updated when a holdings record changes, this could be an extension of this test or a new test.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -408,4 +414,31 @@ private boolean isUUID(String id) {
...
+    });
+  }
+
+  private void updateEffectiveCallNumbers(Items items, HoldingsRecord holdingsRecord, Map<String, String> okapiHeaders,
+      Context vertexContext) {
+    items.getItems().forEach(i -> {
+      String updatedCallNumner = null;
+      if (i.getItemLevelCallNumber() != null && !i.getItemLevelCallNumber().isEmpty()) {
+          updatedCallNumner = i.getItemLevelCallNumber();
+      } else if (i.getHoldingsRecordId() != null && !i.getHoldingsRecordId().isEmpty()) {
```
bohdan-suprun: Is it fine that we are checking that `i.getHoldingsRecordId() != null` but accessing to the `holdingsRecord`

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -117,9 +123,10 @@ public void putItemStorageItemsByItemId(
       String itemId, String lang, Item entity, java.util.Map<String, String> okapiHeaders,
       io.vertx.core.Handler<io.vertx.core.AsyncResult<Response>> asyncResultHandler,
       Context vertxContext) {
-
-    PgUtil.put(ITEM_TABLE, entity, itemId, okapiHeaders, vertxContext,
+    setEffectiveCallNumber(okapiHeaders, vertxContext, entity).thenAccept(i -> {
```
bohdan-suprun:
```suggestion
    setEffectiveCallNumber(okapiHeaders, vertxContext, entity).thenAccept(item -> {
```

ramls/item.json
```diff
@@ -51,6 +51,10 @@
       "type": "string",
       "description": "Identifies the source of the call number, e.g., LCC, Dewey, NLM, etc."
     },
+    "effectiveCallNumber": {
```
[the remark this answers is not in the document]
rladdusaw: The last release (17.0.0) has the item-storage version at 7.5, and it has already been bumped to 7.6 so I think it should be OK.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
rladdusaw: I've changed the holdings update to a transaction and moved the update of effective call numbers so it will only happen after the holdings are updated.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
rladdusaw: This should have been checking the holding record's call number. I've updated the branch to correct this.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
rladdusaw: I've gotten everything working and updated the test to cover more than one item. Unfortunately I nothing happens when the call number is removed from a holdings, would you prefer I add it to this PR or should we open a new issue?

rladdusaw: I've updated the first test to include multiple items on a holdings and created a new test for when the call number is removed from the holdings.

ramls/item.json
```diff
@@ -51,6 +51,11 @@
...
```
marcjohnson-kint: @mreno-EBSCO Raised the question yesterday in refinement. 

Do we want to create some form of grouping object for all of the effective / derived call number parts?

If we do, what would it be called?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
marcjohnson-kint: @rladdusaw Thanks

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -577,20 +577,158 @@ public void tenantIsRequiredForGettingAllHoldings()
...
+    Response getFirstUpdatedItemResponse = get(getFirstItemUrl);
+
+    JsonObject firstUpdatedItemFromGet = getFirstUpdatedItemResponse.getJson();
+
+    assertThat(getFirstUpdatedItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(firstUpdatedItemFromGet.getString("id"), is(firstItemId));
+    assertThat(firstUpdatedItemFromGet.getString("effectiveCallNumber"), is(""));
+  }
+
   public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
```
bohdan-suprun: May I ask you to add a `@Test` annotation for this test method? Seems it disappeared.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -133,6 +140,36 @@ public void deleteItemStorageItemsByItemId(
...
+  private CompletableFuture<Item> setEffectiveCallNumber(Map<String, String> okapiHeaders, Context vertxContext, Item item) {
+    CompletableFuture<Item> completableFuture = null;
+    if (item.getItemLevelCallNumber() != null && !item.getItemLevelCallNumber().isEmpty()) {
+      item.setEffectiveCallNumber(item.getItemLevelCallNumber());
+      completableFuture = CompletableFuture.supplyAsync(() -> item);
+    } else {
+      if (item.getHoldingsRecordId() != null && !item.getHoldingsRecordId().isEmpty()) {
+        completableFuture = getHoldingsRecordById(okapiHeaders, vertxContext, item.getHoldingsRecordId()).thenCombineAsync(CompletableFuture.supplyAsync(() -> item), (hr, i) ->
+        {
+          i.setEffectiveCallNumber(hr.getCallNumber());
```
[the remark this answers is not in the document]
rladdusaw: We use the same value to query for the full holdings record, which we shouldn't do if the id is null.

julianladisch:
Is there a reason that this is implemented in Java? We already have a highly optimized solution using triggers for the effective location than can easily been adopted for the effective call number: https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql

Using Java results in multiple round trips java <-> postgres when a holding has multiple items. The trigger solution doesn't have such round trips and therefore executes faster.

There are several cases where no update is needed, see the effective call number implementation. For example if the holding is updated but the holding call number doesn't change, only other fields change.

marcjohnson-kint:
> Is there a reason that this is implemented in Java? We already have a highly optimized solution using triggers for the effective location than can easily been adopted for the effective call number: https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/itemEffectiveLocation.sql

It was decided to use java for the call number record migration because [determining the sortable call number](https://issues.folio.org/browse/MODINVSTOR-381) requires the use of code in the MARC4J library and hence it made sense for the various aspects of call numbers to migrate using Java rather than a mixture of Java and PostgreSQL triggers.

> Using Java results in multiple round trips java <-> postgres when a holding has multiple items. The trigger solution doesn't have such round trips and therefore executes faster

Can the updates to items not be made in a single database operation, with multiple SQL statements, in a single transaction? 

> There are several cases where no update is needed, see the effective call number implementation. For example if the holding is updated but the holding call number doesn't change, only other fields change.

If you are suggesting that we should avoid unnecessary writes to items when a holding record is updated yet the relevant properties haven't changed, then I agree.

ramls/item.json
marcjohnson-kint: @mreno-EBSCO @bohdan-suprun @jsavell Do we want to try to group the derived call number properties together in an object or keep them separate at the top level of the item?

bohdan-suprun: I'd prefer to have them grouped.

bohdan-suprun:
> Can the updates to items not be made in a single database operation, with multiple SQL statements, in a single transaction?

One option, I think, can be in using an extended update statement:

```
UPDATE item AS it SET jsonb = jsonb_set(jsonb, '{effectiveCallNumber}', updateData.effectiveCallNumber::jsonb)
FROM (VALUES ('id1', 'call number-1'), ... ('idn', 'call number-n')) AS updateData(id, effectiveCallNumber)
WHERE it.id::TEXT = updateData.id;
```

`VALUES` here can be populated dynamically by Java, and then we will have to use `updateWithParams` method on connection, returned by `startTx` method. In this case, only one query will be sent for all the items.

But it sounds to me like a further improvement rather then approach which we need to implement in scope of this PR.

marcjohnson-kint:
> it sounds to me like a further improvement rather then approach which we need to implement in scope of this PR.

Agreed, I was responding to @julianladisch comment without context of the implementation. I hadn't realise it was using PgUtil to update each item, meaning there is network activity for each item, as that is hidden from the module.

src/main/resources/templates/db_scripts/holdingsRecordHrid.sql
```diff
@@ -20,7 +20,7 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.set_holdings_record_hrid(
 
 DROP TRIGGER IF EXISTS set_holdings_record_hrid ON ${myuniversity}_${mymodule}.holdings_record CASCADE;
 CREATE TRIGGER set_holdings_record_hrid
-  BEFORE INSERT ON ${myuniversity}_${mymodule}.holdings_record
+  AFTER INSERT ON ${myuniversity}_${mymodule}.holdings_record
```
bohdan-suprun: As far as I see, the trigger updates the NEW record. I'm not sure if it is going to work properly for AFTER triggers. Do we have a test for this?

mreno-EBSCO: I wouldn't spend much effort on this file or anything HRID related. It is being deleted per the work for MODINVSTOR-363.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -408,4 +431,60 @@ private boolean isUUID(String id) {
...
+      .collect(Collectors.toList());
+
+    SQLConnection connectionResult = connection.result();
+    Future<UpdateResult> lastUpdate = Future.succeededFuture();
+    for (Function<SQLConnection, Future<UpdateResult>> factory : batchFactories) {
+      lastUpdate = lastUpdate.compose(prev -> factory.apply(connectionResult));
+    }
+
+    lastUpdate.setHandler(updateResult -> {
+      if (updateResult.succeeded()) {
```
bohdan-suprun: I think we have to somehow notify calling method about failures (if any) so the transaction will be reverted if an item update has failed. You can use `Future` for this, and handle overall result in handler (`setHandler` method). 

Alternatively, you can use following algorithm:
1. Build a list of update factories (`updateSingleBatchFactory` method);
    1. The first part of the list will be items, that needs to be updated;
    1. The last element will be the holding itself.
1. Then, you will need a single handler for the last `Future` which will decide whether to commit or revert transaction.

The code can be something like:

```java
  private void updateHoldingAndItems(
    PostgresClient postgresClient, HoldingsRecord oldHolding, HoldingsRecord newHolding, 
    Handler<AsyncResult<Void>> handler) {

    if (Objects.equals(oldHolding.getCallNumber(), newHolding.getCallNumber())) {
      // Call number has not been changed
      // just do the holding simple update
      postgresClient.update(HOLDINGS_RECORD_TABLE, newHolding, newHolding.getId(), 
          result -> {/* handle result */});
      return;
    }

    // Call number has changed.
    final List<Item> itemsForHolding = getItemsForHolding(newHolding.getId());
    postgresClient.startTx(connectionResult -> {
      List<Function<SQLConnection, Future<UpdateResult>>> itemsWithNewCallNumber = ...
          /** update effective call number, 
            * remove those for which effective call number has not changed
            * wrap into the function
            */

      List<Function<SQLConnection, Future<UpdateResult>>> allUpdates = 
          new ArrayList<>(itemsWithNewCallNumber);
      allUpdates.add(updateSingleBatchFactory(HOLDINGS_RECORD_TABLE, newHolding.getId(), 
          newHolding, postgresClient));
      
      Future<UpdateResult> lastUpdate = Future.succeededFuture();
      for (Function<SQLConnection, Future<UpdateResult>> update : allUpdates) {
        lastUpdate = lastUpdate
          .compose(prev -> update.apply(connectionResult.result()));
      }
      
      lastUpdate.setHandler(updateResult -> {/* do commit or revert */ });
    });
  }
```
 
Then, you can execute this method in `putHoldingsStorageHoldingsByHoldingsRecordId`, handler will just return 204 or 500, as usual.
Note:  `oldHolding` above can be retrieved from the `holdingsList`, that we fetch from db in `putHoldingsStorageHoldingsByHoldingsRecordId`.

rladdusaw: Is it possible to start a new transaction within an existing transaction?

The alternate approach does seem to have some advantages over the current one, but I don't want to delay this PR any more than necessary. Would it be an acceptable compromise to create an issue with the alternate algorithm information that I could get back to if I have time after the call number prefix and suffix work is done?

mreno-EBSCO: I would think all you need to do is
```suggestion
      if (updateResult.succeeded()) {
        allItemsUpdated.complete(null);
      } else {
        allItemsUpdated.fail(updateResult.cause());
      }
```

Since `compose` will not call its mapper if the future it is called on has failed, the `lastUpdate` will be the failed future with its reason. You'd have to fail it up the chain since you are returning different futures in each method. I think that is a good idea anyway since I think the way it is currently implemented could lead to deadlock if there are failures (if updateResult did not succeed, allItemsUpdate is never completed).

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -408,4 +431,60 @@ private boolean isUUID(String id) {
...
+
+  private CompletableFuture<Void> updateEffectiveCallNumbers(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, List<Item> items, HoldingsRecord holdingsRecord) {
+    CompletableFuture<Void> allItemsUpdated = new CompletableFuture<>();
+    List<Function<SQLConnection, Future<UpdateResult>>> batchFactories = items
+      .stream()
+      .map(item -> updateItemEffectiveCallNumber(item, holdingsRecord))
+      .map(item -> updateSingleBatchFactory(ITEM_TABLE, item.getId(), item, postgresClient))
+      .collect(Collectors.toList());
+
+    SQLConnection connectionResult = connection.result();
```
mreno-EBSCO: Maybe don't unwrap this here so you can pass it to the apply as is and directly to the PostgresClient without having to rewrap it in a `succeededFuture` on use.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -408,4 +431,60 @@ private boolean isUUID(String id) {
       return false;
     }
   }
+
+  private CompletableFuture<Void> updateItemEffectiveCallNumbersByHoldings(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, HoldingsRecord holdingsRecord) {
+    CompletableFuture<Void> future = new CompletableFuture<>();
+    Criterion criterion = new Criterion(
+      new Criteria().addField("holdingsRecordId")
+        .setJSONB(false).setOperation("=").setVal(holdingsRecord.getId()));
+    postgresClient.get(ITEM_TABLE, Item.class, criterion, false, false, response -> {
```
mreno-EBSCO: I think this get should use the transaction connection. If it fails for some reason we need to roll back the transaction.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -408,4 +431,60 @@ private boolean isUUID(String id) {
       return false;
     }
   }
+
+  private CompletableFuture<Void> updateItemEffectiveCallNumbersByHoldings(AsyncResult<SQLConnection> connection, PostgresClient postgresClient, HoldingsRecord holdingsRecord) {
+    CompletableFuture<Void> future = new CompletableFuture<>();
```
mreno-EBSCO: Maybe use a Vert.x `Future` or, better yet, `Promise` here and elsewhere instead of introducing `CompletableFuture`. It appears all we need this for is indicating a PostgresClient operation is complete. Instead of `thenAccept`, we could use `setHandler`, since a Promise is already a handler, it can be passed directly to the `setHandler` method to complete it.

For example:
```java
  final Promise<Void> promise = Promise.promise();
  // stuff
  updateEffectiveCallNumbers(/*args*/).setHandler(promise);
  // stuff
  return promise.future(); // or just "return promise;"
```

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -316,33 +329,43 @@ public void putHoldingsStorageHoldingsByHoldingsRecordId(
...
-                                    update.cause().getMessage())));
-                          }
-                        } catch (Exception e) {
-                          asyncResultHandler.handle(
-                            Future.succeededFuture(
-                              PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
-                                .respond500WithTextPlain(e.getMessage())));
-                        }
+                    postgresClient.startTx(connection -> {
+                      updateItemEffectiveCallNumbersByHoldings(connection, postgresClient, entity).thenAccept(voidResult -> {
```
mreno-EBSCO: We need to rollback if `updateItemEffectiveCallNumbersByHoldings` fails. If it fails, the `thenAccept` consumer won't be called. If we continue to use `CompletableFuture`, we would need an `exceptionally` block that could roll back on error. If we use a Vert.x `Future`, we could use `setHandler` to determine if the call number update succeeded or failed and apply the `postgresClient` update or fail.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
bohdan-suprun: Sure, @rladdusaw, you can use another logic here. My main concern that we don't handle failures for item updates, we have to somehow propagate failure, so we will revert the transaction. 

As an option: change `allItemsUpdated` to `Future` and check whether it failed or completed in `updateItemEffectiveCallNumbersByHoldings` or in `putHoldingsStorageHoldingsByHoldingsRecordId` methods.

I've pushed a holdings update service https://github.com/folio-org/mod-inventory-storage/compare/MODINVSTOR-357-effective-call-number...MODINVSTOR-357-effective-call-number-service?expand=1, hope you find it useful.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
rladdusaw: I originally was using the connection, but the results weren't be supplied until it completed and it was the cause of a lot of timeouts. Since it's just a get it doesn't make any changes to roll back in the case of a failure so it should be safe.

rladdusaw:
@mreno-EBSCO I've converted the CompletableFutures to vert.x Futures and Promises and am now triggering a rollback when the Item update Future is unsuccessful. Does this approach seem acceptable?

ramls/item.json
rladdusaw: @mreno-EBSCO @bohdan-suprun @marc-johnson What should this grouped object look like?

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -36,7 +45,11 @@
   // Has to be lowercase because raml-module-builder uses case sensitive
   // lower case headers
   private static final String TENANT_HEADER = "x-okapi-tenant";
+  private static final String WHERE_CLAUSE = "WHERE id = '%s'";
   public static final String HOLDINGS_RECORD_TABLE = "holdings_record";
+  public static final String ITEM_TABLE = "item";
+  public static final int NO_OFFSET = 0;
```
mreno-EBSCO: Is this used anywhere? I don't see anything in the commit, but it is public, so maybe something else is using it. If it is being used, I am not sure this is the right place for this constant.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -36,7 +45,11 @@
...
+  public static final int NO_LIMIT = -1;
```
mreno-EBSCO: Is this used anywhere? I don't see anything in the commit, but it is public, so maybe something else is using it. If it is being used, I am not sure this is the right place for this constant.

rladdusaw: I stopped using this when I switched to a transaction. I've removed them.

ramls/item.json
```diff
@@ -51,6 +51,16 @@
       "type": "string",
       "description": "Identifies the source of the call number, e.g., LCC, Dewey, NLM, etc."
     },
+    "effectiveCallNumberComponents": {
```
mreno-EBSCO: Since this is generated on the backend for every POST/PUT and any value passed in is ignored, can this be marked `readonly`?

PR #345, opened 2019-10-23
MODINVSTOR-386: Add last checkin properties

mrdpopov:
# Purpose

Store the information about when, where and by whom an item was last checked in, even when there is no loan.
This data is to be displayed in the item details.

This resolves MODINVSTOR-386
To be used in UIIN-749

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -296,6 +296,45 @@
       "folio:linkFromField": "holdingsRecordId",
       "folio:linkToField": "id",
       "folio:includedElement": "holdingsRecords.0"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "readonly": true,
+      "additionalProperties": false,
+      "properties": {
+        "date": {
+          "type": "string",
+          "description": "Date and time of the last check in of the item."
+        },
+        "servicePointName": {
+          "type": "string",
+          "description": "Service point being used by user when item was scanned in Check in app."
+        },
+        "source": {
+          "type": "object",
+          "readonly": true,
+          "additionalProperties": false,
+          "description": "Lastname, first name middle name (with link to user profile) of user who scanned the item",
+          "properties": {
+            "id": {
+              "type": "string",
+              "description": "ID of user who scanned the item"
+            },
+            "firstName": {
+              "type": "string",
+              "description": "First name of user who scanned the item"
+            },
+            "middleName": {
+              "type": "string",
+              "description": "Middle name of user who scanned the item"
+            },
+            "lastName": {
+              "type": "string",
+              "description": "Last name of user who scanned the item"
+            }
+          }
+        }
+      }
     }
   },
   "additionalProperties": false,
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -311,12 +311,12 @@
         "servicePointId": {
           "type": "string",
           "description": "Service point ID being used by a staff member when item was scanned in Check in app.",
-          "pattern": "[A-F0-9]{8}-[A-F0-9]{4}-4[A-F0-9]{3}-[89AB][A-F0-9]{3}-[A-F0-9]{12}"
+          "$ref": "uuid.json"
         },
         "staffMemberId": {
           "type": "string",
           "description": "ID a staff member who scanned the item",
-          "pattern": "[A-F0-9]{8}-[A-F0-9]{4}-4[A-F0-9]{3}-[89AB][A-F0-9]{3}-[A-F0-9]{12}"
+          "$ref": "uuid.json"
         }
       }
     }
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -301,36 +301,38 @@
       "type": "object",
       "readonly": true,
       "additionalProperties": false,
+      "description": "Information about when an item was last scanned on the Inventory app.",
       "properties": {
-        "date": {
+        "dateTime": {
           "type": "string",
-          "description": "Date and time of the last check in of the item."
+          "description": "Date and time of the last check in of the item.",
+          "format": "date-time"
         },
         "servicePointName": {
           "type": "string",
-          "description": "Service point being used by user when item was scanned in Check in app."
+          "description": "Service point being used by a staff member when item was scanned in Check in app."
         },
         "source": {
           "type": "object",
-          "readonly": true,
           "additionalProperties": false,
-          "description": "Lastname, first name middle name (with link to user profile) of user who scanned the item",
+          "description": "Last name, first name middle name of a staff member who scanned the item",
           "properties": {
-            "id": {
+            "staffMemberId": {
               "type": "string",
-              "description": "ID of user who scanned the item"
+              "description": "ID a staff member who scanned the item",
+              "pattern": "[A-F0-9]{8}-[A-F0-9]{4}-4[A-F0-9]{3}-[89AB][A-F0-9]{3}-[A-F0-9]{12}"
             },
             "firstName": {
               "type": "string",
-              "description": "First name of user who scanned the item"
+              "description": "First name of a staff member who scanned the item"
             },
             "middleName": {
               "type": "string",
-              "description": "Middle name of user who scanned the item"
+              "description": "Middle name of a staff member who scanned the item"
             },
             "lastName": {
               "type": "string",
-              "description": "Last name of user who scanned the item"
+              "description": "Last name of a staff member who scanned the item"
             }
           }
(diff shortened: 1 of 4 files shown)


ramls/item.json
```diff
@@ -296,6 +296,45 @@
       "folio:linkFromField": "holdingsRecordId",
       "folio:linkToField": "id",
       "folio:includedElement": "holdingsRecords.0"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "readonly": true,
+      "additionalProperties": false,
+      "properties": {
+        "date": {
```
marcjohnson-kint: This appears to be a date and time. I'm aware that often those are named `date` rather than `date time` in FOLIO, that might be a topic to review at some point.

Could this definition be extended to include the appropriate format, to make it more explicit what is expected in this property.

ramls/item.json
```diff
@@ -296,6 +296,45 @@
       "folio:linkFromField": "holdingsRecordId",
       "folio:linkToField": "id",
       "folio:includedElement": "holdingsRecords.0"
+    },
+    "lastCheckIn": {
```
marcjohnson-kint: There needs to be a description for the object property as well as the member properties, please can one be added.

ramls/item.json
```diff
@@ -296,6 +296,45 @@
...
+          "type": "string",
+          "description": "Service point being used by user when item was scanned in Check in app."
+        },
+        "source": {
+          "type": "object",
+          "readonly": true,
+          "additionalProperties": false,
+          "description": "Lastname, first name middle name (with link to user profile) of user who scanned the item",
+          "properties": {
+            "id": {
```
marcjohnson-kint: The description suggests this is the `user ID` of the staff member or patron who performed the check in, could the name of this property be made more specific to reflect what kind of record this is intended to reference and what role they were performing.

marcjohnson-kint: As this is intended to be a UUID, please add a format to constrain the contents of this property to a UUID

ramls/item.json
```diff
@@ -296,6 +296,45 @@
...
+          "type": "string",
+          "description": "Date and time of the last check in of the item."
+        },
+        "servicePointName": {
+          "type": "string",
+          "description": "Service point being used by user when item was scanned in Check in app."
+        },
+        "source": {
+          "type": "object",
+          "readonly": true,
```
marcjohnson-kint: How are these properties read only? What part of the system will set them, inventory storage cannot know which user performed the check in?

mrdpopov:
I intended to create tests in mod-cirulation, where these new fields are going to be used. Do you think the tests are realy needed in this module?

ramls/item.json
mrdpopov: Done.

ramls/item.json
mrdpopov: Done.

ramls/item.json
mrdpopov: Done.

ramls/item.json
mrdpopov: Done.

ramls/item.json
```diff
@@ -296,6 +296,47 @@
       "folio:linkFromField": "holdingsRecordId",
       "folio:linkToField": "id",
       "folio:includedElement": "holdingsRecords.0"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "readonly": true,
```
mreno-EBSCO: Can a parent object be `readonly` and its children not be? Based on this schema, I don't understand how these properties are to be used in mod-inventory-storage.

mrdpopov: done

mrdpopov:
@marcjohnson-kint , @mreno-EBSCO Do you think this is ready to be merged?

ramls/item.json
```diff
@@ -296,6 +296,59 @@
...
+          "type": "string",
+          "description": "Date and time of the last check in of the item.",
+          "format": "date-time"
+        },
+        "servicePointInfo": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Service point being used by a staff member when item was scanned in Check in app.",
+          "properties": {
+            "id": {
```
bohdan-suprun: Do we need a foreign key for this property?

ramls/item.json
```diff
@@ -296,6 +296,59 @@
...
+              "type": "string",
+              "description": "Service point name being used by a staff member when item was scanned in Check in app."
+            }
+          }
+        },
+        "source": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Last name, first name middle name of a staff member who scanned the item",
+          "properties": {
```
bohdan-suprun: Hi @marcjohnson-kint, @mrdpopov, do we really need first/last/middle names to be set along with `staffMemberId`? 

I think, for similar cases we retrieving this additional info in business modules. For example `item.effective/permanent/tempLocation`s - we store only Id but then mod-inventory adds `name` attribute.

mrdpopov: Id is for the link to the user, other fields reduce the traffic in clients of this API.

bohdan-suprun: Ok. Does it make sense then to specify `staffMemberId` and `first/lastName` to be required fields? So there will be no situation when `staffMemberId` and names are out of sync (i.e. a client has updated id but names are still referencing the old user).

mreno-EBSCO:
I guess what I don't get is why we need store this data with inventory. Don't we have all the info we need in circulation already? I don't see what duping it here, albeit the last check-in only, buys us except more inventory state we need to manage in circulation.

mrdpopov:
Agree, I will explore saving this data in the circulation.

mrdpopov:
@mreno-EBSCO, @bohdan-suprun, @marcjohnson-kint Do you think this can be merged?

ramls/item.json
```diff
@@ -296,6 +296,29 @@
...
+      "properties": {
+        "dateTime": {
+          "type": "string",
+          "description": "Date and time of the last check in of the item.",
+          "format": "date-time"
+        },
+        "servicePointId": {
+          "type": "string",
+          "description": "Service point ID being used by a staff member when item was scanned in Check in app.",
+          "pattern": "[A-F0-9]{8}-[A-F0-9]{4}-4[A-F0-9]{3}-[89AB][A-F0-9]{3}-[A-F0-9]{12}"
```
bohdan-suprun:
```suggestion
          "$ref": "uuid.json"
```

ramls/item.json
```diff
@@ -296,6 +296,29 @@
...
+        },
+        "staffMemberId": {
+          "type": "string",
+          "description": "ID a staff member who scanned the item",
+          "pattern": "[A-F0-9]{8}-[A-F0-9]{4}-4[A-F0-9]{3}-[89AB][A-F0-9]{3}-[A-F0-9]{12}"
```
bohdan-suprun:
```suggestion
          "$ref": "uuid.json"
```

mrdpopov: Done.

ramls/item.json
mrdpopov: Done.

bohdan-suprun:
> I intended to create tests in mod-cirulation, where these new fields are going to be used. Do you think the tests are realy needed in this module?

@mrdpopov, I think we still need tests in this module. The mod-circulation uses fake storage modules (it is an endpoint which stores all the data in a map), it does not have integration with the storage module.

mrdpopov:
These three fields will be populated externaly, in mod-circulation. That means we'll need to mock this behaviour in this module: first populate these fields, then check if they exist. Is there a point of such tests? Which tests do you think should be written?

mreno-EBSCO:
> Do you think this can be merged?

Cross domain bloat makes any arguments I may have for where this should go moot. I guess we can look at it as item specific "last scanned" data. That type of data seems reasonable to put in inventory.

I think "check in" is a bit of an overloaded term. My understanding is librarians can and do use the "check in" app just to see what is up with an item. After scanning they could just put the item back where they found it.

I agree with @bohdan-suprun though, we should have unit tests to ensure the generated code does what we expect it to do, including validation.

I still don't understand how the `lastCheckIn` object is set during check in if it is marked read only in the schema.

ramls/item.json
mrdpopov: It was readonly by mistake

mrdpopov:
@marcjohnson-kint , @mreno-EBSCO Do you think this is ready to be merged?

PR #347, opened 2019-10-25
FOLIO-2321 Remove old MD metadata

dcrossleyau:
See notes at [FOLIO-2321](https://issues.folio.org/browse/FOLIO-2321) and [docs](https://dev.folio.org/guides/module-descriptor/#metadata).


dcrossleyau:
This is a DevOps task. Review not needed.

PR #349, opened 2019-11-01
MODINVSTOR-367 - Maintain item status date even if item record is edited

VRohach:
1. Trigger on update item status date if status changed added.
2. "format": "date-time"  for "date" field was added to the 'item.json'.
3. Tests added.

**One comment:** there isn`t an opportunity to return this item status date in checkin/checkout.
As the mechanism of creating item status date was added as a trigger to DB.

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -297,6 +297,28 @@
       "folio:linkFromField": "holdingsRecordId",
       "folio:linkToField": "id",
       "folio:includedElement": "holdingsRecords.0"
+    },
+    "lastCheckIn": {
+      "type": "object",
+      "additionalProperties": false,
+      "description": "Information about when an item was last scanned in the Inventory app.",
+      "properties": {
+        "dateTime": {
+          "type": "string",
+          "description": "Date and time of the last check in of the item.",
+          "format": "date-time"
+        },
+        "servicePointId": {
+          "type": "string",
+          "description": "Service point ID being used by a staff member when item was scanned in Check in app.",
+          "$ref": "uuid.json"
+        },
+        "staffMemberId": {
+          "type": "string",
+          "description": "ID a staff member who scanned the item",
+          "$ref": "uuid.json"
+        }
+      }
     }
   },
   "additionalProperties": false,
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -195,8 +195,9 @@
           "type": "string"
         },
         "date": {
-          "description": "Date of a status",
-          "type": "string"
+          "description": "Date and time when the status was last changed",
+          "type": "string",
+          "format": "date-time"
         }
       },
       "additionalProperties": false
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -196,8 +196,7 @@
         },
         "date": {
           "description": "Date and time when the status was last changed",
-          "type": "string",
-          "format": "date-time"
+          "type": "string"
         }
       },
       "additionalProperties": false
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -11,7 +11,7 @@ AS $$
   END IF;
 	IF (TG_OP='UPDATE') THEN
 	  IF (newStatus!=OLD.jsonb->'status'->>'name') THEN
-          NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(NOW()), true);
+          NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(CURRENT_TIMESTAMP(3)), true);
 	  END IF;
 	END IF;
 	RETURN NEW;
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -1,0 +1,23 @@
+
+-- Updates item status date if status name was changed.
+-- Item status date stores with timezone.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_status_date() RETURNS TRIGGER
+AS $$
+  DECLARE
+	  newStatus text;
+  BEGIN
+  	newStatus = NEW.jsonb->'status'->>'name';
+  IF newStatus IS NULL THEN  RETURN NEW;
+  END IF;
+	IF (TG_OP='UPDATE') THEN
+	  IF (newStatus!=OLD.jsonb->'status'->>'name') THEN
+          NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(NOW()), true);
+	  END IF;
+	END IF;
+	RETURN NEW;
+  END;
+  $$ LANGUAGE 'plpgsql';
+
+DROP TRIGGER IF EXISTS update_status_date ON ${myuniversity}_${mymodule}.item;
+CREATE TRIGGER update_status_date BEFORE UPDATE
+ON ${myuniversity}_${mymodule}.item FOR EACH ROW EXECUTE PROCEDURE ${myuniversity}_${mymodule}.update_status_date();
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -8,7 +8,8 @@ AS $$
   BEGIN
   	newStatus = NEW.jsonb->'status'->>'name';
 	  IF (newStatus IS DISTINCT FROM OLD.jsonb->'status'->>'name') THEN
-          NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(CURRENT_TIMESTAMP(3)), true);
+	    -- Date time in "yyyy-MM-dd'T'HH:mm:ss.SSS" format at UTC (00:00) time zone
+      NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(CURRENT_TIMESTAMP(3) AT TIME ZONE 'UTC'), true);
 	  END IF;
 	RETURN NEW;
   END;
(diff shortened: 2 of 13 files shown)


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -632,6 +633,125 @@ public void canPlaceAnItemInTransit()
...
+
+    JsonObject item = getResponse.getJson();
+
+    assertThat(item.getString("id"), is(id.toString()));
+
+    assertThat(item.getJsonObject("status").getString("name"),
+      is("Checked out"));
+
+    assertThat(item.getJsonObject("status").getString("date"),
+      notNullValue());
```
mreno-EBSCO: Can we check for more than not null? Since we are relying on PostgreSQL to create and store the date with the proper format, can we at least verify the date is ISO8601 in this test?

VRohach: @mreno-EBSCO done it. Could you please review my approach?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -0,0 +1,23 @@
+
+-- Updates item status date if status name was changed.
+-- Item status date stores with timezone.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_status_date() RETURNS TRIGGER
+AS $$
+  DECLARE
+	  newStatus text;
+  BEGIN
+  	newStatus = NEW.jsonb->'status'->>'name';
+  IF newStatus IS NULL THEN  RETURN NEW;
```
bohdan-suprun: Will it work for the case when OLD.status != null but NEW.status is null?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -0,0 +1,23 @@
...
+  END IF;
+	IF (TG_OP='UPDATE') THEN
```
bohdan-suprun: Do we need this check, since the trigger is already before update?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -0,0 +1,23 @@
...
+	  IF (newStatus!=OLD.jsonb->'status'->>'name') THEN
```
bohdan-suprun: I'm not sure if this will work for case when either newStatus or oldStatus is null. Take a look at `IS DISTINCT FROM` operator.

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -0,0 +1,23 @@
...
+          NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', to_jsonb(CURRENT_TIMESTAMP(3)), true);
```
bohdan-suprun: should we convert it to UTC time zone (+00:00)? I saw we use this time zone for metadata fields. 

@mreno-EBSCO what do you think?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -0,0 +1,23 @@
+
+-- Updates item status date if status name was changed.
+-- Item status date stores with timezone.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_status_date() RETURNS TRIGGER
```
bohdan-suprun:
```suggestion
CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.update_item_status_date() RETURNS TRIGGER
```

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
VRohach: Yes, you are right. Thanks!

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
VRohach: Ok, will do it.

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
VRohach: Thank you for advice. Done

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
VRohach: I just added "IS DISTINCT FROM" and removed this check.

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
bohdan-suprun: Hi @marcjohnson-kint, could you please advice which timezone should we use here? 

UTC with zero offset or DB server default (as in this case)?

ramls/item.json
```diff
@@ -195,8 +195,9 @@
           "type": "string"
         },
         "date": {
-          "description": "Date of a status",
-          "type": "string"
+          "description": "Date and time when the status was last changed",
+          "type": "string",
+          "format": "date-time"
```
marcjohnson-kint: This could be considered a breaking compatible change (though, opinions differ as to whether that is practically the case).

@skoczko should this be a regarded as a breaking compatibility change?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
marcjohnson-kint: The timezone should be UTC

marcjohnson-kint:
> 1. Trigger on update item status date if status changed added.

Does this mean we are effectively converting item status date to a server controlled property?

If we are, then this property becomes read only, and I think this is a breaking compatibility change. @skoczko do you agree?

Are we sure that every change of item status should update this date? I believe there were conversations in the original work that deterred us from doing this. I cannot recall the details.

> 4. Version in pom.xml bumped up to 17.2.0.

Why has the implementation version been changed?

> **One comment:** there isn`t an opportunity to return this item status date in checkin/checkout.
> As the mechanism of creating item status date was added as a trigger to DB.

I don't think I understand this statement. Do you mean the the item status is updated using a PUT which does not provide a representation in the response, meaning clients need to fetch the item again to get this information.

VRohach:
> 
> > 1. Version in pom.xml bumped up to 17.2.0.
> 
> Why has the implementation version been changed?

**1. Regarding version changing** - I already removed this change, because it is not actual. Thanks!
> 
> > **One comment:** there isn`t an opportunity to return this item status date in checkin/checkout.
> > As the mechanism of creating item status date was added as a trigger to DB.
> 
> I don't think I understand this statement. Do you mean the the item status is updated using a PUT which does not provide a representation in the response, meaning clients need to fetch the item again to get this information.

2. So, in the current implementation, new value in the field "date" (item status date) will set after putting the item in DB. That`s why we don`t have an opportunity to return this value (item status date) in the checkout and check in response.  (We can do it only if we will make one more get request)

marcjohnson-kint:
> **1. Regarding version changing** - I already removed this change, because it is not actual. Thanks!

Then please remove this from the description of the pull request, it is now inconsistent with the actual changes and could cause confusion if used for reference.

> 1. So, in the current implementation, new value in the field "date" (item status date) will set after putting the item in DB. That`s why we don`t have an opportunity to return this value (item status date) in the checkout and check in response.  (We can do it only if we will make one more get request)

I think we are agreeing. We might have different emphasis.

I understand that there is a limitation in the database update and I believe there is a RAML Module Builder issue for resolving that.

Even if that field was available to the code after the database update occurred, it wouldn’t be useful, as a PUT request does not provide a response that contains the updated representation (because it is intended to be a complete replacement).

Hence a client would need to make the subsequent GET either way. I imagine this is what mod-circulation is going to need to do.

Does that make sense?

mrdpopov:
@marc-johnson Yes it makes sense, The clients will indeed need to execute GET after updating an item, _in case if they need this information_. I can hardly imagine such a use case when the status date field is desperately needed in PUT response. The status of an item changes in checkin and checkout and the date is hardly needed or useful in the response. In the item details view in FOLIO, where this field is displayed for end-users, GET is executed by the browser anyway. 

Hope this makes sense. Are there any other concerns about this PR?

marcjohnson-kint:
> @marc-johnson Yes it makes sense, The clients will indeed need to execute GET after updating an item, _in case if they need this information_. I can hardly imagine such a use case when the status date field is desperately needed in PUT response. The status of an item changes in checkin and checkout and the date is hardly needed or useful in the response. In the item details view in FOLIO, where this field is displayed for end-users, GET is executed by the browser anyway.

I think I'm confused. The pull request description specifically comments on this limitation in the context of check in or check out, yet now it is suggested these processes don't need it. Anyway, I think we both understand the limitation.

> Hope this makes sense. Are there any other concerns about this PR?

My primary concern with this pull request is that it seems to change the nature of the `status changed date` to be server controlled and hence read only.

If that is the case, then I think the schema needs further changing to reflect that, and this is a compatibility breaking change (@skoczko do you agree?)

marcjohnson-kint:
Given that I allowed the breaking compatibility changes to be split out from [MODINVSTOR-376](https://issues.folio.org/browse/MODINVSTOR-376), it is only consistent that I also allow that same deferment for this issue.

@bohdan-suprun Given that you did the development work for MODINVSTOR-376 (https://github.com/folio-org/mod-inventory-storage/pull/351) please can you review this and advise on how this work should proceed (e.g. by raising a follow up issue, similar to what you did).

Apologies to all involved :-(

VRohach:
@bohdan-suprun Yes, sure.

JIRA-ticket for breaking changes resolving: https://issues.folio.org/browse/MODINVSTOR-392

VRohach:
@marcjohnson-kint Could you please review and approve this PR?

marcjohnson-kint:
@VRohach This pull request has already been approved by @bohdan-suprun why do I need to provide a review?

PR #350, opened 2019-11-01
MODINVSTOR-387 Admin data not updating properly on Instances created from SRS

OleksiiKuzminov:
Expected:

When the instance is first created by data-import, the create/update date/time should be the same, and roughly equivalent to the "ended running" date/time in the import log.
When the instance is updated, the update date/time should change to be a few minutes later, but the create date/time should stay the same.

Actual:

When the instance is first created by data-import, the create/update date/time shows as unknown
When the instance is updated, the update date/time and the create date/time both change to the updated time

### DIFF
--- a/src/main/java/org/folio/rest/impl/InstanceStorageBatchAPI.java
+++ b/src/main/java/org/folio/rest/impl/InstanceStorageBatchAPI.java
@@ -6,24 +6,30 @@ import io.vertx.core.CompositeFuture;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import io.vertx.core.json.JsonObject;
 import io.vertx.core.logging.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/rest/impl/InstanceStorageBatchAPI.java
```diff
@@ -155,4 +160,31 @@ private InstancesBatchResponse constructResponse(List<Future> saveFutures) {
     return response;
   }
 
+  private void populateMetaDataForList(List<Instance> list, Map<String, String> okapiHeaders) {
+    String userId = okapiHeaders.getOrDefault(OKAPI_USERID_HEADER, "");
```
Igor-Gorchakov: Please use `org.apache.commons.lang3.StringUtils.EMPTY`  instead of `""`

PR #351, opened 2019-11-04
MODINVSTOR-376: Handle statusUpdatedDate property for instance.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-376: Instance record. Status updated date does not populate in detailed view

### Approach:

Added new before update trigger for instance table that sets `statusUpdatedDate` = NOW in case instance's `statusId` property has been changed.

The date time in `statusUpdatedDate` has the following format `yyyy-MM-dd'T'HH:mm:ss.SSS` (ISO-8601 that is standard in java) and it is in UTC timezone.

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -1562,7 +1562,7 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
    * Test case for instanceStatusUpdatedDateTrigger.sql trigger.
    */
   @Test
-  public void shouldOverrideStatusUpdatedDate() throws Exception {
+  public void shouldChangeStatusUpdatedDateOnSubsequentStatusChanges() throws Exception {
     UUID id = UUID.randomUUID();
     JsonObject instanceToCreate = smallAngryPlanet(id)
       .put("statusId", getOtherInstanceType().getId().toString());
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1518,6 +1534,97 @@ public void shouldReturnErrorResponseIfAllInstancesFailed()
...
+
+    assertThat(updatedInstance
+        .getString(STATUS_UPDATED_DATE_PROPERTY), withinSecondsBeforeNow(seconds(2)));
+  }
+
+  /**
+   * Test case for instanceStatusUpdatedDateTrigger.sql trigger.
+   */
+  @Test
+  public void shouldOverrideStatusUpdatedDate() throws Exception {
```
marcjohnson-kint: What does `override` mean in this context, is this test about the status updated date changing when the status changes?

bohdan-suprun: this test about updating instance status twice.

marcjohnson-kint: @bohdan-suprun It is about the `status updated date` being updated when the status changes for a second time?

If so, maybe something like `shouldChangeStatusUpdatedDateOnSubsequentStatusChanges`?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1518,6 +1534,97 @@ public void shouldReturnErrorResponseIfAllInstancesFailed()
...
+        .getString(STATUS_UPDATED_DATE_PROPERTY), withinSecondsBeforeNow(seconds(2)));
+    assertThat(updatedInstanceWithOthStatus
+        .getString(STATUS_UPDATED_DATE_PROPERTY), withinSecondsBeforeNow(seconds(1)));
+  }
+
+  /**
+   * Test case for instanceStatusUpdatedDateTrigger.sql trigger.
+   */
+  @Test
+  public void shouldNotOverrideStatusUpdatedDateOnSameStatusId() throws Exception {
```
```suggestion
  public void shouldNotChangeStatusUpdatedDateWhenStatusHasNotChanged() throws Exception {
```

PR #352, opened 2019-11-04
MODINVSTOR-349 create keyword CQL index

evaluk:
This story implements the newest functionality found here: https://github.com/folio-org/raml-module-builder/pull/533.  The example given is a unit test.


evaluk:
I made this PR and then realized we need to do a release for RMB before it will pass.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -448,7 +449,49 @@ public void canGetAllInstances()
     assertThat(secondInstance.getJsonArray("identifiers"),
       hasItem(identifierMatches(UUID_ASIN.toString(), "B01D1PLMDO")));
   }
+  @Test
+  public void canGetWithAllIndex()
```
```suggestion
  public void canSearchUsingKeywordIndex()
```

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -448,7 +449,49 @@ public void canGetAllInstances()
     assertThat(secondInstance.getJsonArray("identifiers"),
       hasItem(identifierMatches(UUID_ASIN.toString(), "B01D1PLMDO")));
   }
+  @Test
```
julianladisch: Can you add a blank line between methods?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -448,7 +449,49 @@ public void canGetAllInstances()
...
+    JsonObject firstInstance = allInstances.getJsonObject(0);
+
+
+    assertThat(firstInstance.getString("id"), is(firstInstanceId.toString()));
+    assertThat(firstInstance.getString("title"), is("Long Way to a Small Angry Planet"));
+
+    assertThat(firstInstance.getJsonArray("identifiers").size(), is(1));
+    assertThat(firstInstance.getJsonArray("identifiers"),
+      hasItem(identifierMatches(UUID_ISBN.toString(), "9781473619777")));
+  }
```
julianladisch: Can you add a blank line between methods?

julianladisch: Can you use the helper method for testing the search: searchForInstancesWithin5(String cql)?

marcjohnson-kint:
The example in the test appears to search for a single instance using both the title and author. Is this how we are intending this search to be used?

If not, please could examples be added that demonstrate how this behaves differently to a title search, for example.

As we are intending to use the `all` keyword (@zburke please confirm) in the UI for this search, maybe we could demonstrate how that works in these tests.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -448,7 +449,49 @@ public void canGetAllInstances()
...
+    client.get(instancesStorageUrl("?query=keyword%3D%22Long%20Way%20to%20a%20Small%20Angry%20Planet%20Chambers%2C%20Becky%209781473619777%22"), StorageTestSuite.TENANT_ID,
+        ResponseHandler.json(getCompleted));
+
+    Response response = getCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject responseBody = response.getJson();
+
+    JsonArray allInstances = responseBody.getJsonArray("instances");
+
+    assertThat(allInstances.size(), is(1));
```
julianladisch: This test (size is 1) is the only check needed for this integration test. All other assertThat tests should be removed because they have already been checked in other tests.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
evaluk: as per a previous comment I removed these tests.

evaluk: removed these lines due to not being needed

marcjohnson-kint:
@evaluk The tests for this pull request fail in CI, please could these be investigated before a review in conducted

src/test/java/org/folio/rest/api/InstanceStorageTest.java
evaluk: lines removed

evaluk: lines removed

evaluk:
I created another unit test using the all keyword.

PR #353, opened 2019-11-05
MODINVSTOR-353: implement batch update for instances, items and holdings

julianladisch:
Add three new POST endpoints:
* `/instance-storage/batch/synchronous`
* `/holdings-storage/batch/synchronous`
* `/item-storage/batch/synchronous`

Declare the existing POST endpoint `/instance-storage/batch/instances` as deprecated (there
are no existing holdings and items batch endpoints).

The existing batch endpoint uses one transaction per record, the new endpoints use a
single INSERT statment with multiple JSONs values to insert all records in a single transaction.

There is no streaming. RMB loads all records into memory and does schema validation. Then all
records are passed to mod-inventory-storage and then sent to the database.

The "413 Payload Too Large" response is declared but not implemented; this needs to be
implemented in RMB because RMB must stop a POST upload if the payload exceeds some limit.
Doing this in mod-inventory-storage module is too late.

*Content-Length*/*413 Payload Too Large* and using a streaming method without loading the complete batch into mod-inventory-storage memory requires RMB support that doesn't exist yet. Therefore it cannot be implemented within the scope of this issue. It this has been split into RMB-505.

This pull request's implementation uses PostgresClient.saveBatch that does not support streaming.


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1625,6 +1627,78 @@ public void shouldNotChangeStatusUpdatedDateWhenStatusHasNotChanged() throws Exc
...
+    assertExists(uprooted);
+    assertExists(smallAngryPlanet);
+    assertExists(temeraire);
+  }
+
+  @Test
+  public void cannotSyncPostWithInvalidInstance() throws Exception {
+    JsonArray instancesArray = new JsonArray();
+    instancesArray.add(uprooted(UUID.randomUUID()));
+    instancesArray.add(smallAngryPlanet(UUID.randomUUID()).put("foo", "bar"));  // "foo" is invalid
```
bohdan-suprun: I'd extract the `"foo"` to a variable (e.g. `invalidPropertyName`), so there will be no need for the comment.

ramls/holdingsrecord-with-id.json
```diff
@@ -0,0 +1,282 @@
+{
```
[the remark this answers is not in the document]
julianladisch: No, because the `id` property is optional in `holdingsrecord.json` and required in `holdingsrecord-with-id.json`: https://github.com/folio-org/mod-inventory-storage/blob/6cf0931ca346523ca30f0f39a90161b78df9b0ad/src/test/java/org/folio/rest/unit/SchemaTest.java#L24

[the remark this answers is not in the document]
bohdan-suprun: As far as I understand, these endpoints will be used for instances/holdings/items upload, not for updates. Why do we need ids to be required?

descriptors/ModuleDescriptor-template.json
```diff
@@ -33,6 +33,17 @@
         }
       ]
     },
+    {
+      "id": "item-storage-batch-sync",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/item-storage/sync",
```
marcjohnson-kint: I think this path is ambiguous. The `sync` part could correspond to a synchronisation process rather than a synchronous operation.

Why does the path need to explain that this operation is synchronous, is it to avoid conflicts with the existing batch API or to create space for a future asynchronous variant?

The path does not include any mention that it is for a batch, why is that? Is it to avoid issues with the existing batch API?

cc: @skoczko

src/test/java/org/folio/rest/api/InstanceStorageTest.java
julianladisch: Good point, I've replaced "foo" by "invalidPropertyName" and removed the comment.
I don't think that extracting the String into a variable improves readability.

ramls/holdingsrecord-with-id.json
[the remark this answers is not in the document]
bohdan-suprun: Personally, I'd prefer to have a java based validation for the id then, instead of maintaining two separate files of the schemes (considering that these schemes are changed frequently).

What do you think @marcjohnson-kint, @mreno-EBSCO?

marcjohnson-kint:
> Declare the existing POST endpoint /instance-storage/batch/instances as deprecated (there
> are no existing holdings and items batch endpoints).

Does that mean that the intention is for the data import modules to convert over to using this new API?

Has this deprecation been discussed with the data import team, who introduced the experimental batch API?

> There is no streaming. RMB loads all records into memory and does schema validation. Then all
> records are passed to mod-inventory-storage and then sent to the database.
> 
> The "413 Payload Too Large" response is declared but not implemented; this needs to be
> implemented in RMB because RMB must stop a POST upload if the payload exceeds some limit.
> Doing this in mod-inventory-storage module is too late.

Is this because checking for the payload being too large needs to be integrated with the streaming approach, so RAML Module Builder stops reading the stream when it is too large?

Is there a RAML Module Builder issue for this functionality? As I imagine this implementation will need to change when this is introduced? (@skoczko based upon the recent conversations about interface versioning, would introducing a maximum batch size be considered a breaking change?)

Will that payload size be configurable by the implementing module?

descriptors/ModuleDescriptor-template.json
```diff
@@ -33,6 +33,17 @@
         }
       ]
     },
+    {
+      "id": "item-storage-batch-sync",
+      "version": "1.0",
```
marcjohnson-kint: Personally, I'd consider marking these interfaces as experimental, until streaming support / too large payloads etc are in place

ramls/holdingsrecord-with-id.json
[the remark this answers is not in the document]
marcjohnson-kint: This is a long running challenge with using the same schema for both read and write representations of the same record, it leads to compromises like this.

The `id` property is only optional in the record schema because it can be specified by the client during creation (using POST). In all other circumstances it should be effectively required.

If we decide to have two schema, I think we need to figure out how to help maintainers know they need to change both schema together.

I personally think we could do the check after the records have been received and issue a 422 if the ID isn't present. We do this for other conditional validation that cannot be expressed in the schema.

Which trade off do we want to make?

If we do decide to use two schema, I'd prefer we investigate adopting read and write schema (or similar, because create is slightly different to update) rather than two variants of the schema whose usage is less clear. I'm aware this isn't ideal.

Including @skoczko in this conversation as well.

descriptors/ModuleDescriptor-template.json
julianladisch: "sync" is used to differentiate from an async API that may be introduced at a later stage.
While it is true that both synchronisation and synchronous have the same abbreviation sync this is not ambiguous because there is no synchronisation endpoint. The path name is used to distinguish existing endpoints, it doesn't replace the API documentation.

The GET /item-storage/items API also returns a batch of records without mentioning it in the path name and there exists POST /item-storage/items with the same path where only a single record can be uploaded in one request. Therefore the path is short and does not necessarily contain the complete documentation.

marcjohnson-kint: > "sync" is used to differentiate from an async API that may be introduced at a later stage.
> While it is true that both synchronisation and synchronous have the same abbreviation sync this is not ambiguous because there is no synchronisation endpoint. The path name is used to distinguish existing endpoints, it doesn't replace the API documentation.

I personally find it ambiguous and confusing.

I would prefer something like `/item-storage/items/batch/synchronous` as I think it expresses it intent more clearly.

Please feel free to disregard this feedback given that you disagree with it.

> The GET /item-storage/items API also returns a batch of records without mentioning it in the path name 

I would suggest that the collection level get returns a collection or subset of records, not a batch.

I think of a batch as set of records submitted for processes, rather than any endpoint that involves multiple records.

> and there exists POST /item-storage/items with the same path where only a single record can be uploaded in one request. 

As I understand it, that is a specific convention that the collection endpoints are plural and creation occurs at the collection endpoint.

> Therefore the path is short and does not necessarily contain the complete documentation.

Agreed, I think it can be a reasonable indicator as to the intent of the endpoint.

ramls/holdingsrecord-with-id.json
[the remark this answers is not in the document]
mreno-EBSCO: Since we use the schema for API documentation, I think that has to be accurate as possible. Indicating that an ID is not required in the schema, but in the code making required in certain cases makes the APIs harder to use, IMHO. There are several cases where this occurs in existing modules and it is frustrating.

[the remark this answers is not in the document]
julianladisch: The id is required because this API does not return anything and therefore cannot report the generated id.
The client that calls this API needs the instance id to put it into the holding.instanceId field. The same for item.holdingsRecordId.
If the insert fails the id allows to easily track which records caused the failure.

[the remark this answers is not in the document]
julianladisch: Having two schema that differ by id requirement only has several advantages:
* RMB can do the validation. No need to split the validation (schema and java).
* The schema shown in the API documentation https://dev.folio.org/reference/api/ is correct, one can rely on the list of required fields.
* Maintainers are forced to update both schemas (with and without id) at the same time because the schemasWithIdAreInSync unit test will fail if the files differ: https://github.com/folio-org/mod-inventory-storage/blob/938fcecd8ee96764704320f496854102a653e570/src/test/java/org/folio/rest/unit/SchemaTest.java#L27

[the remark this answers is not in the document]
bohdan-suprun: Thanks for explaining it, @julianladisch.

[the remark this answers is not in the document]
adamdickmeiss: I also think it's horrible to have two schemas and is in no way convinced about that being elengant because some module has a unit test for some schema. We started with DB generated UUID, then Java based UUID. The final stage would be client side generated in which case we'd have one schem only - with a required UUID. That would be a major change, but nutural step forward IMHO.

descriptors/ModuleDescriptor-template.json
```diff
@@ -33,6 +33,17 @@
...
       ]
     },
+    {
+      "id": "item-storage-batch-sync",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/item-storage/batch/synchronous",
+          "permissionsRequired": ["inventory-storage.items.batch.post"]
```
marcjohnson-kint: If we might introduce an asynchronous batch approach, would we want to constrain access to that with different permissions?

If so, I suggest the permissions for these endpoints be named specifically for the synchronous version.

ramls/holdings-sync.raml
```diff
@@ -0,0 +1,40 @@
+#%RAML 1.0
+title: Inventory Storage Holdings Batch Sync API
+version: v1.0
```
marcjohnson-kint: This version should match the one used in the module descriptor. I believe that was recently changed to 0.1. If so, please update the RAML files as well.

ramls/holdingsrecord-with-id.json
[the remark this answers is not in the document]
marcjohnson-kint: How do we come to an agreement on this topic for this change?

I agree that having more descriptive schema is beneficial.

At the moment, that would require having (at least) two different schema. 

Is the improved description of the `id` property valuable enough to introduce that extra maintenance (inc. the tests)?

I think it is worth considering @adamdickmeiss suggestion for FOLIO standards, as it would remove the challenge around this particular property.

In general, I think there will often be a gap (conditional validation, external state etc) between what can be expressed in a schema and what the server considers valid, so clients will likely always need to be prepared for records that are valid against the schema to be rejected.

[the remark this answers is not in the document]
marcjohnson-kint: If we were to introduce two schema, I’d prefer we made the special case (single record creation, POST and PUT) the explicit second schema and use the other schema for all other situations (assuming that is the distinction).

src/main/java/org/folio/rest/impl/StorageHelper.java
```diff
@@ -61,4 +71,16 @@ protected static PostgresClient postgresClient(Context vertxContext, Map<String,
     return PostgresClient.getInstance(vertxContext.owner(), TenantTool.tenantId(okapiHeaders));
   }
 
+  protected static <T> void postSync(String table, List<T> entities, Map<String, String> okapiHeaders,
```
marcjohnson-kint: Is this method intended to eventually become part of PgUtil?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -603,6 +610,60 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
...
+      }
+      holdingsArray.add(new JsonObject()
+          .put("id", UUID.randomUUID().toString())
+          .put("instanceId", instanceId.toString())
+          .put("permanentLocationId", mainLibraryLocationId.toString()));
+    }
+    return holdingsArray;
+  }
+
+  private Response postSync(JsonArray holdingsArray) {
```
marcjohnson-kint:
```suggestion
  private Response postSyncronousBatch(JsonArray holdingsArray) {
```

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -603,6 +610,60 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
...
+  public void canSyncPost() {
+    JsonArray holdingsArray = threeHoldings();
+    assertThat(postSync(holdingsArray), statusCodeIs(HttpStatus.HTTP_CREATED));
+    for (Object holding : holdingsArray) {
+      assertExists((JsonObject) holding);
+    }
+  }
+
+  @Test
+  public void cannoteSyncPostWithDuplicateId() {
```
marcjohnson-kint: I think it could be valuable to also have a test where a batch is submitted with a record ID that already exists in the system

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1625,6 +1627,78 @@ public void shouldNotChangeStatusUpdatedDateWhenStatusHasNotChanged() throws Exc
       .getString(STATUS_UPDATED_DATE_PROPERTY), withinSecondsBeforeNow(seconds(2)));
   }
 
+  private void assertExists(JsonObject expectedInstance) throws Exception {
```
marcjohnson-kint: Given that this method is used by multiple tests, I think it is hard to place it near any one usage, and would be better moved to after the tests

descriptors/ModuleDescriptor-template.json
```diff
@@ -33,6 +33,17 @@
         }
       ]
     },
+    {
+      "id": "item-storage-batch-sync",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["POST"],
+          "pathPattern": "/item-storage/batch/synchronous",
```
marcjohnson-kint: It appears the paths have only been updated in the module descriptor and not in the RAML or the tests, is that intentional?

src/main/java/org/folio/rest/impl/StorageHelper.java
julianladisch: Yes.

descriptors/ModuleDescriptor-template.json
julianladisch: Thanks for finding this discrepancy, fixed.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
julianladisch: Yes, I've reordered the methods.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
julianladisch: Yes, I've added an test to InstanceStorageTest, now there are three kinds of failing tests:
* cannotPostSynchronousBatchWithInvalidInstance
* cannotPostSynchronousBatchWithExistingId
* cannotPostSynchronousBatchWithDuplicateId
For Holdings and Items I only have a single failing test because it is the same StorageHelper.postSync code, no need to duplicate the tests.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
julianladisch: Changed.

ramls/holdings-sync.raml
julianladisch: Done.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
marcjohnson-kint: > For Holdings and Items I only have a single failing test because it is the same StorageHelper.postSync code, no need to duplicate the tests.

I understand this trade off. In general, I'm not keen on relying on implementation similarity for API tests, which I think are intended to be as ignorant of that as possible and act as documentation for API usage.

descriptors/ModuleDescriptor-template.json
julianladisch: I don't think that there is a need for this. If a user is allowed to POST to one of the batch API endpoints why should the POST to the other batch API endpoints been restricted? They all give access to the same table (for example instance).
And we can easily create subpermissions at a later time if there actually is a need for it.

marcjohnson-kint: Ok, I'll leave the decision to you. 

I raised the question because I tend to err on the the side of caution with permissions, making them more specific if unsure.

This is due to the UI modules composing these into higher level permission sets, which introduces a subtle coupling which effects how easy it is to change the permissions later.

julianladisch:
> instance.json is used as argument for collection raml and that is the same type used for both GET, POST, PUT

For POST the id is optional because the server generates an id if needed. For PUT the id is optional because the last path segment of the URL is the id and copied into the entity: https://github.com/folio-org/raml-module-builder/blob/v28.1.0/domain-models-runtime/src/main/java/org/folio/rest/persist/PgUtil.java#L768

julianladisch:
> Has this deprecation been discussed with the data import team, who introduced the experimental batch API?

Yes, this was discussed on the pull request #295:

> (ongoing discussion is in FOLIO-2050). We will definitely change this implementation once the design for batch endpoints is defined.
> Since it is an "experimental" API (version 0.1) and we plan to change it anyway, we probably can move forward with this PR.

We know that /instance-storage/batch/instances has API design flaws, but how to proceed with it is off-topic for this pull request.

julianladisch:
> RAML Module Builder stops reading the stream when it is too large?

> Is there a RAML Module Builder issue for this functionality?

https://issues.folio.org/browse/RMB-505

> As I imagine this implementation will need to change when this is introduced?

Yes.

> based upon the recent conversations about interface versioning, would introducing a maximum batch size be considered a breaking change?

No, because the interface doesn't change, only the implementation. The interface already contains the       413 "Payload Too Large" response code in the three .raml files of this pull request. 

> Will that payload size be configurable by the implementing module?

I've copied this question to https://issues.folio.org/browse/RMB-505 because this is off-topic for this pull request.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -603,6 +612,62 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
...
+  public void canPostSynchronousBatch() {
+    JsonArray holdingsArray = threeHoldings();
+    assertThat(postSynchronousBatch(holdingsArray), statusCodeIs(HttpStatus.HTTP_CREATED));
+    for (Object holding : holdingsArray) {
+      assertExists((JsonObject) holding);
+    }
+  }
+
+  @Test
+  public void cannotePostSynchronousBatchWithDuplicateId() {
```
bohdan-suprun:
```suggestion
  public void cannotPostSynchronousBatchWithDuplicateId() {
```

src/main/java/org/folio/rest/impl/StorageHelper.java
```diff
@@ -61,4 +71,16 @@ protected static PostgresClient postgresClient(Context vertxContext, Map<String,
...
+      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext, Supplier<Response> respond201) {
+    PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+    postgresClient.saveBatch(table, entities, result -> {
+      if (result.failed()) {
+        logger.error(result.cause().getMessage(), result.cause());
```
bohdan-suprun:
```suggestion
        logger.error("Error has occurred during executing post-sync", result.cause());
```

Sometimes exceptions has no message (like NPE), I think it more informatively to add a message.

marcjohnson-kint:
@julianladisch This pull request has already been reviewed by @adamdickmeiss can you help me understand why an additional review has been requested?

julianladisch:
> @julianladisch This pull request has already been reviewed by @adamdickmeiss can you help me understand why an additional review has been requested?

https://wiki.folio.org/display/FOLIJET/Core%3Afunctional+-+Definition+of+Done requires:
> Peer code review is performed with at least one developer from the team that "owns" the module

Adam is not a member of https://github.com/orgs/folio-org/teams/inventory-backend-code-owners/members

In addition @bohdan-suprun raised a new issue after Adam's review and that new issue hasn't been resolved, it is still open and requires review.

marcjohnson-kint:
> > @julianladisch This pull request has already been reviewed by @adamdickmeiss can you help me understand why an additional review has been requested?
> 
> https://wiki.folio.org/display/FOLIJET/Core%3Afunctional+-+Definition+of+Done requires:
> 
> > Peer code review is performed with at least one developer from the team that "owns" the module
> 
> Adam is not a member of https://github.com/orgs/folio-org/teams/inventory-backend-code-owners/members
> 
> In addition @bohdan-suprun raised a new issue after Adam's review and that new issue hasn't been resolved, it is still open and requires review.

Whilst I agree that is what the policy states, in practice, I believe changes made by the core platform team have been approved and merged without a code owner being the approver. Often, other folks don't have the context to review the work conducted by the platform team.

We could add the core platform developers to the inventory and circulation code owners teams to make this explicit?

This would have the effect of them being notified for every pull request for the modules set up for review by those teams.

@skoczko @adamdickmeiss what do you think?

src/main/java/org/folio/rest/impl/StorageHelper.java
julianladisch: `result.cause().getMessage()` may contain useful information like `"duplicate key value violates unique contraint "instance_pkey"`. Therefore I do not drop that information but prepend "postSync: " to give more context.

julianladisch:
> Peer code review is performed with at least one developer from the team that "owns" the module

Simply add "or from the core platform team" to this Definition of Done entry.

src/main/java/org/folio/rest/impl/StorageHelper.java
bohdan-suprun: Thanks @julianladisch, works for me.

PR #354, opened 2019-11-07
Implement MODINVSTOR-363: Create settings for hrId

mreno-EBSCO:
* Created a new API under `hrid-settings-storage` for retrieving and
updating the HRID settings.
* Created an HRID manager that will manage the HRID settings as well as
issue new instance/holding/item HRIDs.
* Added DB enablement/upgrade scripts to add the new sequences/tables
for HRID settings and to remove the old sequences/functions/triggers.


ramls/hridsettings.json
```diff
@@ -0,0 +1,38 @@
...
+      "description": "The ID of the one and only HRID settings record",
+      "type": "string",
+      "readonly": true,
+      "$ref": "uuid.json"
+    },
+    "instance": {
+      "description": "The HRID settings for instances",
+      "type": "object",
+      "$ref": "hridsetting.json",
+      "default": {"prefix": "in", "startNumber": 1}
```
marcjohnson-kint: What does this property do? Does it set the values of those properties when they aren’t present in the record?

If so, what happens when settings are submitted that intend to remove the prefix?

ramls/examples/hridsettings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "id": "ca81eafd-ddb8-416a-89e1-4f772ab6afee",
+  "instance": {
+    "prefix": "in",
+    "startNumber": 1
+  },
+  "holding": {
```
marcjohnson-kint: Hmm, is a holding singular, for the records themselves we refer to an individual record as a holdings record? 

I’m aware this is rather pedantic. The UI headings for these settings appears to be inconsistent, in that item is singular, and instances and holdings are plural, which adds to my confusion. 

I’m not really sure whether to try to raise rugs?

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
```diff
@@ -0,0 +1,473 @@
...
+    assertThat(actualHridSettings.getItem().getPrefix(),
+        is(newHridSettings.getItem().getPrefix()));
+    assertThat(actualHridSettings.getItem().getStartNumber(),
+        is(newHridSettings.getItem().getStartNumber()));
+
+    log.info("Finished cannotUpdateHridSettingsID()");
+  }
+
+  @Test
+  public void canGetNextInstanceHrid(TestContext testContext)
```
marcjohnson-kint: This test appears not to use the API, is that the case?

If so, is there a more appropriate place for it? Does it rely on setup that is done by test base?

ramls/hridsettings.json
mreno-EBSCO: As far as I know, they are just instructional (API docs). Though, I see what you mean about removing the prefix. I was on the fence with the "default" fields, I'll get rid of them. They aren't really defaults, they are initial settings. I suspect there isn't a json-schema field for that. 😄

marcjohnson-kint: > I suspect there isn't a json-schema field for that. 😄

I suspect not

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -0,0 +1,233 @@
...
+  }
+
+  private Future<Void> updateSequence(AsyncResult<SQLConnection> conn, String field,
+      HridSetting existingHridSetting, HridSetting hridSetting) {
+    final Promise<ResultSet> promise = Promise.promise();
+
+    // Only update the sequence if the start number has changed
+    if (existingHridSetting.getStartNumber().intValue() !=
+        hridSetting.getStartNumber().intValue()) {
+      final String sql = String.format("select setval('hrid_%s_seq',%d,FALSE)",
```
bohdan-suprun: Maybe more requirement question. As far as I remember the HRIDs should be unique per items/instances/holdings, am I correct?

If so, how should we handle following case then:
1. Let's say the start number is **6** now;
1. We've issued some hrids, so the current val for a sequence is 9 how.
1. then we modifying start number to either **1-8** now (it is less then current value of the sequence);
1. this method updates the sequences, so the current `next val` becomes less then previous `next val`;
1. when we call the sequence to get a next value, the resulting hrid will be duplicating already issued one.

marcjohnson-kint: @bohdan-suprun I agree with this concern. I asked a similar question during refinement I believe, and IIRC the requirements position was that staff _should not_ change the configuration of the working system. Yet, we aren't going to stop them from doing so, the systems behaviour after which is regarded as unsupported.

@mreno-EBSCO Do you have a similar recollection of the requirements conversation?

Maybe this needs stating explicitly in the JIRA issue so that folks understand this limitation was by design?

ramls/examples/hridsettings.json
mreno-EBSCO: Oh, my bad. I didn't look carefully at the UI settings page mockups to try to correlate the property names to the section headers, though they are quite verbose. I don't have a problem changing `holding` to `holdingsRecord`. Do you think that makes these intended use for these settings clearer? I could also make them all plural if that is better.

I agree that the UI itself is inconsistent and likely needs to be addressed. Perhaps they should have used "Inventory instance and MARC bibliographic records" instead of "Inventory instances and MARC bibliographic records". I'm not sure how to break down the "and" here. Is it "Inventory instances" and "MARC bibliographic records" or is it "Inventory instances records" and "MARC bibliographic records"? Given there is an "Inventory item records", I think they mean "Inventory instances records".

mreno-EBSCO:
> What should be done with existing records? Should they be left as is, or the HRID set using the new scheme? Given that clients are allowed to provide their own HRID, I imagine we leave the existing records as is, and individual institutions need to consider what to do to update any as needed.

Leaving existing fields alone was my thought. What's currently there is likely what the tenant wants or is at least willing to deal with. They would need to ensure that their "new" start number/prefix will not clash with existing hrids.

> The approach taken appears to use a PostgreSQL sequence and changes to the settings updates the sequence. Is that a reasonable summary?

I think so.

src/main/java/org/folio/rest/support/HridManager.java
mreno-EBSCO: You are correct. That is my understanding as well. The tenant needs to know that resetting the start number is not recommended unless they change the prefix. If they do this, they won't be able save the record since there is a unique index on the hrid. This is something a technical manual should detail with some sort of reference in the UI. I'm not sure how apparent we can make this in the API docs. I'll think about that and see if I can come up with something.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
mreno-EBSCO: Many of these tests do not use the API directly. I went back and forth on whether or not to split them out into a separate class but I do, as far as I know, need to extend `TestBase` to leverage the embedded PostgreSQL DB support. I suppose I could have mocked the PostgresClient or duped the DB management code from `StorageTestSuit`. I was trying to save some time, not the best reason for sure.

I'm not sure what a more appropriate place would be. Do you have any suggestions? I see `org.folio.rest.unit`, but these really aren't unit tests.

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: @mreno-EBSCO I've asked this question for confirmation on the issue. I think we should add it to the module readme as a starting point.

(I'm aware the readme isn't very discoverable, and has the challenge of interface vs. implementation)

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
marcjohnson-kint: > Many of these tests do not use the API directly. I went back and forth on whether or not to split them out into a separate class but I do, as far as I know, need to extend `TestBase` to leverage the embedded PostgreSQL DB support. I suppose I could have mocked the PostgresClient or duped the DB management code from `StorageTestSuit`. I was trying to save some time, not the best reason for sure.

Yeah, I wondered if this was because infrastructure was needed.

> I'm not sure what a more appropriate place would be. Do you have any suggestions? I see `org.folio.rest.unit`, but these really aren't unit tests.

I believe this is an unsolved issue for the storage modules. The nature of them means that most require infrastructure, making them integration (depending on the use of the term) tests.

We could introduce a `integration` package name. A concern with this (and the `unit` package, is that it can interfere with code coverage reporting in some cases).

Maybe this a technical debt that we address separately, by separating out the database infrastructure preparation and the API preparation and having different base test classes for API and integration (with external resources) tests.

src/main/java/org/folio/rest/support/HridManager.java
mreno-EBSCO: @marcjohnson-kint OK, I'll put something in the README and add a note in the JIRA for PO visibility. I am pretty sure this is the expected behavior. Trying to manage this level of bookkeeping would be challenging, doable, but challenging. Likewise, if the set the start number to 99999999, they get exactly 1 hrid before the sequence is maxed out.

Another thing that just occurred to me is, if the prefix is changed, but not the start number, then the sequence is not reset to the start start number. I am not sure if that is the correct behavior. Hmm...

marcjohnson-kint: I agree, there a number of different scenarios around this that could cause issues.

This is an example of settings that should only be set in an initial setup period, and then locked (it's not even write once, because they have defaults). 

However FOLIO has no explicit modelling of this concept, which I think is what causes this.

marcjohnson-kint: I think for the moment, the most we can do is issue a warning in the documentation for the module (and the UI maybe).

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
mreno-EBSCO: These tests _may_ become obsolete once the inventory APIs are integrated with the new HRID management. I needed a way to prevent regression while working on this code.

I like the idea of a tech debt story for test restructuring. Also, maybe move to JUnit5, consider getting rid of the test suite, etc.

ramls/examples/hridsettings.json
marcjohnson-kint: I don't really mind if these use the singular or plural of the different records, as long as it is consistent. What I don't know is what the singular is for holdings (as each record represents a set, so the singular could be holdings).

I think `holdings` vs `holdingsRecord` depends upon the plurality question and whether we general refer to this as holdings record. Most of the time, when folks refer to an item, they don't say item record, rather item. I wonder if `record` in the context of holdings is a domain term rather than a technical term.

marcjohnson-kint: I think the and is somewhat misleading since we adapted the requirements.

It used to be that external parts of FOLIO could get an instance HRID and assign it to a MARCCat record or a source record before the instance existed. We've since changed it to only allocate HRIDs upon inventory record creation.

I think this means that these settings don't actually affect MARC records, they are only mapped to them. I'm not sure if I know how or whether I should try to express that.

src/main/java/org/folio/rest/support/HridManager.java
OleksiiKuzminov: Sorry, it's not clear for me how we can change the start value for the sequence on a fly using API?

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
marcjohnson-kint: I'd like to move to JUnit 5, it offers much better options for sharing setup than the suite workaround.

When I last checked, it didn't provide a capability for only running expensive setup once for a set of tests, only sharing the definition. If a pattern has emerged for that, it would help.

marcjohnson-kint: Due to how much of the behaviour of storage modules are defined by RAML Module Builder, pure unit testing for storage modules is a rather large trade off which doesn't really confirm the actual behaviour

marcjohnson-kint: I'm comfortable with these tests being a place holder for API tests that demonstrate allocation of HRIDs for new records. Maybe worth a comment noting that is the purpose of them?

Whether we also want more focused tests can be a separate topic to address at that point.

src/main/java/org/folio/rest/support/HridManager.java
mreno-EBSCO: > Sorry, it's not clear for me, but how we can change the start value for the sequence on a fly using API?

@OleksiiKuzminov, that's the whole purpose of this PR. Are you unclear on how it is implemented or whether or not this should even be possible?

ramls/examples/hridsettings.json
marcjohnson-kint: > Singular is holdings or holdings record. Plural is holdings or holdings records.

I think we can pick whether we have plural or singular property names. It appears that either way, the holdings entry should be `holdings`.

I don't mind which we pick, as long as we are consistent. Given the naming for holdings, it might be that folks less familiar with the domain might find the plural less confusing.

mreno-EBSCO: I am going to pluralize each field by adding an 's'. So, "instances", "holdings" and "items".

src/main/java/org/folio/rest/impl/HridSettingsStorageAPI.java
```diff
@@ -0,0 +1,96 @@
...
+
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.core.logging.Logger;
+import io.vertx.core.logging.LoggerFactory;
+
+public class HridSettingsStorageAPI implements HridSettingsStorage {
+  private static final Logger log = LoggerFactory.getLogger(HridSettingsStorageAPI.class);
```
bohdan-suprun:
```suggestion
  private static final Logger LOG = LoggerFactory.getLogger(HridSettingsStorageAPI.class);
```

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -0,0 +1,233 @@
...
+      });
+    } catch (Exception e) {
+      fail(promise, "Failed to execute updating the HRID settings in a context", e);
+    }
+
+    return promise.future();
+  }
+
+  private Future<HridSettings> getHridSettings(AsyncResult<SQLConnection> conn) {
+    final String sql = "SELECT jsonb FROM " + HRID_SETTINGS_TABLE;
```
bohdan-suprun: I'd extract the whole SQL query to a constant.

src/main/java/org/folio/rest/impl/HridSettingsStorageAPI.java
mreno-EBSCO: The general consensus is to use capitalization to express immutability. Loggers typically are not immutable. The Vert.x `Logger` lets you retrieve the underlying framework logger and modify it. This can be done from any class at any time.

bohdan-suprun: I usually use capitalized names for final fields. Did not think about mutability/immutability.

mreno-EBSCO: Yeah, I didn't either until a few years ago. Google recommends not capitalizing static finals that are mutable.

https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

PR #355, opened 2019-11-13
Modinvstor 391 effective call number migration script

rladdusaw:
- Created a migration script.
- Created tests to verify correctness.


src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
```diff
@@ -0,0 +1,25 @@
+UPDATE ${myuniversity}_${mymodule}.item AS it
```
bohdan-suprun: We have to include this file into schema.json file (see `script` section)

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
```diff
@@ -0,0 +1,25 @@
...
+SET jsonb =
+  CASE
+    WHEN it.jsonb->'itemLevelCallNumber' IS NOT NULL
+      THEN
+        CASE
+          WHEN it.jsonb->'effectiveCallNumberComponents' IS NULL THEN
+            jsonb_set(it.jsonb, '{effectiveCallNumberComponents}',
+              jsonb_set('{"callNumber":null}', '{callNumber}', it.jsonb->'itemLevelCallNumber'))
```
bohdan-suprun: There is also `json_build_object` function, that can be used to build an object:


Example | Example Result
-- | --
json_build_object('foo',1,'bar',2) | {"foo": 1, "bar": 2}

rladdusaw:
@bohdan-suprun I've updated my PR with the changes you requested.

marcjohnson-kint:
> Thanks, looks good to me.
> 
> I think we will have to restructure a bit the migration file in future, since introducing of new properties will be reducing file maintainability.

When sortable call number is introduced, it is likely we will need Java based migration as well.

PR #357, opened 2019-11-14
Modinvstor 360 effective call number suffix

rladdusaw:
- Added call number suffix to effective call number components.
- Extracted item update logic to a utility class.
- Created tests for call number suffix.


src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,40 @@
...
+
+  private EffectiveCallNumberComponentsUtil() {
+    throw new UnsupportedOperationException("Do not instantiate");
+  }
+
+  public static EffectiveCallNumberComponents buildComponents(HoldingsRecord holdings, Item item) {
+    return updateComponents(new EffectiveCallNumberComponents(), holdings, item);
+  }
+
+  public static EffectiveCallNumberComponents updateComponents(EffectiveCallNumberComponents components, HoldingsRecord holdings, Item item) {
```
marcjohnson-kint: Please can this signature be formatted to fit within the line length

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    String updatedCallNumberSuffix = null;
+    if (StringUtils.isNotBlank(item.getItemLevelCallNumberSuffix())) {
+      updatedCallNumberSuffix = item.getItemLevelCallNumberSuffix();
+    } else if (StringUtils.isNotBlank(holdings.getCallNumberSuffix())) {
+      updatedCallNumberSuffix = holdings.getCallNumberSuffix();
+    }
+
+    components.setCallNumber(updatedCallNumber);
```
bohdan-suprun:
```suggestion
   components.setCallNumber(org.apache.commons.lang3.StringUtils.firstNonBlank(
       item.getItemLevelCallNumberSuffix(), holdings.getCallNumberSuffix()
   ));
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
rladdusaw: @marcjohnson-kint Is there a style guide somewhere I can refer to?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -143,22 +145,15 @@ public void deleteItemStorageItemsByItemId(
 
   private CompletableFuture<Item> setEffectiveCallNumber(Map<String, String> okapiHeaders, Context vertxContext, Item item) {
     CompletableFuture<Item> completableFuture = null;
-    EffectiveCallNumberComponents components = new EffectiveCallNumberComponents();
-    if (item.getItemLevelCallNumber() != null && !item.getItemLevelCallNumber().isEmpty()) {
-      components.setCallNumber(item.getItemLevelCallNumber());
-      item.setEffectiveCallNumberComponents(components);
+    if (StringUtils.isNotBlank(item.getItemLevelCallNumber()) && StringUtils.isNotBlank(item.getItemLevelCallNumberSuffix())) {
```
bohdan-suprun: I think, it make sense to extract this logic to a separate method (e.g. `shouldRetrieveHoldingsRecord`)

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
bohdan-suprun:
```suggestion
  private static EffectiveCallNumberComponents updateComponents(EffectiveCallNumberComponents components, HoldingsRecord holdings, Item item) {
```

It seems, that this method is used only within the single class.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
bohdan-suprun: Feel free to ignore it :) I see that holding may be null in some cases, so it is not so easy.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
rladdusaw: Especially since were going to be adding the prefix to this condition soon.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -143,22 +144,15 @@ public void deleteItemStorageItemsByItemId(
 
   private CompletableFuture<Item> setEffectiveCallNumber(Map<String, String> okapiHeaders, Context vertxContext, Item item) {
     CompletableFuture<Item> completableFuture = null;
-    EffectiveCallNumberComponents components = new EffectiveCallNumberComponents();
-    if (item.getItemLevelCallNumber() != null && !item.getItemLevelCallNumber().isEmpty()) {
-      components.setCallNumber(item.getItemLevelCallNumber());
-      item.setEffectiveCallNumberComponents(components);
+    if (shouldRetreiveHoldingsRecord(item)) {
```
bohdan-suprun: Sorry, my bad. Maybe let's rename it to `shouldNotRetreiveHoldingsRecord`. Since this branch is executed when item has all properties == no needs to retrieve holding.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
marcjohnson-kint: Yes, although it is not complete - https://dev.folio.org/guidelines/contributing/#coding-style

marcjohnson-kint: The [Java code standard](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#262) which FOLIO chose, defines the line limit as 80 characters.

ramls/item.json
```diff
@@ -58,6 +58,10 @@
         "callNumber": {
           "type": "string",
           "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+        },
+        "callNumberSuffix": {
+          "type": "string",
+          "description": "Effective Call Number Suffix is the suffix of the identifier assigned to an item or its holding and associated with the item."
```
marcjohnson-kint: Is this read only?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -143,22 +144,15 @@ public void deleteItemStorageItemsByItemId(
...
-          components.setCallNumber(hr.getCallNumber());
-          i.setEffectiveCallNumberComponents(components);
-          return i;
-        });
-      } else {
-        completableFuture = CompletableFuture.supplyAsync(() -> item);
-      }
+      completableFuture = getHoldingsRecordById(okapiHeaders, vertxContext, item.getHoldingsRecordId()).thenCombineAsync(CompletableFuture.supplyAsync(() -> item), (hr, i) ->
+      {
+        i.setEffectiveCallNumberComponents(EffectiveCallNumberComponentsUtil.buildComponents(hr, item));
```
marcjohnson-kint: Why is there a need to refer to `i` and `item` are these not the same item?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -185,4 +179,9 @@ public void deleteItemStorageItemsByItemId(
 
     return readItemFuture;
   }
+
+  private boolean shouldNotRetreiveHoldingsRecord(Item item) {
```
marcjohnson-kint:
```suggestion
  private boolean shouldNotRetrieveHoldingsRecord(Item item) {
```

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
rladdusaw: It seems like it is available in the context of the lambda function so I'm not sure why it's being passed in like that. It seems to work without supplying it, so I think we can remove the redundancy altogether.

ramls/item.json
rladdusaw: Yes it should be.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,42 @@
...
+
+    String updatedCallNumber = null;
+    if (StringUtils.isNotBlank(item.getItemLevelCallNumber())) {
+      updatedCallNumber = item.getItemLevelCallNumber();
+    } else if (StringUtils.isNotBlank(holdings.getCallNumber())) {
+      updatedCallNumber = holdings.getCallNumber();
+    }
+
+    String updatedCallNumberSuffix = null;
+    if (StringUtils.isNotBlank(item.getItemLevelCallNumberSuffix())) {
```
bohdan-suprun: Hi @marcjohnson-kint, @mreno-EBSCO, @rladdusaw,

I was reviewing circulation logic and spotted that there we have a bit different logic to for effective suffix/prefix. We have just a single check on `holdings.callNumber` and `item.itemLevelCallNumber` and depending on the result, we pick all the components from either item or holding (not per-component logic). This looks to me more correct, since item and holdings call number may be of different type and mixing them looks incorrect.

Maybe I'm missing something.

marcjohnson-kint: @bohdan-suprun 

I think that is a question for Cate, please write a comment on the JIRA issue.

My understanding is that we do want to mix the properties at different levels. It is worth confirming that with Cate.

The logic defined in inventory should supersede the interim logic put into circulation by Vega for notices.

bohdan-suprun: Hi @rladdusaw, @marcjohnson-kint, Cate responded that we should use per-component logic, so no work observed here.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
     assertThat(updatedItemFromGet.getJsonObject("effectiveCallNumberComponents").getString("callNumber"), is("itemLevelCallNumber"));
   }
 
+  @Test
+  public void updatingHoldingsUpdatesItemEffectiveCallNumberSuffix()
```
marcjohnson-kint: Could another test be added to check that an item under a different holdings does not get updated?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
...
+    itemToCreate.put("temporaryLocationId", annexLibraryLocationId.toString());
+    itemToCreate.put("materialTypeId", bookMaterialTypeID.toString());
+
+    Response postFirstItemResponse = create(itemsStorageUrl(""), itemToCreate);
+    Response postSecondItemResponse = create(itemsStorageUrl(""), itemToCreate);
+
+    assertThat(postFirstItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+    assertThat(postSecondItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    JsonObject firstItemFromPost = postFirstItemResponse.getJson();
```
marcjohnson-kint:
```suggestion
    JsonObject firstItem = postFirstItemResponse.getJson();
```

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
...
+    JsonObject secondItemFromPost = postSecondItemResponse.getJson();
```
marcjohnson-kint:
```suggestion
    JsonObject secondItem = postSecondItemResponse.getJson();
```

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
...
+    assertThat(firstItemFromGet.getString("id"), is(firstItemId));
+    assertThat(firstItemFromGet.getString("holdingsRecordId"), is(holdingId.toString()));
+    assertThat(firstItemFromGet.getJsonObject("effectiveCallNumberComponents").getString("callNumberSuffix"), is("testCallNumberSuffix"));
+    assertThat(secondItemFromGet.getString("id"), is(secondItemId));
+    assertThat(secondItemFromGet.getString("holdingsRecordId"), is(holdingId.toString()));
+    assertThat(secondItemFromGet.getJsonObject("effectiveCallNumberComponents").getString("callNumberSuffix"), is("testCallNumberSuffix"));
+
+    URL holdingsUrl = holdingsStorageUrl(String.format("/%s", holdingId));
+
+    holding.remove("callNumberSuffix");
```
marcjohnson-kint: I think the `put` below will overwrite an existing value so this remove could likely be omitted

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
...
+    assertThat(getFirstUpdatedItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(firstUpdatedItemFromGet.getString("id"), is(firstItemId));
+    assertThat(firstUpdatedItemFromGet.getJsonObject("effectiveCallNumberComponents").getString("callNumberSuffix"), is("updatedCallNumberSuffix"));
+    assertThat(getSecondUpdatedItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(secondUpdatedItemFromGet.getString("id"), is(secondItemId));
+    assertThat(secondUpdatedItemFromGet.getJsonObject("effectiveCallNumberComponents").getString("callNumberSuffix"), is("updatedCallNumberSuffix"));
+  }
+
+  @Test
+  public void clearingHoldingsCallNumberSuffixUpdatesItemEffectiveCallNumberSuffix()
```
marcjohnson-kint:
```suggestion
  public void removingHoldingsCallNumberSuffixUpdatesItemEffectiveCallNumberSuffix()
```

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -791,6 +791,212 @@ public void holdingsCallNumberDoesNotUpdateItemWithItemLevelCallNumber()
...
+
+    JsonObject firstUpdatedItemFromGet = getFirstUpdatedItemResponse.getJson();
+
+    assertThat(getFirstUpdatedItemResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(firstUpdatedItemFromGet.getString("id"), is(firstItemId));
+    assertThat(firstUpdatedItemFromGet.getJsonObject("effectiveCallNumberComponents").containsKey("callNumberSuffix"), is(false));
+  }
+
+  @Test
+  public void holdingsCallNumberSuffixDoesNotUpdateItemWithItemLevelCallNumberSuffix()
```
marcjohnson-kint:
```suggestion
  public void holdingsCallNumberSuffixDoesNotSupersedeItemLevelCallNumberSuffix()
```

ramls/item.json
```diff
@@ -57,7 +57,13 @@
       "properties": {
         "callNumber": {
           "type": "string",
-          "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+          "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item.",
+          "readonly": true
+        },
+        "callNumberSuffix": {
```
marcjohnson-kint: As this property is within the call number components object, does it need the leading `callNumber`? Are we envisaging other suffixes?

rladdusaw: I don't have a strong preference either way. @mreno-EBSCO @bohdan-suprun opinions?

mreno-EBSCO: I'd prefer, in absence of conflict, that we remove redundant info in property names. I think `callNumberSuffix` is likely a candidate. I think `suffix` would suffice, but like you I can go either way since we've been referring to these things as `callNumberThis` and `callNumberThat` awhile. If I had to choose, I'd drop `callNumber` from these nested field names, well, obviously not for `callNumber` proper. 😄

rladdusaw:
@marcjohnson-kint Style and test changes were applied to #358 so they could be added to prefixes at the same time.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
rladdusaw: I've addressed it #358

ramls/item.json
rladdusaw: I'm going to close this PR as redundant. All the code is in #358 and it's easier to apply these updates to a single PR.

PR #358, opened 2019-11-18
Modinvstor 358, 360 effective call number prefix and suffix

rladdusaw:
- Added call number prefix to effective call number components.
- Added call number suffix to effective call number components
- Created tests for call number prefix and suffix.


src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -185,4 +179,9 @@ public void deleteItemStorageItemsByItemId(
 
     return readItemFuture;
   }
+
+  private boolean shouldNotRetrieveHoldingsRecord(Item item) {
+    return StringUtils.isNotBlank(item.getItemLevelCallNumber())
+      && StringUtils.isNotBlank(item.getItemLevelCallNumberSuffix());
```
bohdan-suprun: It seems we have to add a check for prefix here as well.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -471,15 +470,7 @@ private boolean isUUID(String id) {
...
-    String updatedCallNumber = null;
-    if (StringUtils.isNotBlank(item.getItemLevelCallNumber())) {
-      updatedCallNumber = item.getItemLevelCallNumber();
-    } else if (StringUtils.isNotBlank(holdingsRecord.getCallNumber())) {
-      updatedCallNumber = holdingsRecord.getCallNumber();
-    }
-    EffectiveCallNumberComponents components = new EffectiveCallNumberComponents();
-    components.setCallNumber(updatedCallNumber);
-    item.setEffectiveCallNumberComponents(components);
+    item.setEffectiveCallNumberComponents(EffectiveCallNumberComponentsUtil.buildComponents(holdingsRecord, item));
```
marcjohnson-kint:
```suggestion
    item.setEffectiveCallNumberComponents(
      EffectiveCallNumberComponentsUtil.buildComponents(holdingsRecord, item));
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public static EffectiveCallNumberComponents buildComponents(HoldingsRecord holdings, Item item) {
+    return updateComponents(new EffectiveCallNumberComponents(), holdings, item);
+  }
+
+  private static EffectiveCallNumberComponents updateComponents(
+    EffectiveCallNumberComponents components, HoldingsRecord holdings, Item item) {
+
+    String updatedCallNumber = null;
```
marcjohnson-kint: Given the `EffectiveCallNumberComponents` class is mutable, I'm not sure there is a need for these intermediary variables.

This is very minor and doesn't stop an approval, I just wanted to mention it

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,51 @@
...
+    if (StringUtils.isNotBlank(item.getItemLevelCallNumber())) {
```
marcjohnson-kint: I find it easier to read statements like this with static imports, as I find the utility class names distracting (and long for extension methods, that at first I really disliked).

```suggestion
    if (isNotBlank(item.getItemLevelCallNumber())) {
```

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
```diff
@@ -1,25 +1,12 @@
...
-        CASE
-          WHEN it.jsonb->'effectiveCallNumberComponents' IS NULL THEN
-            jsonb_set(it.jsonb, '{effectiveCallNumberComponents}',
-              jsonb_build_object('callNumber', hr.jsonb->'callNumber'))
-          ELSE
-            jsonb_set(it.jsonb, '{effectiveCallNumberComponents,callNumber}', hr.jsonb->'callNumber')
-        END
-    ELSE it.jsonb
-  END
+SET jsonb = jsonb_set(
```
marcjohnson-kint: This is neat, thank you for consolidating this logic.

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
```diff
@@ -1,25 +1,12 @@
...
+  it.jsonb,
+  '{effectiveCallNumberComponents}',
+  jsonb_build_object(
+    'callNumber', COALESCE(it.jsonb->'itemLevelCallNumber', hr.jsonb->'callNumber'),
+    'prefix', COALESCE(it.jsonb->'itemLevelCallNumberPrefix', hr.jsonb->'callNumberPrefix'),
+    'suffix', COALESCE(it.jsonb->'itemLevelCallNumberSuffix', hr.jsonb->'callNumberSuffix')
+  )
+)
```
marcjohnson-kint: This is neat, thank you for tidying up this logic

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,51 @@
...
+import org.folio.rest.jaxrs.model.HoldingsRecord;
+import org.folio.rest.jaxrs.model.Item;
+
+/**
+ * Utility methods for Effective Call Number Components object.
+ */
+public final class EffectiveCallNumberComponentsUtil {
+
+  private EffectiveCallNumberComponentsUtil() {
+    throw new UnsupportedOperationException("Do not instantiate");
```
marcjohnson-kint: I usually omit the exception, and only have an empty constructor. I accept that folks could use reflection to invoke it.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,51 @@
+package org.folio.rest.support;
+
+import org.apache.commons.lang.StringUtils;
+import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
+import org.folio.rest.jaxrs.model.HoldingsRecord;
+import org.folio.rest.jaxrs.model.Item;
+
+/**
+ * Utility methods for Effective Call Number Components object.
+ */
```
marcjohnson-kint: Personally, I don't think this comment adds much over the name of the class.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -407,6 +407,326 @@ public void canUpdateAnItemWithItemLevelCallNumberAsEffectiveCallNumber()
...
+    assertThat(itemFromGet.getString("permanentLoanTypeId"),
+      is(canCirculateLoanTypeID));
+    assertThat(itemFromGet.getString("temporaryLocationId"),
+      is(annexLibraryLocationId.toString()));
+
+    List<String> tags = itemFromGet.getJsonObject("tags").getJsonArray("tagList").getList();
+
+    assertThat(tags.size(), is(1));
+    assertThat(tags, hasItem(TAG_VALUE));
+    assertThat(itemFromGet.getJsonObject("effectiveCallNumberComponents").getString("suffix"), is("testItemCallNumberSuffix"));
```
marcjohnson-kint:
```suggestion
    assertThat(
       itemFromGet.getJsonObject("effectiveCallNumberComponents").getString("suffix"), 
       is("testItemCallNumberSuffix"));
```

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
```diff
@@ -110,6 +110,142 @@ public void canInitializeEffectiveCallNumberAfterItemChange() throws Exception {
     assertThat(getItem(item.getId()).getEffectiveCallNumberComponents().getCallNumber(), is("updatedCallNumber"));
   }
 
+  @Test
+  public void canInitializeEffectiveCallNumberPrefix() throws Exception {
+
```
marcjohnson-kint: Personally, I'd remove these blank lines from the beginning of the tests

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
```diff
@@ -110,6 +110,142 @@ public void canInitializeEffectiveCallNumberAfterItemChange() throws Exception {
...
+    assertThat(getItem(item.getId()).getEffectiveCallNumberComponents().getPrefix(), is("testCallNumberPrefix"));
+
+    runSql(POPULATE_EFFECTIVE_CALL_NUMBER_SQL);
+
+    assertThat(getHoldings(holding).getCallNumberPrefix(), is("updatedCallNumberPrefix"));
+    assertThat(getItem(item.getId()).getEffectiveCallNumberComponents().getPrefix(), is("updatedCallNumberPrefix"));
+  }
+
+  @Test
+  public void canInitializeEffectiveCallNumberPrefixAfterItemChange() throws Exception {
```
marcjohnson-kint: I'm not sure I understand this test. As I understand it, these are for testing the migration script. And that is a static operation, why is it relevant that an item cold change?

I think this is demonstrating that the initial value should be the item level prefix when it is present?

If so, I suggest simplifying this and renaming it to
```suggestion
  public void canInitializeEffectiveCallNumberPrefixToItemLevelWhenPresent() throws Exception {
```

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -407,6 +407,326 @@ public void canUpdateAnItemWithItemLevelCallNumberAsEffectiveCallNumber()
...
+    client.put(itemsStorageUrl("/" + itemId), itemToCreate, StorageTestSuite.TENANT_ID,
+        ResponseHandler.text(completed));
+
+    Response getResponse = getById(UUID.fromString(itemId));
+
+    assertThat(getResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+
+    JsonObject itemFromGet = getResponse.getJson();
+
+    assertThat(itemFromGet.getJsonObject("effectiveCallNumberComponents").getString("prefix"), is("testCallNumberPrefix"));
```
marcjohnson-kint:
```suggestion
    assertThat(
      itemFromGet.getJsonObject("effectiveCallNumberComponents").getString("prefix"), 
      is("testCallNumberPrefix"));
```

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
```diff
@@ -44,53 +44,64 @@ public static void createInstance() throws Exception {
...
     runSql(POPULATE_EFFECTIVE_CALL_NUMBER_SQL);
 
-    assertThat(getItem(item.getId()).getEffectiveCallNumberComponents().getCallNumber(), is("testCallNumber"));
+    assertThat(
+      getItem(item.getId()).getEffectiveCallNumberComponents().getCallNumber(),
+      is("testCallNumber"));
   }
 
   @Test
   public void canInitializeEffectiveCallNumberAfterHoldingsChange() throws Exception {
```
marcjohnson-kint: Similarly to the other test, I don’t understand why the update is important, as these scripts are to be run on upgrade, where the records are already fixed?

marcjohnson-kint: Similarly to the other test, I don’t understand why the update is important, as these scripts are to be run on upgrade, where the records are already fixed?

ramls/item.json
```diff
@@ -57,7 +57,18 @@
       "properties": {
         "callNumber": {
           "type": "string",
-          "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item."
+          "description": "Effective Call Number is an identifier assigned to an item or its holding and associated with the item.",
+          "readonly": true
+        },
+        "prefix": {
```
marcjohnson-kint: These new properties need reflecting in the interface versions (in both the descriptor and RAML files)

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
rladdusaw: Since I have to use sql for any interaction with the database that avoids the business logic it was easier to update an existing item than to create a new valid item. My concern was to show the script updates the value of the components, and an existing one seemed just as good for that.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -143,22 +144,15 @@ public void deleteItemStorageItemsByItemId(
 
   private CompletableFuture<Item> setEffectiveCallNumber(Map<String, String> okapiHeaders, Context vertxContext, Item item) {
     CompletableFuture<Item> completableFuture = null;
-    EffectiveCallNumberComponents components = new EffectiveCallNumberComponents();
-    if (item.getItemLevelCallNumber() != null && !item.getItemLevelCallNumber().isEmpty()) {
-      components.setCallNumber(item.getItemLevelCallNumber());
-      item.setEffectiveCallNumberComponents(components);
+    if (shouldNotRetrieveHoldingsRecord(item)) {
+      item.setEffectiveCallNumberComponents(EffectiveCallNumberComponentsUtil.buildComponents(null, item));
       completableFuture = CompletableFuture.supplyAsync(() -> item);
```
mreno-EBSCO: I know this line isn't part of this PR, but it seems odd here to be using a Supplier for an value we already have. There is an overhead to creating a lambda and using the forkjoin pool to asynchronously supply the value. I think it makes sense to simplify this to:
```suggestion
      completableFuture = CompletableFuture.completedFuture(item);
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,44 @@
+package org.folio.rest.support;
+
+import static org.apache.commons.lang.StringUtils.isNotBlank;
```
mreno-EBSCO:
```suggestion
import static org.apache.commons.lang3.StringUtils.isNotBlank;
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,44 @@
...
+public final class EffectiveCallNumberComponentsUtil {
+
+  public static EffectiveCallNumberComponents buildComponents(HoldingsRecord holdings, Item item) {
+    return updateComponents(new EffectiveCallNumberComponents(), holdings, item);
+  }
+
+  private static EffectiveCallNumberComponents updateComponents(
+    EffectiveCallNumberComponents components, HoldingsRecord holdings, Item item) {
+
+    String updatedCallNumber = null;
```
mreno-EBSCO: Maybe this instead:
```suggestion
    String updatedCallNumber = StringUtils.firstNonBlank(item.getItemLevelCallNumber(), holdings.getCallNumber());
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,44 @@
...
+    if (isNotBlank(item.getItemLevelCallNumber())) {
+      updatedCallNumber = item.getItemLevelCallNumber();
+    } else if (isNotBlank(holdings.getCallNumber())) {
+      updatedCallNumber = holdings.getCallNumber();
+    }
+
+    String updatedCallNumberPrefix = null;
```
mreno-EBSCO: Maybe this instead:
```suggestion
    String updatedCallNumberPrefix = StringUtils.firstNonBlank(item.getItemLevelCallNumberPrefix(), holdings.getCallNumberPrefix());
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -0,0 +1,44 @@
...
+    if (isNotBlank(item.getItemLevelCallNumberPrefix())) {
+      updatedCallNumberPrefix = item.getItemLevelCallNumberPrefix();
+    } else if (isNotBlank(holdings.getCallNumberPrefix())) {
+      updatedCallNumberPrefix = holdings.getCallNumberPrefix();
+    }
+
+    String updatedCallNumberSuffix = null;
```
mreno-EBSCO: Maybe this instead:
```suggestion
    String updatedCallNumberSuffix = StringUtils.firstNonBlank(item.getItemLevelCallNumberSuffix(), holdings.getCallNumberSuffix());
```

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
bohdan-suprun: It seems that `holdings` will be null here if `shouldNotRetrieveHoldings` method returns `true`, so possible NPE here.

rladdusaw: If we do this we will just end up replacing the string logic with unwrapping the optional. It seems to be pretty equivalent to me.

mreno-EBSCO: Good point. Though, I think using `Optional` adds some complexity. Instead of sending `null`, we could send `new HoldingsRecord()`. Nothing would need to change in this code. Since the public static method is a utility method, we may want to add `Objects.requireNonNull()` checks for the holdings and item arguments to indicate to a user of this utility class that they cannot send `null` objects to the public method.

rladdusaw: I haven't done something like this is in the ItemStorageAPI.java because I'm passing null for the holdings after explicitly checking that all the item properties exist to avoid fetching the holdings unnecessarily.

mreno-EBSCO: Right, but to Bohdan's point, when we call `holdings.getCallNumber()` below we'll get a NPE since `holdings` is `null`. So, my thought is that in `ItemStorageAPI`, instead of calling:
```java
     item.setEffectiveCallNumberComponents(EffectiveCallNumberComponentsUtil.buildComponents(null, item));
```
we could do something like:
```java
     item.setEffectiveCallNumberComponents(EffectiveCallNumberComponentsUtil.buildComponents(new HoldingsRecord(), item));
```
Then `holdings.getCallNumber()` won't NPE and will return `null` since the `HoldingsRecord` has nothing set. We already have checked that we don't need holdings info. It may be a bit of a hack, but I've seen worse. We could create a constant `EMPTY_HOLDINGS_RECORD` or something like that to pass in if it seems strange to create a new `HoldingsRecord` after determining that we don't need one. Another option is another static method here without the holdings record argument and pass in a newly constructed `HoldingsRecord`. At least that takes some of the confusion out of the code using the utility.
```java
  public static EffectiveCallNumberComponents buildComponents(Item item) {
    return updateComponents(new EffectiveCallNumberComponents(), new HoldingsRecord(), item);
  }
```

PR #359, opened 2019-11-19
MODINVSTOR-337 Change sample instance source to FOLIO

marcjohnson-kint:
*Purpose*

This change involves changing the `source` property for all sample instances to `FOLIO`. 

The existing `MARC` value was confusing since the MARC records within mod-inventory-storage are no longer used and there are no corresponding MARC source records in mod-source-record-storage.

Given that the sample source records functionality has been superseded, the sample records for these have also been removed to reduce confusion. The API should also be removed as a separate issue.

*Approach*
* Remove the sample MARC instance source records
* Remove the loading of sample MARC instance 
* Change the sample instance records

I used `jq` to change the sample instance records (see script temporarily committed). This resulted in some reformatting of the sample instances.

I am aware it is preferable to separate reformatting from changes, in order to avoid distraction during review. I folks feel strongly about that, I will close this pull request and issue a separate pull request for reformatting first.

I also took the liberty to remove the `hrid` property from one of the sample records as this is soon to be invalid. Once again, if this is a point of contention, I will remove this change from this pull request.


sample-data/instances/semantic-web-primer.json
```diff
@@ -1,53 +1,73 @@
 {
-  "id" : "5bf370e0-8cca-4d9c-82e4-5170ab2a0a39",
-  "source" : "manual",
```
mreno-EBSCO: What was a `manual` source?

marcjohnson-kint:
Thanks @mreno-EBSCO

sample-data/instances/semantic-web-primer.json
marcjohnson-kint: I don't know. I think it was a (potentially unofficial) precursor to the `FOLIO` source. I think it referred to the record being manually created in the inventory UI and has no source record.

PR #360, opened 2019-11-20
MODINVSTOR-373: instance APIs need to set the instance HRID if not set

mreno-EBSCO:
Implement MODINVSTOR-373 to require instance POST APIs to check the submitted instance(s) and if the `hrid` field is not set, we generate a new HRID and update the instance with this HRID. If the `hrid` field is present, we preserve the submitted HRID. For the PUT API, we no longer allow the HRID to be updated. We replace or add the `hrid` from the existing stored instance into the submitted instance.

This logic was applied to the instance API, the batch instances API and the synchronous batch instances API.

While creating the tests, some tests, especially the batch APIs, failed due to a request timeout which allows operations to bleed into other tests causing cascading failures. To minimize this, the request wait timeout was increased on several tests. Also noticed some cleanup was being done asynchronously without waiting for completion. Fixed that as well. The instance storage units now complete consistently on my over-taxed laptop.

Note: as described in a commit comment, `commons-lang3` was added as an explicit dependency. Previously, it was being used via an implicit dependency of some explicit dependency, I believe RMB. We are also using `commons-lang`, which is a legacy library at this point. We'll need to move away from this library via a tech-debt story.


marcjohnson-kint:
Thank you for improving the stability of the tests

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1783,6 +1805,510 @@ public void cannotPostSynchronousBatchWithDuplicateId() throws Exception {
...
+    final CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+
+    client.post(instancesStorageUrl(""), instanceToCreateWithSameHRID, TENANT_ID,
+      text(createCompleted));
+
+    final Response response = createCompleted.get(5, SECONDS);
+
+    assertThat(response.getStatusCode(), is(400));
+    assertThat(response.getBody(),
+        is("duplicate key value violates unique constraint \"instance_hrid_idx_unique\": Key (lower(f_unaccent(jsonb ->> 'hrid'::text)))=(in00000001) already exists."));
```
marcjohnson-kint: Are we able to replace this message with a more human friendly message?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1783,6 +1805,510 @@ public void cannotPostSynchronousBatchWithDuplicateId() throws Exception {
...
+    final JsonObject instance = getById(id).getJson();
+    final String expectedHRID = "in00000001";
+
+    assertThat(instance.getString("hrid"), is(expectedHRID));
+
+    instance.put("hrid", "testHRID");
+
+    final JsonObject updatedInstance = updateInstance(instance).getJson();
+
+    assertThat(updatedInstance.getString("hrid"), is(expectedHRID));
```
marcjohnson-kint: Does this mean that any new HRID is silently ignored, rather than the whole request being rejected?

mreno-EBSCO: Yes. We could do this either way. Does it make more sense to reject the request with a message stating the `hrid` field cannot be changed or ignore the change and complete the update? I suppose if the only change is the `hrid` field, this could be a bit confusing.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
mreno-EBSCO: Sure. I was using `PgExceptionUtil` here because I thought that might be the standard way we want to send back SQL errors so the UI can deal with them. I also wanted to avoid processing this string. It is a pretty gnarly message, but it is somewhat consistent across all 3 instance creation APIs, each which a different format of error message/structure. I'll look into making a more human readable message that will work across all 3 APIs.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: If the `hrid` was readonly in the schema because we used different schema for reads and writes, then I'd suggest ignoring it would be best.

As it isn't, I think if it is no more work to implement, then refusing the request is better. It allows the client to understand. Most of the time, the client should be basing their PUT request on a recently fetched representation, so it should be the same.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
marcjohnson-kint: I agree that depending upon parsing database error text is not ideal, and potentially fiddly and brittle.

I think that is preferable to sending raw database errors to the client (or at the least, only those).

src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
```diff
@@ -118,29 +123,56 @@ public void postInstanceStorageInstances(
...
+                        PostInstanceStorageInstancesResponse
+                          .respond201WithApplicationJson(entity,
+                              PostInstanceStorageInstancesResponse.headersFor201().withLocation(reply.result()))));
+                  }
+                  else {
+                    if (PgExceptionUtil.isUniqueViolation(reply.cause())) {
+                      asyncResultHandler.handle(
+                          io.vertx.core.Future.succeededFuture(
+                            PostInstanceStorageInstancesResponse
+                              .respond400WithTextPlain(PgExceptionUtil.badRequestMessage(reply.cause()))));
```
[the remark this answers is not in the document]
mreno-EBSCO: Agreed. The problem here is that the `instance-storage` API does not use the validation RAML trait and therefore is not supposed to return 422 (no `respond422XXX` method in the interface). Of course, we could create our own response, but again, it would be unexpected. Since this is a major, at least in my mind, change to the interface, I was reluctant to make this change. I think we need a separate story (which we do already have, in fact, since I created it about 2 years ago 😈) to add/support validation to all mod-inventory-storage APIs (some other APIs have validation enabled but do not have the Validate annotation on the impl methods in order for validation to work).

src/test/java/org/folio/rest/api/InstanceStorageTest.java
mreno-EBSCO: @marcjohnson-kint the more I look into this, the more challenging this appears to be. For the single API call, it may not be so bad, but to make this work in the sync batch API, I need to either modify the RMB ValidationHelper or put some special case logic in StorageHelper (at least it is in this module), which is used by all the sync APIs. I can do it, but it will take some time. I think the same thing is true for the older batch API as the single instance API, the code is just entirely different.

Do you feel strongly about hiding the SQL for this PR? Maybe we just modify the single instance case and let the batch APIs return the SQL message? What do you think? I am concerned we are creating a special case for this one error and all other SQL errors are just errors with SQL messages. It'd be nice if we had a SQL to human readable message mapper, which is what I was hoping `PgExceptionUtil` would do.

marcjohnson-kint: Not so strongly as to delay this work to that degree.

When I was introducing the early database constraints, prior to the exception until, I did some messy string matching.

I think this probably needs more thought than this individual work really allows for.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1783,6 +1811,524 @@ public void cannotPostSynchronousBatchWithDuplicateId() throws Exception {
...
+
+    assertThat(response.getStatusCode(), is(500));
+    assertThat(response.getBody(),
+        is("ErrorMessage(fields=[(Severity, ERROR), (V, ERROR), (SQLSTATE, 2200H), (Message, nextval: reached maximum value of sequence \"hrid_instances_seq\" (99999999)), (File, sequence.c), (Line, 700), (Routine, nextval_internal)])"));
+
+    log.info("Finished cannotCreateInstanceWithHRIDFailure");
+  }
+
+  @Test
+  public void cannotChageHRIDAfterCreation() throws Exception {
```
```suggestion
  public void cannotChangeHRIDAfterCreation() throws Exception {
```

PR #361, opened 2019-11-22
[BUGFIX] MODINVSTOR-376: Fix date time format.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-376 - Instance record. Status updated date does not populate in detailed view.

Previously `statusUpdatedDate` was returned with no time zone information, which causes UI to display it in a wrong time zone. 

This PR changes `statusUpdatedDate` to has `yyyy-MM-dd'T'HH:mm:ss.SSS'+0000'` pattern (the same that used for metadata fields).

### DIFF
--- a/src/test/java/org/folio/rest/support/matchers/DateTimeMatchers.java
+++ b/src/test/java/org/folio/rest/support/matchers/DateTimeMatchers.java
@@ -1,11 +1,15 @@
 package org.folio.rest.support.matchers;
 
+import java.util.Arrays;
+import java.util.List;
+
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeZone;
 import org.joda.time.Seconds;
+import org.joda.time.format.DateTimeFormat;
 
 public final class DateTimeMatchers {
 
@@ -37,5 +41,36 @@ public final class DateTimeMatchers {
 
   public static Matcher<String> withinSecondsBeforeNow(Seconds seconds) {
     return withinSecondsBefore(seconds, DateTime.now(DateTimeZone.UTC));
+  }
+
+  public static Matcher<String> hasIsoFormat() {
+    List<String> acceptableFormats = Arrays.asList(
+      "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+      "yyyy-MM-dd'T'HH:mm:ss.SSS+0000"
+    );
+
+    return new TypeSafeMatcher<String>() {
+      @Override
+      protected boolean matchesSafely(String dateTimeAsString) {
+        return acceptableFormats.stream()
+          .anyMatch(dateTimeFormat -> {
+            try {
+              DateTimeFormat.forPattern(dateTimeFormat)
+                .parseDateTime(dateTimeAsString);
+            } catch (IllegalArgumentException ex) {
+              return false;
+            }
+
+            return true;
+          });
+      }
+
+      @Override
+      public void describeTo(Description description) {
+        description
+          .appendText("Has an ISO-8601 format:")
+          .appendValue(String.join("; ", acceptableFormats));
+      }
+    };
   }
 }
--- a/src/test/java/org/folio/rest/support/matchers/DateTimeMatchers.java
+++ b/src/test/java/org/folio/rest/support/matchers/DateTimeMatchers.java
@@ -1,8 +1,5 @@
 package org.folio.rest.support.matchers;
 
-import java.util.Arrays;
-import java.util.List;
-
 import org.hamcrest.Description;
 import org.hamcrest.Matcher;
 import org.hamcrest.TypeSafeMatcher;
@@ -44,32 +41,26 @@ public final class DateTimeMatchers {
   }
 
   public static Matcher<String> hasIsoFormat() {
-    List<String> acceptableFormats = Arrays.asList(
-      "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
-      "yyyy-MM-dd'T'HH:mm:ss.SSS+0000"
-    );
+    String acceptableFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
 
     return new TypeSafeMatcher<String>() {
       @Override
       protected boolean matchesSafely(String dateTimeAsString) {
-        return acceptableFormats.stream()
-          .anyMatch(dateTimeFormat -> {
-            try {
-              DateTimeFormat.forPattern(dateTimeFormat)
-                .parseDateTime(dateTimeAsString);
-            } catch (IllegalArgumentException ex) {
-              return false;
-            }
+        try {
+          DateTimeFormat.forPattern(acceptableFormat)
+            .parseDateTime(dateTimeAsString);
+        } catch (IllegalArgumentException ex) {
+          return false;
+        }
 
-            return true;
-          });
+        return true;
       }
 
       @Override
       public void describeTo(Description description) {
         description
-          .appendText("Has an ISO-8601 format:")
-          .appendValue(String.join("; ", acceptableFormats));
+          .appendText("Has ISO-8601 format:")
+          .appendValue(acceptableFormat);
       }
     };
   }
--- a/src/main/resources/templates/db_scripts/instanceStatusUpdatedDateTrigger.sql
+++ b/src/main/resources/templates/db_scripts/instanceStatusUpdatedDateTrigger.sql
@@ -3,8 +3,11 @@ RETURNS trigger
 AS $function$
 	BEGIN
 		IF (OLD.jsonb->'statusId' IS DISTINCT FROM NEW.jsonb->'statusId') THEN
-			-- Date time in "yyyy-MM-dd'T'HH:mm:ss.SSS" format at UTC (00:00) time zone
-			NEW.jsonb = jsonb_set(NEW.jsonb, '{statusUpdatedDate}', to_jsonb(CURRENT_TIMESTAMP(3) AT TIME ZONE 'UTC'));
+			-- Date time in "yyyy-MM-dd'T'HH:mm:ss.SSS'+0000'" format at UTC (00:00) time zone
+			NEW.jsonb = jsonb_set(
+		    NEW.jsonb, '{statusUpdatedDate}',
+			  to_jsonb(to_char(CURRENT_TIMESTAMP(3) AT TIME ZONE 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.ms"+0000"'))
+		  );
 		END IF;
 		RETURN NEW;
 	END;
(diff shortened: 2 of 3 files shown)


src/test/java/org/folio/rest/support/matchers/DateTimeMatchers.java
```diff
@@ -38,4 +42,35 @@ protected boolean matchesSafely(String textRepresentation) {
   public static Matcher<String> withinSecondsBeforeNow(Seconds seconds) {
     return withinSecondsBefore(seconds, DateTime.now(DateTimeZone.UTC));
   }
+
+  public static Matcher<String> hasIsoFormat() {
+    List<String> acceptableFormats = Arrays.asList(
+      "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'",
+      "yyyy-MM-dd'T'HH:mm:ss.SSS+0000"
```
marcjohnson-kint: My interpretation is that [ISO8601 section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) is that hour and minute are intended to be separated by a colon. 

This format does not include the colon, why is that? Have I misinterpreted the specification?

I recall @mreno-EBSCO leaving a note somewhere else about an issue like this I think. Does java support a different subset of ISO8601?

bohdan-suprun: Actually they're `HH:mm:ss.SSS`

marcjohnson-kint: Apologies for not being sufficiently specific, I was referring to the time offset.

It shows as `+0000`, and my interpretation of the specification is that it allows plus or minus, and has a non-optional colon between hours and minutes.

bohdan-suprun: got it, well, this is the same format, that we use within circulation module, for example metadata:
```
"metadata": {
    "createdDate": "2019-11-25T05:44:32.653+0000",
    "createdByUserId": null,
    "updatedDate": "2019-11-25T05:44:32.653+0000",
    "updatedByUserId": null
}
```

Maybe it worth asking UI guys about supported formats (the format with `+0000` is 100% supported, since we already use this format).

Also, when we specify `"format": "date-time"` for this field, then the value will be deserialized with this format.

marcjohnson-kint: > got it, well, this is the same format, that we use within circulation module

What do you mean by within the circulation module? Can you give me an example that isn't the change metadata, as that is controlled by the RAML Module Builder.

, for example metadata:
> 
> ```
> "metadata": {
>     "createdDate": "2019-11-25T05:44:32.653+0000",
>     "createdByUserId": null,
>     "updatedDate": "2019-11-25T05:44:32.653+0000",
>     "updatedByUserId": null
> }
> ```

Maybe this issue is bigger than this issue. That example above is incorrect in at least two ways, with regards to the schema, in my interpretation.

@bohdan-suprun @mreno-EBSCO I'd appreciate your thoughts on my interpretation of the specification.

If my interpretation is correct, then the format used in that change metadata example is invalid.

> Maybe it worth asking UI guys about supported formats (the format with `+0000` is 100% supported, since we already use this format).

Sure, if this issue is broader than I anticipated, we should ask folks like @zburke @mkuklis and @Baroquem for their thoughts.

> Also, when we specify `"format": "date-time"` for this field, then the value will be deserialized with this format.

By deserialized what do you mean? This is the value that is written out into a response by the storage modules?

bohdan-suprun: Ok, I checked `itemDamagedStatusDate` and it got stored as `2019-11-25T11:35:44.332Z`, probably will remove this pattern with +0000 and stick with `Z` format.

@Baroquem, @zburke, @mkuklis, do you think this format will be correctly handled when converting to a time zone?

Because now, when I update a status for an instance 

Record Last Updated Date is: `11/25/2019 6:41 AM`
But `Status updated date` is converted to: `11/25/2019 4:41 AM` but they have to be the same values.

zburke: The [JavaScript spec](http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15) expects `YYYY-MM-DDTHH:mm:ss.sssZ` where "`Z` is the time zone offset specified as "`Z`" (for UTC) or either “`+`” or “`-`” followed by a time expression `HH:mm` ... The value of an absent time zone offset is “`Z`”."

As noted in the [stripes i18n best practices documentation](https://github.com/folio-org/stripes/blob/master/doc/i18n.md#introduction) it is generally the UI's expectation that date and time information to/from the backend will be specified in UTC. That said, the UI will correctly localize strings with non-UTC offsets as above. 

Please let me know if you need additional details.

bohdan-suprun: Thanks @zburke, data will be with `Z`.

mreno-EBSCO: @marcjohnson-kint, the Java 8 time package formatter ISO_OFFSET_DATE_TIME requires the offset to be "HH:SS" which is the "extended" offset format for ISO 8601. This can be overcome via customization of the formatter.

> **Offset X and x**: This formats the offset based on the number of pattern letters. One letter outputs just the hour, such as '+01', unless the minute is non-zero in which case the minute is also output, such as '+0130'. Two letters outputs the hour and minute, without a colon, such as '+0130'. Three letters outputs the hour and minute, with a colon, such as '+01:30'. Four letters outputs the hour and minute and optional second, without a colon, such as '+013015'. Five letters outputs the hour and minute and optional second, with a colon, such as '+01:30:15'. Six or more letters throws IllegalArgumentException. Pattern letter 'X' (upper case) will output 'Z' when the offset to be output would be zero, whereas pattern letter 'x' (lower case) will output '+00', '+0000', or '+00:00'.

I probably should have looked into it a bit more before combining java.time and JodaTime where that comment is. In edge-sip2 we don't use JodaTime at all and so we created a formatter that handles the cases we need to handle:
```java
  public static DateTimeFormatter getFolioDateTimeFormatter() {
    return new DateTimeFormatterBuilder()
        .parseCaseInsensitive()
        .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
        .appendPattern("[XXX][XX][X]")
        .toFormatter();
  }
``` 

As you can see from the builder, it handles these scenarios:
```
Zone offset: Pattern letters to output ZoneOffset.

  Pattern  Count  Equivalent builder methods
  -------  -----  --------------------------
    O       1      appendLocalizedOffsetPrefixed(TextStyle.SHORT);
    OOOO    4      appendLocalizedOffsetPrefixed(TextStyle.FULL);
    X       1      appendOffset("+HHmm","Z")
    XX      2      appendOffset("+HHMM","Z")
    XXX     3      appendOffset("+HH:MM","Z")
    XXXX    4      appendOffset("+HHMMss","Z")
    XXXXX   5      appendOffset("+HH:MM:ss","Z")
    x       1      appendOffset("+HHmm","+00")
    xx      2      appendOffset("+HHMM","+0000")
    xxx     3      appendOffset("+HH:MM","+00:00")
    xxxx    4      appendOffset("+HHMMss","+0000")
    xxxxx   5      appendOffset("+HH:MM:ss","+00:00")
    Z       1      appendOffset("+HHMM","+0000")
    ZZ      2      appendOffset("+HHMM","+0000")
    ZZZ     3      appendOffset("+HHMM","+0000")
    ZZZZ    4      appendLocalizedOffset(TextStyle.FULL);
    ZZZZZ   5      appendOffset("+HH:MM:ss","Z")
```

We could all of the 'X' patterns in case we need to store dates with a seconds based offset. I'm not sure there are any though. I don't see any in zone offset lists I have looked at.

marcjohnson-kint: @mreno-EBSCO Thanks for the detailed follow up. I think I might be failing to follow all of it.

For parsing input, I think supporting a range of ISO8601 variants likely makes sense, with regards to Postel's law. I believe this is what many of the libraries do in practice. Are you comments above stating that the Java 8 core libraries have to be configured to do so?

For producing output, I think we need to be specific in the format we produce. 

My interpretation of the JSON schema specification is that it is based upon RFC3339 which I interpret to use either `Z` or a positive or negative offset with a colon between hours and minutes. It is a fairly strict variant of ISO8601.

@bohdan-suprun @mreno-EBSCO Is my interpretation reasonable?

@zburke Indeed, I think the UI (as other clients) will happily parse ISO8601 variants, so a lot of this is redundant. I think there probably are some bugs in mod-circulation (and likely others) where date time properties are not moved to UTC for output.

mreno-EBSCO: @marcjohnson-kint, yes, the Java 8 java.time static ISO offset formatter, out of the box, requires the offset to be formatted"+HH:MM:ss". It has to be explicitly configured otherwise (or a custom fomatter used) or else there is an exception when parsing "+HHMM" dates, like the format we have been using being generated by JodaTime. Generating a date-time, using java.time, should produce a 'Z' in the case with zero offset.

What you are describing is what Java refers to as "ISO-8601 extended offset date-time format". I'd prefer if we were consistent as well. I'd prefer we use "+HH:MM" as that is what is typically described by authorities when dealing with offsets.

Hmmm... It looks like, maybe, this has been addressed in Java 9+. A bug report indicates they have added "+HHmmss" to the static ISO offset date time formatter, which should allow offsets like "-0500" to parse. I have not tried though, so I can't confirm. It appears that generation wise, it will still output "+HH:MM:ss" by default, so, if we ever go to java.time, we might have to deal with multiple offset formats.

marcjohnson-kint: > @marcjohnson-kint, yes, the Java 8 java.time static ISO offset formatter, out of the box, requires the offset to be formatted"+HH:MM:ss". It has to be explicitly configured otherwise (or a custom fomatter used) or else there is an exception when parsing "+HHMM" dates, like the format we have been using being generated by JodaTime. 

> Generating a date-time, using java.time, should produce a 'Z' in the case with zero offset.

The default for Java 8 java.time is to always convert the representation to UTC before writing output?
 
> What you are describing is what Java refers to as "ISO-8601 extended offset date-time format". I'd prefer if we were consistent as well. I'd prefer we use "+HH:MM" as that is what is typically described by authorities when dealing with offsets.

Right, I recall having this debate when FOLIO was first having these conversations. 

ISO8601 is a very broad (and closed) format. RFC-3339 is a very specific sub-variant of it, with the intent of simplifying the situation for most systems. Which is what JSON schema adopts for `date-time`. 

The two standards are easily confused unfortunately. In part because most libraries actually cater for a broader part of ISO8601 than RFC3339.
 
> Hmmm... It looks like, maybe, this has been addressed in Java 9+. A bug report indicates they have added "+HHmmss" to the static ISO offset date time formatter, which should allow offsets like "-0500" to parse. I have not tried though, so I can't confirm. It appears that generation wise, it will still output "+HH:MM:ss" by default, so, if we ever go to java.time, we might have to deal with multiple offset formats.

I think we are agreed that we should attempt to be tolerant in parsing. I didn't realise it could take this degree of configuration to make it do so.

PR #362, opened 2019-11-22
MODINVSTOR-374: Add holdings HRID support

mreno-EBSCO:
When creating holdings records set a generated HRID when not present.
When modifying holdings records ensure the HRID cannot be changed.

### DIFF
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -1,7 +1,10 @@
 package org.folio.rest.impl;
 
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -16,6 +19,7 @@ import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.HoldingsRecords;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.rest.jaxrs.resource.HoldingsStorage;
+import org.folio.rest.persist.PgExceptionUtil;
 import org.folio.rest.persist.PgUtil;
 import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.persist.Criteria.Criteria;
@@ -23,7 +27,9 @@ import org.folio.rest.persist.Criteria.Criterion;
 import org.folio.rest.persist.Criteria.Limit;
 import org.folio.rest.persist.Criteria.Offset;
 import org.folio.rest.persist.cql.CQLWrapper;
+import org.folio.rest.support.HridManager;
 import org.folio.rest.tools.utils.TenantTool;
+import org.folio.rest.tools.utils.ValidationHelper;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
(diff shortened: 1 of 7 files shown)


src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -333,47 +363,60 @@ public void putHoldingsStorageHoldingsByHoldingsRecordId(
...
-                                  postgresClient.endTx(connection, done -> {
-                                    asyncResultHandler.handle(
-                                      Future.succeededFuture(
-                                        PutHoldingsStorageHoldingsByHoldingsRecordIdResponse
-                                          .respond204()));
-                                  });
-                                }
-                                else {
+                      final HoldingsRecord existingHoldings = holdingsList.get(0);
+                      if (Objects.equals(entity.getHrid(), existingHoldings.getHrid())) {
```
bohdan-suprun: Does it make sense to move this validation right after `if (holdingsList.size() == 1) {` so no transaction will be started at all?

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -383,29 +426,51 @@ public void putHoldingsStorageHoldingsByHoldingsRecordId(
...
-                                .respond204()));
-                        }
-                        else {
+                  final Future<String> hridFuture;
+                  if (isBlank(entity.getHrid())) {
+                    final HridManager hridManager = new HridManager(vertxContext, postgresClient);
+                    hridFuture = hridManager.getNextHoldingsHrid();
+                  } else {
+                    hridFuture = Promise.succeededPromise(entity.getHrid()).future();
+                  }
```
bohdan-suprun: Seems we can extract this to a separate method.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -813,6 +829,443 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
       is("permanentLocationId"));
   }
 
+  @Test
+  public void canCreateAHoldingsWhenHRIDNotSupplied()
```
marcjohnson-kint: What is the difference between this test and `canCreateAHolding`? Could they be combined?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -813,6 +829,443 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
...
+
+    assertThat(response.getStatusCode(), is(HttpURLConnection.HTTP_BAD_REQUEST));
+    assertThat(response.getBody(),
+        is("The hrid field cannot be changed: new=null, old=ho00000001"));
+
+    log.info("Finished cannotRemoveHRIDAfterCreation");
+  }
+
+  @Test
+  public void canUsePutToCreateAHoldingsWhenHRIDNotSupplied()
```
marcjohnson-kint: Similarly, can this be incorporated in the regular can put to create test?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -813,6 +829,443 @@ public void cannotCreateHoldingWithoutPermanentLocation() throws Exception {
...
+          final Response response = getById(holdings.getString("id"));
+          assertExists(response, holdings);
+          assertHRIDRange(response, "ho00000001", "ho00000003");
+        });
+
+    log.info("Finished canPostSynchronousBatchWithGeneratedHRID");
+  }
+
+  @Test
+  public void canPostSynchronousBatchWithSuppliedAndHRID() {
```
marcjohnson-kint: Should this be supplied and generated?

```suggestion
  public void canPostSynchronousBatchWithSuppliedAndGeneratedHRID() {
```

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
mreno-EBSCO: I don't like modifying existing tests with new functionality. I'm afraid I'll mask regression by changing existing behavior expectations. I understand that leads to a lot of tests that do pretty much the same thing, but with different expectations. I can combine the tests.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
mreno-EBSCO: Same here, I like my unit tests to test just functionality I added and let existing tests ensure I didn't break any existing expectations. But, I can combine these tests as well.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
marcjohnson-kint: > I don't like modifying existing tests with new functionality. I'm afraid I'll mask regression by changing existing behavior expectations. I understand that leads to a lot of tests that do pretty much the same thing, but with different expectations. I can combine the tests.

I sympathise with that approach. And if these were lightweight unit tests I'd be in total agreement, because they better reflect intent. However, with hefty API tests there seems to be more of a compromise.

Hopefully in this case, the only expectation being changed is that the `hrid` has a generated value?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
marcjohnson-kint: I can very much understand doing it in multiple steps as well. I quite often expand the test suite and then contract it in multiple commits (but then I tend to use smaller commits than most folks on FOLIO)

PR #364, opened 2019-11-25
[BUGFIX]: MODINVSTOR-367: Maintain item status date even if item record is edited

VRohach:
Bugfix regarding item status date at the "item" object:
Time Zone didn't exist in this field.

### DIFF
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -8,8 +8,9 @@ AS $$
   BEGIN
   	newStatus = NEW.jsonb->'status'->>'name';
 	  IF (newStatus IS DISTINCT FROM OLD.jsonb->'status'->>'name') THEN
-	    -- Date time in "yyyy-MM-dd'T'HH:mm:ss.SSS" format at UTC (00:00) time zone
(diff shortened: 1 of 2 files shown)


bohdan-suprun:
@VRohach, let's wait a reply on https://github.com/folio-org/mod-inventory-storage/pull/361.

VRohach:
> @VRohach, let's wait a reply on #361.

Sure.

marcjohnson-kint:
What is the difference between this change and #361 ?

VRohach:
> What is the difference between this change and #361 ?

So, this PR is for the item status date. (In object "item" situated "status" object. In it, there is "date" field). But https://github.com/folio-org/mod-inventory-storage/pull/361 implemented for the instance status date. 

So, the approach is the same, but for other functionalities.

bohdan-suprun:
Hi @marcjohnson-kint, it is similar changes to #361, except for a different property. 

I think it is ready to be merged, unless you have any objection.

marcjohnson-kint:
> Hi @marcjohnson-kint, it is similar changes to #361, except for a different property.
> 
> I think it is ready to be merged, unless you have any objection.

No objection, I deleted my comment asking about the difference. Feel free to approve.

PR #365, opened 2019-11-27
MODINVSTOR-375: Add item HRID support

mreno-EBSCO:
When creating items set a generated HRID when not present.
When modifying items ensure the HRID cannot be changed.

Note: also removed some existing code smells with simple, minimal risk, changes in `ItemStorageAPI.java` which is why there are a couple of seemingly unrelated code changes.

### DIFF
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -1,13 +1,18 @@
 package org.folio.rest.api;
 
+import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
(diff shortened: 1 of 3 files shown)


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -749,6 +809,38 @@ public void canUpdateAnItemWithItemLevelCallNumberSuffixAsEffectiveCallNumberSuf
       is("testItemCallNumberSuffix"));
   }
 
+  @Test
+  public void canUpdateAnItemHRIDDoesNotChange()
```
marcjohnson-kint:
```suggestion
  public void canUpdateAnItemWhenHRIDHasNotChanged()
```

PR #372, opened 2019-12-09
MODINVSTOR-361: Populate effective call number type id on item.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-361: Store effective call number type on item record

* Store effective call number type id on item record.
* Update migration script.
* Update API version.
* Refactor existing item call number related test cases and create parametrized tests instead.

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -69,6 +69,12 @@
           "type": "string",
           "description": "Effective Call Number Suffix is the suffix of the identifier assigned to an item or its holding and associated with the item.",
           "readonly": true
+        },
+        "typeId": {
+          "type": "string",
+          "description": "Effective Call Number Type Id is the call number type id assigned to the item or associated holding.",
+          "$ref": "uuid.json",
+          "readonly": true
         }
       },
       "additionalProperties": false
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -72,7 +72,7 @@
         },
         "typeId": {
           "type": "string",
-          "description": "Effective Call Number Type Id is the call number type id assigned to the item or associated holding.",
+          "description": "Effective Call Number Type Id is the call number type id of the item, if available, otherwise that of the holding.",
           "$ref": "uuid.json",
           "readonly": true
         }
--- a/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
+++ b/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
@@ -2,6 +2,9 @@ package org.folio.rest.support;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
+import java.util.Optional;
+
+import org.apache.commons.lang3.StringUtils;
 import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.Item;
@@ -36,9 +39,17 @@ public final class EffectiveCallNumberComponentsUtil {
       updatedCallNumberSuffix = holdings.getCallNumberSuffix();
     }
 
+    String updatedCallNumberTypeId = StringUtils.firstNonBlank(
+      item.getItemLevelCallNumberTypeId(),
+      Optional.ofNullable(holdings).map(HoldingsRecord::getCallNumberTypeId)
+        .orElse(null)
+    );
+
     components.setCallNumber(updatedCallNumber);
     components.setPrefix(updatedCallNumberPrefix);
     components.setSuffix(updatedCallNumberSuffix);
+    components.setTypeId(updatedCallNumberTypeId);
+
     return components;
   }
 }
--- a/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
+++ b/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
@@ -2,8 +2,6 @@ package org.folio.rest.support;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-import java.util.Optional;
-
 import org.apache.commons.lang3.StringUtils;
 import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
@@ -41,8 +39,7 @@ public final class EffectiveCallNumberComponentsUtil {
 
     String updatedCallNumberTypeId = StringUtils.firstNonBlank(
       item.getItemLevelCallNumberTypeId(),
-      Optional.ofNullable(holdings).map(HoldingsRecord::getCallNumberTypeId)
-        .orElse(null)
+      holdings != null ? holdings.getCallNumberTypeId() : null
     );
 
     components.setCallNumber(updatedCallNumber);
--- a/src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
+++ b/src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
@@ -5,7 +5,8 @@ SET jsonb = jsonb_set(
   jsonb_build_object(
     'callNumber', COALESCE(it.jsonb->'itemLevelCallNumber', hr.jsonb->'callNumber'),
     'prefix', COALESCE(it.jsonb->'itemLevelCallNumberPrefix', hr.jsonb->'callNumberPrefix'),
-    'suffix', COALESCE(it.jsonb->'itemLevelCallNumberSuffix', hr.jsonb->'callNumberSuffix')
+    'suffix', COALESCE(it.jsonb->'itemLevelCallNumberSuffix', hr.jsonb->'callNumberSuffix'),
+    'typeId', COALESCE(it.jsonb->'itemLevelCallNumberTypeId', hr.jsonb->'callNumberTypeId')
   )
 )
 FROM ${myuniversity}_${mymodule}.holdings_record AS hr
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -698,7 +698,7 @@
     {
       "run": "after",
       "snippetPath": "populateEffectiveCallNumberComponentsForExistingItems.sql",
-      "fromModuleVersion": "17.1.0"
+      "fromModuleVersion": "18.2.0"
     },
     {
       "run": "after",
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -702,7 +702,7 @@
     {
       "run": "after",
       "snippetPath": "populateEffectiveCallNumberComponentsForExistingItems.sql",
-      "fromModuleVersion": "18.2.0"
+      "fromModuleVersion": "18.3.0"
     },
     {
       "run": "after",
--- a/src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
+++ b/src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
@@ -1,7 +1,6 @@
 package org.folio.rest.api;
 
 
-import static org.apache.commons.lang3.StringUtils.capitalize;
 import static org.folio.rest.api.ItemStorageTest.nod;
 import static org.hamcrest.CoreMatchers.is;
 import static org.junit.Assert.assertNotNull;
@@ -9,18 +8,13 @@ import static org.junit.Assert.assertThat;
 
 import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeoutException;
 
 import org.apache.commons.lang3.StringUtils;
-import org.apache.commons.lang3.tuple.ImmutablePair;
-import org.apache.commons.lang3.tuple.ImmutableTriple;
-import org.apache.commons.lang3.tuple.Pair;
-import org.apache.commons.lang3.tuple.Triple;
+import org.folio.rest.api.testdata.ItemEffectiveCallNumberComponentsTestData;
+import org.folio.rest.api.testdata.ItemEffectiveCallNumberComponentsTestData.CallNumberComponentPropertyNames;
 import org.folio.rest.support.IndividualResource;
 import org.folio.rest.support.Response;
 import org.folio.rest.support.builders.HoldingRequestBuilder;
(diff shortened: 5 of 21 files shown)


src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
```diff
@@ -5,7 +5,8 @@ SET jsonb = jsonb_set(
   jsonb_build_object(
     'callNumber', COALESCE(it.jsonb->'itemLevelCallNumber', hr.jsonb->'callNumber'),
     'prefix', COALESCE(it.jsonb->'itemLevelCallNumberPrefix', hr.jsonb->'callNumberPrefix'),
-    'suffix', COALESCE(it.jsonb->'itemLevelCallNumberSuffix', hr.jsonb->'callNumberSuffix')
+    'suffix', COALESCE(it.jsonb->'itemLevelCallNumberSuffix', hr.jsonb->'callNumberSuffix'),
+    'typeId', COALESCE(it.jsonb->'itemLevelCallNumberTypeId', hr.jsonb->'callNumberTypeId')
```
marcjohnson-kint: Can we do this here, because this script needs to be used for different versions of the module?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -698,7 +698,7 @@
     {
       "run": "after",
       "snippetPath": "populateEffectiveCallNumberComponentsForExistingItems.sql",
-      "fromModuleVersion": "17.1.0"
+      "fromModuleVersion": "18.2.0"
```
marcjohnson-kint: Hmm. How does this work, does this mean that this definition will only work for version 18.2.0?

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+@RunWith(JUnitParamsRunner.class)
+public class ItemEffectiveCallNumberComponentsTest extends TestBaseWithInventoryUtil {
+  private static final String HOLDINGS_CALL_NUMBER_TYPE = UUID.randomUUID().toString();
+  private static final String HOLDINGS_CALL_NUMBER_TYPE_SECOND = UUID.randomUUID().toString();
+  private static final String ITEM_LEVEL_CALL_NUMBER_TYPE = UUID.randomUUID().toString();
+  private static final String ITEM_LEVEL_CALL_NUMBER_TYPE_SECOND = UUID.randomUUID().toString();
+
+  @BeforeClass
+  public static void createCallNumberTypes() throws Exception {
```
marcjohnson-kint: What happens if these records already exist?

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
```diff
@@ -0,0 +1,294 @@
...
+  private List<Object[]> createPropertiesParams() {
+    List<Object[]> testCases = new ArrayList<>();
+
+    testCases.addAll(createPropertyTestCase("callNumber", "callNumber"));
+    testCases.addAll(createPropertyTestCase("callNumberSuffix", "suffix"));
+    testCases.addAll(createPropertyTestCase("callNumberPrefix", "prefix"));
+    testCases.addAll(createPropertyTestCase(
+      "callNumberTypeId", HOLDINGS_CALL_NUMBER_TYPE,
+      ITEM_LEVEL_CALL_NUMBER_TYPE, "typeId"
+    ));
```
marcjohnson-kint: Why has this line been split sooner than the line length limit?

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
```diff
@@ -10,22 +13,29 @@
...
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+
 import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
 import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
 
 @RunWith(JUnitParamsRunner.class)
 public class ItemEffectiveCallNumberTest extends TestBaseWithInventoryUtil {
```
marcjohnson-kint: What is the difference between this class and the `ItemEffectiveCallNumberComponentsTest`?

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
bohdan-suprun: It is not so clear from the documentation. I've asked @TarasSpashchenko, it seems that it is fine to extend this migration file and change  the `fromModuleVersion`.

Am I correct understand that the inventory-storage module has already been released? And new code, merged to the master, won't affect the release?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Yes, for version >= `18.2.0`  I think.

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
bohdan-suprun: I don't think it is possible, because the record will be created only once, and will have a random ID.

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
bohdan-suprun: The `ItemEffectiveCallNumberTest ` it is test for the migration script, while `ItemEffectiveCallNumberComponentsTest` it is test for the update logic itself.

bohdan-suprun:
> I find the parameterised tests quite hard to follow, so I think it would be useful for someone else more familiar with this style of testing to review this change

@marcjohnson-kint, Do you think I should replace them with the "usual", not parametrized, tests? Because my main goal was to reduce test duplications. We had ~3-4 test cases per each call number property which were pretty-much the same.

marcjohnson-kint:
> > I find the parameterised tests quite hard to follow, so I think it would be useful for someone else more familiar with this style of testing to review this change
> 
> @marcjohnson-kint, Do you think I should replace them with the "usual", not parametrized, tests? Because my main goal was to reduce test duplications. We had ~3-4 test cases per each call number property which were pretty-much the same.

No, I think other folks have used parameterised tests. I personally find them indirect and hard to read, that is probably more my challenge than the tests.

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberTest.java
marcjohnson-kint: Could we rename `ItemEffectiveCallNumberTest` to something that reflect that it is for the migration script, e.g. `ItemEffectiveCallNumberDataUpgradeTest`?

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
marcjohnson-kint: Are there any constraints on the name of the call number type?

My concern is that these records aren't cleaned up between tests.

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
marcjohnson-kint: Yes, mod-inventory-storage has been released.

I think it is potentially confusing to alter existing scripts and change when they are executed.

Thanks for asking for clarification from folks.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -36,9 +39,17 @@ private static EffectiveCallNumberComponents updateComponents(
       updatedCallNumberSuffix = holdings.getCallNumberSuffix();
     }
 
+    String updatedCallNumberTypeId = StringUtils.firstNonBlank(
+      item.getItemLevelCallNumberTypeId(),
+      Optional.ofNullable(holdings).map(HoldingsRecord::getCallNumberTypeId)
+        .orElse(null)
```
mreno-EBSCO: I think we may be abusing `Optional`, I know I am guilty of it as well and I need to stop, but this is better suited for a ternary operator. `Optional` is really meant for return types as described here (from an Oracle blog):
> The purpose of `Optional` is not to replace every single null reference in your codebase but rather to help design better APIs in which—just by reading the signature of a method—users can tell whether to expect an optional value.
```suggestion
        holdings != null ? holdings.getCallNumberTypeId() : null
```

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
bohdan-suprun: https://github.com/folio-org/raml-module-builder/blob/master/DB-schema-migration.md:

It seems that the only case when the script won't be executed is when `current module version` > `fromModuleVersion`.

For example: 
current module version is: 18.2.0.
fromModuleVersion is: 17.1.0.

In this case the script/table creation/etc. won't be executed because the RMB assumes that the change has already been applied before.

But for release example:
current module version (released) is: 17.1.0.
fromModuleVersion is: 18.2.0.

The migration script should be applied.

@TarasSpashchenko, @skoczko, @julianladisch am I correct understand?

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
bohdan-suprun: Yes, we have a UK for the name field. I'm specifying a test-specific name, I don't think that we have the same name. Alternatively, I can re-use existing types, from reference data. Does it sound better?

src/test/java/org/folio/rest/support/builders/HoldingRequestBuilder.java
```diff
@@ -38,7 +40,8 @@ private HoldingRequestBuilder(
     String callNumber,
     String callNumberPrefix,
     String callNumberSuffix,
-    String hrid) {
+    String hrid,
+    String callNumberTypeId) {
```
mreno-EBSCO: This is just aesthetic, but I think we should keep the `callNumber*` components together in the argument list, i.e. swap `hrid` and `callNumberTypeId`.

src/main/resources/templates/db_scripts/populateEffectiveCallNumberComponentsForExistingItems.sql
julianladisch: Yes.

src/main/resources/templates/db_scripts/schema.json
julianladisch: Yes, the populate script will be executed when upgrading from an old version that is `< 18.2.0`. For example when upgrading from 17.0.0 to 19.0.0. But not when upgrading from 18.2.0 to 19.0.0.

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
marcjohnson-kint: I’m not suggesting reusing records from the reference data. I’d prefer the tests didn’t do that because I think they should be independent of it.

My concern is that these records don’t appear to be removed. Are they removed elsewhere in a general cleanup?

bohdan-suprun: I don't think, I can remove it in `@AfterClass` method.

marcjohnson-kint: I typically suggest cleaning up records before tests run rather than after, in order to aid diagnostics if tests fail.

ramls/item.json
```diff
@@ -69,6 +69,12 @@
           "type": "string",
           "description": "Effective Call Number Suffix is the suffix of the identifier assigned to an item or its holding and associated with the item.",
           "readonly": true
+        },
+        "typeId": {
+          "type": "string",
+          "description": "Effective Call Number Type Id is the call number type id assigned to the item or associated holding.",
```
julianladisch: Can the description be clear which id is chosen if both item and holding have such an id?
For example "is the call number type id of the item, if available, otherwise that of the holding."

julianladisch:
The tests in ItemEffectiveCallNumberComponentsTest.java are difficult to read because one cannot see which values gets passed into the test methods.
This is the same set of test values but visible from the source code: https://github.com/folio-org/mod-inventory-storage/blob/0b879ea9720bc86ba2222450b3058edd3cbfdd92/src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java

bohdan-suprun:
> The tests in ItemEffectiveCallNumberComponentsTest.java are difficult to read because one cannot see which values gets passed into the test methods.
> This is the same set of test values but visible from the source code: https://github.com/folio-org/mod-inventory-storage/blob/0b879ea9720bc86ba2222450b3058edd3cbfdd92/src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java

Yo're right, but another side is that params string becomes too long for update case, and it will reduce readability.

Let me try to refactor them.

bohdan-suprun:
@julianladisch, I've refactored tests, they should be more readable now, please take a look.

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
```diff
@@ -42,12 +44,12 @@
...
-  protected static final String ANNEX_LIBRARY_LOCATION = "Annex Library";
-  protected static final String ONLINE_LOCATION = "Online";
-  protected static final String THIRD_FLOOR_LOCATION = "Third Floor";
-  protected static final String FOURTH_FLOOR_LOCATION = "Fourth Floor";
+  public static final String MAIN_LIBRARY_LOCATION = "Main Library";
+  public static final String SECOND_FLOOR_LOCATION = "Second Floor";
+  public static final String ANNEX_LIBRARY_LOCATION = "Annex Library";
+  public static final String ONLINE_LOCATION = "Online";
+  public static final String THIRD_FLOOR_LOCATION = "Third Floor";
+  public static final String FOURTH_FLOOR_LOCATION = "Fourth Floor";
```
marcjohnson-kint: Why have locations needed to be made public for a call number related test?

src/test/java/org/folio/rest/support/http/ResourceClient.java
```diff
@@ -160,22 +170,26 @@ public Response getById(UUID id)
...
+  public void delete(UUID id) throws MalformedURLException, InterruptedException,
+    ExecutionException, TimeoutException {
+
+    Response response = deleteIfPresent(id != null ? id.toString() : null);
 
     assertThat(String.format(
       "Failed to delete %s %s: %s", resourceName, id, response.getBody()),
-      response.getStatusCode(), is(204));
+      response.getStatusCode(),
+      is(204));
```
marcjohnson-kint: Why has this line been split?

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
bohdan-suprun: I've moved data provider classes to a separate package, and `src/test/java/org/folio/rest/api/testdata/ItemEffectiveLocationTestDataProvider.java` uses this properties.

src/test/java/org/folio/rest/support/http/ResourceClient.java
bohdan-suprun: For better readability. Now each param is on a separate line, i guess it is more readable now.

marcjohnson-kint: Why does having the property and the matcher on separate lines make this more readable?

bohdan-suprun: at least you can easily see which parameters are passed to the method. Previously it was like the getStatusCode() is a parameter of String.format(...).

marcjohnson-kint: I really don't follow. 

The only changes reported here show splitting the `response.getStatusCode()` and `is(204)` onto separate lines, and I don't understand why that is more readable. 

What am I missing?

bohdan-suprun: I've reverted this change.

marcjohnson-kint:
@julianladisch are you content with the changes that @bohdan-suprun has made to the tests?

marcjohnson-kint:
Thanks @julianladisch

PR #375, opened 2019-12-10
MODINVSTOR-405 weaken HRID error message checks

marcjohnson-kint:
The error messages produced by PostgreSQL vary by version, meaning that the build could fail using a PostgreSQL server to the one used by embedded PostgreSQL.

This even occurs within the same major version (the line number reference differs).

This pull request weakens the check on the error message, to not include the line number or the maximum value of the sequence.

### DIFF
--- a/src/test/java/org/folio/rest/support/matchers/PostgresErrorMessageMatchers.java
+++ b/src/test/java/org/folio/rest/support/matchers/PostgresErrorMessageMatchers.java
@@ -1,14 +1,12 @@
 package org.folio.rest.support.matchers;
 
 import org.hamcrest.Matcher;
-import org.jetbrains.annotations.NotNull;
 
 import static org.hamcrest.CoreMatchers.containsString;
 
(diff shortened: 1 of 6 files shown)


src/test/java/org/folio/rest/support/matchers/PostgresErrorMessageMatchers.java
```diff
@@ -0,0 +1,18 @@
+package org.folio.rest.support.matchers;
+
+import org.hamcrest.Matcher;
+import org.jetbrains.annotations.NotNull;
```
bohdan-suprun: Is it fine to use this annotation? I think guys who uses another IDE may have some issues with this.

marcjohnson-kint: That is a good question. I don't think it is IDE specific, I'm happy to remove it.

PR #376, opened 2019-12-12
MODINVSTOR-410 increase number of digits for HRID

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-410: Increase digits for Inventory Instance HRIDs
https://issues.folio.org/browse/MODINVSTOR-411: Increase digits for Inventory Holdings HRIDs
https://issues.folio.org/browse/MODINVSTOR-412: Increase digits for Inventory Item HRIDs
https://issues.folio.org/browse/MODINVSTOR-406: Make the HRID sequences consistent

* Add alter holdings/instance/item HRID sequences script.
* Expand HRID to be 11-digits instead of 8-digits.
* Add unit test to verify that current value of the sequence has not been changed.
* Modify existing unit tests.

**Open question**: Do we need to migrate existing issued HRIDs to be 11-digits?

### DIFF
--- a/src/main/java/org/folio/rest/support/HridManager.java
+++ b/src/main/java/org/folio/rest/support/HridManager.java
@@ -152,10 +152,9 @@ public class HridManager {
     final Promise<ResultSet> promise = Promise.promise();
 
     // Only update the sequence if the start number has changed
-    if (existingHridSetting.getStartNumber().intValue() !=
-        hridSetting.getStartNumber().intValue()) {
+    if (!Objects.equals(existingHridSetting.getStartNumber(), hridSetting.getStartNumber())) {
       final String sql = String.format("select setval('hrid_%s_seq',%d,FALSE)",
-          field, hridSetting.getStartNumber().intValue());
+          field, hridSetting.getStartNumber());
       try {
         postgresClient.select(conn, sql, promise);
       } catch (Exception e) {
@@ -193,7 +192,7 @@ public class HridManager {
     final String hridPrefix = hridSetting.getPrefix();
 
     return promise.future()
-        .map(sequence -> String.format("%s%08d", Objects.toString(hridPrefix, ""),
+        .map(sequence -> String.format("%s%011d", Objects.toString(hridPrefix, ""),
             sequence.getLong(0)));
   }
 
--- a/src/main/resources/templates/db_scripts/alterHridSequences.sql
+++ b/src/main/resources/templates/db_scripts/alterHridSequences.sql
@@ -1,10 +1,13 @@
 -- Increase the initial size of the HRID sequences
 
 ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_instances_seq
-  AS BIGINT MAXVALUE 99999999999;
+  AS BIGINT MAXVALUE 99999999999
+  OWNED BY ${myuniversity}_${mymodule}.hrid_settings.jsonb;
 
 ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_holdings_seq
-  AS BIGINT MAXVALUE 99999999999;
+  AS BIGINT MAXVALUE 99999999999
+  OWNED BY ${myuniversity}_${mymodule}.hrid_settings.jsonb;
 
 ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_items_seq
-  AS BIGINT MAXVALUE 99999999999;
+  AS BIGINT MAXVALUE 99999999999
+  OWNED BY ${myuniversity}_${mymodule}.hrid_settings.jsonb;
--- a/src/main/resources/templates/db_scripts/alterHridSequences.sql
+++ b/src/main/resources/templates/db_scripts/alterHridSequences.sql
@@ -1,0 +1,10 @@
+-- Increase the initial size of the HRID sequences
+
+ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_instances_seq
+  AS BIGINT MAXVALUE 99999999999;
+
+ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_holdings_seq
+  AS BIGINT MAXVALUE 99999999999;
+
+ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_items_seq
+  AS BIGINT MAXVALUE 99999999999;
--- a/src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
+++ b/src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
@@ -46,13 +46,13 @@ public class HridSettingsIncreaseMaxValueMigrationTest extends TestBase {
     executeSql(dropSequenceSql(HOLDINGS_SEQ));
     executeSql(dropSequenceSql(ITEMS_SEQ));
 
-    executeSqlFile(CREATE_SEQUENCE_SQL);
+    executeMultipleSqlStatements(CREATE_SEQUENCE_SQL);
   }
 
   private void alterSequences() throws InterruptedException, ExecutionException,
     TimeoutException {
 
-    executeSqlFile(ALTER_SEQUENCE_SQL);
+    executeMultipleSqlStatements(ALTER_SEQUENCE_SQL);
   }
 
   private Long incrementSequence(String sequenceName)
--- a/src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
+++ b/src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
@@ -2,22 +2,21 @@ package org.folio.rest.api;
 
 import static org.folio.rest.api.StorageTestSuite.TENANT_ID;
 import static org.folio.rest.api.StorageTestSuite.getVertx;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
 
 import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.folio.rest.persist.PostgresClient;
 import org.folio.util.ResourceUtil;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 
-import io.vertx.core.CompositeFuture;
-import io.vertx.core.Future;
-import io.vertx.core.Promise;
 import io.vertx.core.json.JsonArray;
-import io.vertx.ext.unit.TestContext;
-import io.vertx.ext.unit.junit.VertxUnitRunner;
 
-@RunWith(VertxUnitRunner.class)
 public class HridSettingsIncreaseMaxValueMigrationTest extends TestBase {
   private static final String CREATE_SEQUENCE_SQL = loadCreateSequenceSqlFile();
   private static final String ALTER_SEQUENCE_SQL = loadAlterSequenceSqlFile();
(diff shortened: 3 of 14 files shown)


src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import io.vertx.core.CompositeFuture;
+import io.vertx.core.Future;
+import io.vertx.core.Promise;
+import io.vertx.core.json.JsonArray;
+import io.vertx.ext.unit.TestContext;
+import io.vertx.ext.unit.junit.VertxUnitRunner;
+
+@RunWith(VertxUnitRunner.class)
+public class HridSettingsIncreaseMaxValueMigrationTest extends TestBase {
```
marcjohnson-kint: These tests are a very different style to the other tests. Please can you help me understand why this approach was chosen?

In particular it appears to be binding closer to vert.x futures, which are getting a significant change in vert.x 4.0

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -0,0 +1,139 @@
...
+  private static final String CREATE_SEQUENCE_SQL = loadCreateSequenceSqlFile();
+  private static final String ALTER_SEQUENCE_SQL = loadAlterSequenceSqlFile();
+  private static final String INSTANCES_SEQ = "hrid_instances_seq";
+  private static final String HOLDINGS_SEQ = "hrid_holdings_seq";
+  private static final String ITEMS_SEQ = "hrid_items_seq";
+
+  @Test
+  public void canSaveCurrentValueOfSequences(TestContext testContext) {
```
marcjohnson-kint:
```suggestion
  public void retainsCurrentValueOfSequences(TestContext testContext) {
```

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -0,0 +1,139 @@
...
+      .compose(notUsed -> incrementSequence(HOLDINGS_SEQ))
+      .map(hrid -> testContext.assertEquals(8L, hrid))
+      .compose(notUsed -> incrementSequence(ITEMS_SEQ))
+      .map(hrid -> testContext.assertEquals(1_000_000_00L, hrid))
+      // Handle test result
+      .setHandler(testContext.asyncAssertSuccess());
+  }
+
+  @Test
+  public void canSetBigIntValue(TestContext testContext) {
```
marcjohnson-kint: Why is this test needed? Do the regular tests not try to set a sequence value higher than the previous limit (and bigger than an integer)?

I think it would be preferable to have regular tests that successfully get a large (nearly maximum) value for the HRID.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
bohdan-suprun: I chose the async style because there are a lot of DB operations that have to be executed to create/destroy/get next value/set value, which are async.

If you think that this approach may cause compatibility issues, then I think I can convert it to the blocking style.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
bohdan-suprun: Yes, we have such test `canGetNextHridWhenStartNumberIsLong`. This test to check it on DB level.

marcjohnson-kint: I personally don't think we need a separate test for it here as well.

If we could express it without integration, then I might understand the value of it, however this still requires external integration so I'm not sure it is worth it.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
marcjohnson-kint: Ok, I personally find these kinds of tests, with the vert.x context and chaining to not be very expressive. And I reckon they will need changing soon.

I suggest we leave this one as is, as an isolated case. Maybe we could add a comment to describe the reasoning for this style.

To hide the asynchronous pain, I've typically extracted methods that block.

bohdan-suprun: I've refactored it to be blocking.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
bohdan-suprun: Ok, removed it, because we have the integration test.

src/main/resources/templates/db_scripts/alterHridSequences.sql
```diff
@@ -0,0 +1,10 @@
+-- Increase the initial size of the HRID sequences
+
+ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_instances_seq
+  AS BIGINT MAXVALUE 99999999999;
+
+ALTER SEQUENCE IF EXISTS ${myuniversity}_${mymodule}.hrid_holdings_seq
+  AS BIGINT MAXVALUE 99999999999;
```
mreno-EBSCO: Since we are updating the sequences here, does it make sense to do [MODINVSTOR-406](https://issues.folio.org/browse/MODINVSTOR-406) as well?

bohdan-suprun: Make sense to me. 
>  We should determine if the sequences should be owned by this table for ease of clean up and either add the owned by to the holdings/items sequence creation or remove it from the instances sequence creation.

Do you have any concerns that it is not needed here?

mreno-EBSCO: I am not sure what is the best practice for sequences. I assume that binding them to the HRID settings table is the right thing to do, but I am not a DBA, so I could be wrong. The previous HRID sequences were bound to the actual jsonb column of the inventory type they generated HRIDs for, since there were no HRID settings. I suspect we want to add the same owner as the instances sequence to the holdings and items sequences. I can't imagine we'd want to drop the owner from the instances sequence. But that is just my opinion. 😄

bohdan-suprun: Ok, I've add owner for the sequences.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+        } else {
+          nextValue.complete(results.get(0).getLong(0));
+        }
+      }
+    });
+
+    return nextValue.get(5, TimeUnit.SECONDS);
+  }
+
+  private Long incrementSequence(String sequenceName, int times)
```
marcjohnson-kint: Is this function used?

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -78,4 +79,40 @@ void assertGetNotFound(URL url) {
       throw new RuntimeException(e);
     }
   }
+
+  void executeSqlFile(String fileContent)
```
marcjohnson-kint: What is the difference between SQL that comes from a file and elsewhere? At this point they are both represented as strings?

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
bohdan-suprun: yes, see HridSettingsIncreaseMaxValueMigrationTest lines # 31-32

src/test/java/org/folio/rest/api/TestBase.java
bohdan-suprun: Yes, I use different RMB function to execute file, it splits the file content to lines, and executes it line by line.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
marcjohnson-kint: Ok

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: Is this really about it coming from a file, or that it includes multiple SQL statements?

marcjohnson-kint:
@mreno-EBSCO Do you have any more feedback for this change?

src/test/java/org/folio/rest/api/TestBase.java
bohdan-suprun: It coming from a file, for example migration scripts (alterHridSequences.sql file or hridSettings.sql)

marcjohnson-kint: I understand that it is coming from a file. I'm trying to understand if the special processing is because it's come from a file or because it is multiple statements?

Neither of these methods can know that it has come from a file as they take String parameters.

bohdan-suprun: I think because it is multiple statements.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -0,0 +1,120 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.folio.rest.persist.PostgresClient;
+import org.folio.util.ResourceUtil;
```
mreno-EBSCO: We should create a direct `test` dependency to `util` in the pom file.

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -152,10 +152,9 @@ public HridManager(Context context, PostgresClient postgresClient) {
     final Promise<ResultSet> promise = Promise.promise();
 
     // Only update the sequence if the start number has changed
-    if (existingHridSetting.getStartNumber().intValue() !=
-        hridSetting.getStartNumber().intValue()) {
+    if (!Objects.equals(existingHridSetting.getStartNumber(), hridSetting.getStartNumber())) {
       final String sql = String.format("select setval('hrid_%s_seq',%d,FALSE)",
-          field, hridSetting.getStartNumber().intValue());
+          field, hridSetting.getStartNumber());
```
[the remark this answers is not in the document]
bohdan-suprun: the `startNumber` is a long now. The same for above.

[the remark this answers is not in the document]
mreno-EBSCO: Duh! Of course. 😄 That's one of the many reasons I dislike autoboxing, it hides the details.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
bohdan-suprun: I've added the dependency, but the scope have to be `compile` because the utilities is used by maven plugins and RMB itself.

mreno-EBSCO: Yeah, the dependencies in this project, as well as most FOLIO projects, are not managed well. The `GenerateRunner` main class we call should really be a fat jar so we could include a direct runtime dependency on `domain-models-interface-extensions`. Since it is not, as you've mentioned, we can't limit the scope of `utils` to test. Interestingly, if we do add a runtime dependency on `domain-models-interface-extensions`, `ShelfLocation` breaks. It is using a dependency on the legacy `commons-lang` that is available via `domain-models-interface-extensions`, but no where else, since we want to use `commons-lang3`. There are probably other problems with the dependencies that a maven plugin would expose (just tried and there are many missing dependencies and one unused one). I guess we should cross that tech-debt bridge at a later date.

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: Ok, then I think we should reflect that in the method name (it's a shame the RAML Module Builder method name cannot be so easily changed).

It's not major, so shouldn't block the merging of this change.

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
marcjohnson-kint: Yeah, alas it really is. I didn't know there was a maven plugin for detecting dependency issues until a few years into the project

src/test/java/org/folio/rest/api/TestBase.java
bohdan-suprun: Does `executeMultipleSqlStatements` work well?

marcjohnson-kint: I'm good with that

PR #377, opened 2019-12-13
MODINVSTOR-409: effectiveLocationId foreign key upgrade script

julianladisch:
RMB creates the UUID column `effectiveLocationId` with NULL as inital value
for all existing item records.

This is wrong.

This upgrade script populates the UUID column with the value from the jsonb column.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -697,7 +697,7 @@
     {
       "run": "after",
       "snippetPath": "populateEffectiveLocationForeignKey.sql",
-      "fromModuleVersion": "18.1.1"
+      "fromModuleVersion": "18.2.0"
     },
     {
       "run": "after",
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -696,6 +696,11 @@
     },
     {
       "run": "after",
+      "snippetPath": "populateEffectiveLocationForeignKey.sql",
+      "fromModuleVersion": "18.1.1"
+    },
+    {
+      "run": "after",
       "snippetPath": "populateEffectiveCallNumberComponentsForExistingItems.sql",
       "fromModuleVersion": "17.1.0"
     },
--- a/src/main/resources/templates/db_scripts/populateEffectiveLocationForeignKey.sql
+++ b/src/main/resources/templates/db_scripts/populateEffectiveLocationForeignKey.sql
@@ -1,0 +1,1 @@
+UPDATE ${myuniversity}_${mymodule}.item SET effectiveLocationId = (jsonb->>'effectiveLocationId')::uuid;


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -694,6 +694,11 @@
       "snippetPath": "populateEffectiveLocationForExistingItems.sql",
       "fromModuleVersion": "17.1.0"
     },
+    {
+      "run": "after",
+      "snippetPath": "populateEffectiveLocationForeignKey.sql",
+      "fromModuleVersion": "18.1.1"
```
marcjohnson-kint: This will still run if the next released version is 18.2.0?

julianladisch: Yes.
When upgrading from `previousVersion` to `newVersion` it runs if `previousVersion` < `fromModuleVersion`: https://github.com/folio-org/raml-module-builder/blob/v29.1.2/dbschema/src/main/java/org/folio/rest/persist/ddlgen/Versioned.java#L36

marcjohnson-kint: Ok, thanks, I think this `fromModuleVersion` should reflect the version next being released where possible, which I believe is 18.2.0?

julianladisch: `fromModuleVersion` should be the next released version.
I don't know what the next version will be. These issues still have 18.1.1 as Fix Version:
https://issues.folio.org/browse/MODINVSTOR-407
https://issues.folio.org/browse/MODINVSTOR-409

marcjohnson-kint: Those issues might well have incorrect fix versions (that is a fairly common occurrence in JIRA, along with missing fix versions). I will investigate.

[Master](https://github.com/folio-org/mod-inventory-storage/blob/719b914f4cf8c508f7e761463df7520f747d9105/pom.xml#L5) currently states that the next mainline release of mod-inventory-storage is intended to 18.2.0.

I believe this is inline with the release guidelines, that the next release is mean to be the next appropriate minor version.

If this change was to become a back ported bug fix, then I imagine it could also be part of 18.1.1 as well.

(this is one of the challenges of coupling database changes to module versions)

marcjohnson-kint: @julianladisch It appears that MODINVSTOR-409 was allocated to 18.1.1 by yourself.

Was this because it was the first fix version available in JIRA?

julianladisch: This was because I consider it a bugfix for 18.1.0.

marcjohnson-kint: Bug fixes for 18.1.0 need approval by the release triage group. Please ask that group to consider this as a 2019 Q4 bug fix issue. 

I think bug fixes should be applied to master and tested prior to back porting.

The bug fix releases themselves are not conducted on master, rather a long lived release branch.

If I have misinterpreted the release documentation, please let me know.

julianladisch: Do you want `"fromModuleVersion": "18.2.0"` on master and `"fromModuleVersion": "18.1.1"` on the 18.1.x release branch?

marcjohnson-kint: I think the `fromModuleVersion` in master should be `18.2.0`. 

I think the focus should be on resolving the issues in master. 

We have requested an exception from the Q4 bug fix release process for the HRID work. If this work (and others hoped for Q4) is merged and tested prior to that release, then we don't need a special bug fix release.

PR #381, opened 2019-12-19
MODINVSTOR-405: Weaken duplicate HRID error message check

julianladisch:


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -2255,12 +2255,8 @@ public void cannotCreateACollectionOfInstancesWithDuplicatedHRIDs() throws Excep
...
 
     assertThat(ibr.getErrorMessages(), notNullValue());
-    assertThat(ibr.getErrorMessages().get(0), is(
-      "ErrorMessage(fields=[(Severity, ERROR), (V, ERROR), (SQLSTATE, 23505), " +
-        "(Message, duplicate key value violates unique constraint \"instance_hrid_idx_unique\"), " +
-        "(Detail, Key (lower(f_unaccent(jsonb ->> 'hrid'::text)))=(in00000000001) already exists.), " +
-        "(s, test_tenant_mod_inventory_storage), (t, instance), (n, instance_hrid_idx_unique), " +
-        "(File, nbtinsert.c), (Line, 434), (Routine, _bt_check_unique)])"));
+    assertThat(ibr.getErrorMessages().get(0), allOf(
+      containsString("(SQLSTATE, 23505)"), containsString("instance_hrid_idx_unique")));
```
marcjohnson-kint: I think it would be useful if this also contained reference to the reason for the failure beyond just the index name. Unless that message differs significantly by PostgreSQL versions?

julianladisch: Yes, the message may change by PostgreSQL version, see https://github.com/folio-org/raml-module-builder/blob/v29.1.4/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/ValidationHelper.java#L146-L147 for an example for such a change.
The SQLSTATE error code is stable.

marcjohnson-kint: @julianladisch Ok, thanks, I'm happy with the SQL state approach instead. Could this be extracted to a matcher that could be used with any unique constraint check?

PR #382, opened 2019-12-19
[BREAKING] MODINVSTOR-416, MODINVSTOR-392, MODINVSTOR-332: item storage 8.0

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-332: Item: Change array copyNumbers to simple property copyNumber (breaking)
https://issues.folio.org/browse/MODINVSTOR-283: Restrict item states
https://issues.folio.org/browse/MODINVSTOR-392: Make item.status.date read only field.
https://issues.folio.org/browse/MODINVSTOR-425: Make item.status required.

Item-storage 8.0 breaking changes.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -727,7 +727,7 @@
     {
       "run": "after",
       "snippetPath": "migrateItemCopyNumberToSingleValue.sql",
-      "fromModuleVersion": "18.2.0"
+      "fromModuleVersion": "19.0.0"
     }
   ]
 }
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -723,6 +723,11 @@
       "run": "after",
       "snippetPath": "alterHridSequences.sql",
       "fromModuleVersion": "18.2.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "migrateItemCopyNumberToSingleValue.sql",
+      "fromModuleVersion": "18.2.0"
     }
   ]
 }
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -16,7 +16,6 @@ AS $$
     ELSE
       NEW.jsonb = NEW.jsonb #- '{status, date}';
 	  END IF;
-
 	  RETURN NEW;
   END;
   $$ LANGUAGE 'plpgsql';
--- a/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
+++ b/src/main/resources/templates/db_scripts/updateItemStatusDate.sql
@@ -11,8 +11,13 @@ AS $$
 	    -- Date time in "YYYY-MM-DD"T"HH24:MI:SS.ms'Z'" format at UTC (00:00) time zone
       NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}',
        to_jsonb(to_char(CURRENT_TIMESTAMP(3) AT TIME ZONE 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.ms"Z"')), true);
+    ELSIF (OLD.jsonb->'status'->'date' IS NOT NULL) THEN
+      NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', OLD.jsonb->'status'->'date', true);
+    ELSE
+      NEW.jsonb = NEW.jsonb #- '{status, date}';
 	  END IF;
-	RETURN NEW;
+
+	  RETURN NEW;
   END;
   $$ LANGUAGE 'plpgsql';
 
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -198,6 +198,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
 
     JsonObject itemToCreate = new JsonObject()
       .put("id", id.toString())
+      .put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()))
       .put("holdingsRecordId", holdingsRecordId.toString())
       .put("materialTypeId", journalMaterialTypeID)
       .put("permanentLoanTypeId", canCirculateLoanTypeID)
@@ -231,6 +232,27 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
 
     assertThat(tags.size(), is(1));
     assertThat(tags, hasItem(TAG_VALUE));
+  }
+
+  @Test
+  public void canReplaceItemWithNewProperties() throws Exception {
+    final UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
+    final UUID id = UUID.randomUUID();
+    final String expectedCopyNumber = "copy1";
+
+    JsonObject itemToCreate = smallAngryPlanet(id, holdingsRecordId);
+    createItem(itemToCreate);
+
+    JsonObject createdItem = getById(id).getJson();
+    assertThat(createdItem.getString("copyNumber"), nullValue());
+
+    JsonObject updatedItem = createdItem.copy()
+      .put("copyNumber", expectedCopyNumber);
+
+    itemsClient.replace(id, updatedItem);
+
+    JsonObject updatedItemResponse = itemsClient.getById(id).getJson();
+    assertThat(updatedItemResponse.getString("copyNumber"), is(expectedCopyNumber));
   }
 
   @Test
@@ -403,6 +425,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
 
     final JsonObject itemToCreate = new JsonObject()
       .put("id", id.toString())
+      .put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()))
       .put("holdingsRecordId", holdingsRecordId.toString())
       .put("materialTypeId", journalMaterialTypeID)
       .put("permanentLoanTypeId", canCirculateLoanTypeID)
@@ -437,6 +460,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     JsonObject itemToCreate = new JsonObject();
     String itemId = UUID.randomUUID().toString();
     itemToCreate.put("id", itemId);
+    itemToCreate.put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()));
     itemToCreate.put("holdingsRecordId", holdingsRecordId.toString());
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("materialTypeId", bookMaterialTypeID);
@@ -466,6 +490,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     JsonObject itemToCreate = new JsonObject();
     String itemId = UUID.randomUUID().toString();
     itemToCreate.put("id", itemId);
+    itemToCreate.put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()));
     itemToCreate.put("holdingsRecordId", holdingsRecordId.toString());
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("materialTypeId", bookMaterialTypeID);
@@ -604,6 +629,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     JsonObject itemToCreate = new JsonObject();
     String itemId = UUID.randomUUID().toString();
     itemToCreate.put("id", itemId);
+    itemToCreate.put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()));
     itemToCreate.put("holdingsRecordId", holdingsRecordId.toString());
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("materialTypeId", bookMaterialTypeID);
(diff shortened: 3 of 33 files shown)


mreno-EBSCO:
Congratulations, you also fixed: [MODINVSTOR-380](https://issues.folio.org/browse/MODINVSTOR-380)

Can you assign yourself and update MODINVSTOR-380?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -11,8 +11,13 @@ AS $$
 	    -- Date time in "YYYY-MM-DD"T"HH24:MI:SS.ms'Z'" format at UTC (00:00) time zone
       NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}',
        to_jsonb(to_char(CURRENT_TIMESTAMP(3) AT TIME ZONE 'UTC', 'YYYY-MM-DD"T"HH24:MI:SS.ms"Z"')), true);
+    ELSIF (OLD.jsonb->'status'->'date' IS NOT NULL) THEN
+      NEW.jsonb = jsonb_set(NEW.jsonb, '{status,date}', OLD.jsonb->'status'->'date', true);
+    ELSE
```
VRohach: @bohdan-suprun Could you please describe - why did you add this line:  `NEW.jsonb = NEW.jsonb` ?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
```diff
@@ -11,8 +11,13 @@ AS $$
...
+      NEW.jsonb = NEW.jsonb #- '{status, date}';
 	  END IF;
-	RETURN NEW;
+
```
VRohach: I think it will be better to remove this extra line. What do you think?

src/main/resources/templates/db_scripts/updateItemStatusDate.sql
bohdan-suprun: Every `readonly` field is being removed by RMB when receiving a request. This prevents a readonly field to be changed. But another side of this behavior that it is being removed regardless whether it has changed or not.

Because of this we have to restore the value of the field every time we do an update.

This code sets the `NEW.jsonb->'status'->'date'` to `OLD.jsonb->'status'->'date'` if previous date is not null. And the `ELSE` block it is additional logic required to temporary fix https://issues.folio.org/browse/RMB-538.

VRohach: Oh, I catch your point. Thanks!

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -723,6 +723,11 @@
       "run": "after",
       "snippetPath": "alterHridSequences.sql",
       "fromModuleVersion": "18.2.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "migrateItemCopyNumberToSingleValue.sql",
+      "fromModuleVersion": "18.2.0"
```
marcjohnson-kint: The next release is scheduled to be 18.3.0, should this be 18.3.0?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1633,17 +1661,166 @@ public void checkIfStatusDateChangesWhenItemStatusUpdated()
...
+      is("newItemLevelCallNumber"));
+
+    String oldStatusDate = updatedStatus.getString("date");
+    String newStatusDate = itemWithUpdatedCallNumberResponse.getJsonObject("status").getString("date");
+
+    assertThat(oldStatusDate, is(newStatusDate));
+  }
+
+  @Test
+  public void statusUpdatedDateIsNullOnSubsequentUpdates() throws Exception {
```
marcjohnson-kint: When should status updated date be null?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1633,17 +1661,166 @@ public void checkIfStatusDateChangesWhenItemStatusUpdated()
...
+    assertThat(secondUpdatedItemResponse.getString("temporaryLocationId"),
+      is(onlineLocationId.toString()));
+    assertThat(secondUpdatedItemResponse.getJsonObject("status").getString("name"),
+      is("Available"));
+    assertThat(secondUpdatedItemResponse.getJsonObject("status").getString("date"),
+      nullValue());
+  }
+
+  @Test
+  public void canUpdateItemWithCopyNumber() throws Exception {
```
marcjohnson-kint: Does this need to be a specific test, or can it be included in the general updating of an item?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2131,8 +2308,49 @@ public void testItemHasLastCheckInProperties()
...
+
+    Response postResponse = createCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(postResponse.getStatusCode(), is(400));
+    assertThat(postResponse.getBody(),
+      matchesPattern("(?s)Json content error Cannot construct instance of `.+`, problem: Wrong status name.+")
+    );
+  }
+
+  @Test
+  public void canRemoveItemStatus() throws Exception {
```
marcjohnson-kint: Why are we allowing an item status to be removed?

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Actually, this change introduces a breaking compatibility change, which I think means the implementation version should change to 19.0.0

bohdan-suprun: Ok, make sense. Should I change the implementation version as well?

src/test/java/org/folio/rest/api/ItemStorageTest.java
bohdan-suprun: Yes, correct.

marcjohnson-kint: Ok, thanks

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Yes, please change the implementation version to 19.0.0

src/test/java/org/folio/rest/api/ItemStorageTest.java
bohdan-suprun: Well, the status is not a required property, so I assumed it is possible case.

marcjohnson-kint: I'm wondering if it might be worth making it required as part of this change. 

It shouldn't be removable, it is only marked not required because it is defaulted during initial creation.

src/test/java/org/folio/rest/api/ItemStorageTest.java
bohdan-suprun: Seems we don't have a separate test case for updating an item, only for creating.

src/test/java/org/folio/rest/api/ItemStorageTest.java
bohdan-suprun: Do we need to declare it in schema or do it in code?

We won't be allowed to do the defaulting through JsonSchema if the entire `status` field  is not present, it will be possible for `name` field only.

For example:
1.
```
{
  "id": "id",
  "hrid": "hrid"
  ... <no-status>
}
```
Won't set a default value and will throw validation error.

2. 
```
{
  "id": "id",
  "hrid": "hrid",
  "status": {}
}
```
This case will work, we will be able to set a default value for name.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: Hmm, ok, I wonder if this test could be changed to a more general can replace item with new properties test?

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: The default is done in code. (I don't find defaults in schema particularly useful).

I think the defaulting should be removed as part of this change and instead the status be required to be set by the client.

We have had issues with defaulting values, and don't really want to do it unless absolutely necessary. This was likely done trying to be helpful or avoid a breaking change, I don't think it adds much value at this stage.

bohdan-suprun: Ok, I've removed the defaulting logic and declared the status as required field.

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
```diff
@@ -217,7 +218,7 @@ protected static Item buildItem(UUID holdingsRecordId,
     itemToCreate.put("id", UUID.randomUUID().toString());
     itemToCreate.put("holdingsRecordId", holdingsRecordId.toString());
     itemToCreate.put("barcode", Long.toString(new Random().nextLong()));
-    itemToCreate.put("status", new JsonObject().put("name", "Available"));
+    itemToCreate.put("status", new JsonObject().put("name", Status.Name.AVAILABLE.value()));
```
marcjohnson-kint: Is that a generated constant from the production code?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1559,11 +1629,9 @@ public void checkIfStatusDateExistsWhenItemStatusUpdated()
     assertThat(item.getId(), is(id.toString()));
 
     assertThat(item.getStatus().getName(),
-      is("Checked out"));
-
-    assertThat(item.getStatus().getDate(), withinSecondsBeforeNow(seconds(2)));
+      is(Status.Name.CHECKED_OUT));
 
-    assertThat(item.getStatus().getDate(), hasIsoFormat());
+    assertThat(item.getStatus().getDate().toInstant(), withinSecondsBeforeNow(seconds(2)));
```
marcjohnson-kint: Thank you for making this check stricter

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
bohdan-suprun: Yes, it is enum value generated by RMB.

marcjohnson-kint: I think that API tests should be minimally coupled to the details of the production code. Both in order to reduce the chance of false diagnostics (e.g. passing or failing for the wrong reason) and to improve maintainability by making the less brittle to internal design changes.

I can understand the desire to use a constant value. I think that constant should be defined in the test context not taken from the production code.

What do you think?

bohdan-suprun: I agree, there might be a place for false passing. Changed

marcjohnson-kint: Please can the mod-inventory pull request be updated similarly, if you haven't already done so

src/test/java/org/folio/rest/api/MaterialTypeTest.java
```diff
@@ -7,6 +7,8 @@
 import io.vertx.core.http.HttpClientResponse;
 import io.vertx.core.http.HttpMethod;
 import io.vertx.core.json.JsonObject;
+
+import org.folio.rest.jaxrs.model.Status;
```
marcjohnson-kint: What is this import used for?

bohdan-suprun: Thanks, good catch. I wonder, why sonar did not notify about it...

src/test/java/org/folio/rest/support/matchers/ResponseMatcher.java
```diff
@@ -0,0 +1,58 @@
...
+    return new TypeSafeMatcher<Response>() {
+      @Override
+      protected boolean matchesSafely(Response response) {
+        if (response.getStatusCode() != 422) {
+          return false;
+        }
+
+        try {
+          JsonArray errors = response.getJson().getJsonArray("errors");
+          if (errors != null && errors.size() == 1) {
```
OleksiiKuzminov: Can it be mapped to Error object to avoid getting properties by name?

src/test/java/org/folio/rest/support/matchers/ResponseMatcher.java
```diff
@@ -0,0 +1,58 @@
...
+              String message = error.getString("message");
+              String key = parameters.getJsonObject(0).getString("key");
+              String value = parameters.getJsonObject(0).getString("value");
+
+              return Objects.equals(expectedMessage, message)
+                && Objects.equals(expectedKey, key)
+                && Objects.equals(expectedValue, value);
+            }
+          }
+
```
OleksiiKuzminov: formatting

src/test/java/org/folio/rest/support/matchers/ResponseMatcher.java
```diff
@@ -0,0 +1,58 @@
...
+          return false;
+        } catch (DecodeException ex) {
+          return false;
```
OleksiiKuzminov: Should it be logged?

bohdan-suprun: Personally, I don't like to log anything in test classes. If class fails with the DecodeException then JUnit will show actual vs expected error, e.g:

```
java.lang.AssertionError: 
Expected: Response has 422 status and 'message' - "error", 'key' - "key" and 'value' - "value"
     but: was <[statusCode=422, contentType=application/json, body={"foo":"bar"]>
```

Does it work for you?

src/test/java/org/folio/rest/support/matchers/ResponseMatcher.java
bohdan-suprun: Fixed.

PR #394, opened 2020-01-14
MODINVSTOR-415: Document execution order of effective location triggers

julianladisch:


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1455,6 +1455,9 @@ public void canSearchItemByEffectiveLocation() throws Exception {
       assertTrue(createdItem.getJson().containsKey("effectiveLocationId"));
     }
 
+    // This checks that the two triggers update_effective_location and
```
bohdan-suprun: Does this verification rely on RMB internals?

bohdan-suprun: My concern, that someday RMB implementation can be changed so the expected behavior won't  be checked.

Could you please add a separate test to verify it?

julianladisch: canCalculateEffectiveLocationOnItemUpdate runs 56 examples where an item is created and then updated.
After insert the effective location is checked.
After update the effective location is checked.
If the trigger order changes these checks will fail.
Can you provide an example that is not covered by these 56 tests?

bohdan-suprun: > canCalculateEffectiveLocationOnItemUpdate runs 56 examples where an item is created and then updated.
After insert the effective location is checked.
After update the effective location is checked.

If I'm not missing something, this test verifies JSON only, but we have to verify that value in JSON is the same as in the FK column.

> If the trigger order changes these checks will fail.
Can you provide an example that is not covered by these 56 tests?

It is not about missing coverage, to me it is not clear why the test should fail if the order of triggers is changed, seems it relays on some RMB internal logic that can be changed in some release.

I'd prefer a separate test (or modifying existing one) that verifies JSON value is the same as in FK column on item create/update.

Probably I'm missing something.

PR #395, opened 2020-01-25
MODINVSTOR-435: Add indexes for callNumber and accessionNumber.

julianladisch:
Accents are not removed because they are relevant for libraries that use them.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -582,6 +582,14 @@
           "fieldName": "hrid",
           "tOps": "ADD"
         }
+      ],
+      "index": [
+        {
+          "fieldName": "callNumber",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        }
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -582,6 +582,14 @@
           "fieldName": "hrid",
           "tOps": "ADD"
         }
+      ],
```
bohdan-suprun: Do we need to update `fromModuleVersion` for the table then?

julianladisch: No, indexes are created and updated regardless of `fromModuleVersion`.

bohdan-suprun: Ok, what is the purpose of `fromModuleVersion` for table then? Is it required for table?

julianladisch: Yes, when the value of "withMetadata" or "withAuditing" changes `fromModuleVersion` must be updated to the current version to trigger the change.

PR #396, opened 2020-01-28
MODINVSTOR-433: Add 'Claimed returned' to allowed item statuses list.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-433: Backend: Claim returned: mark an item claim returned.

* Add `Claimed returned` to allowed item statuses list.
* Update `item-storage` and `item-sync` API versions.

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -231,7 +231,8 @@
             "On order",
             "Paged",
             "Declared lost",
-            "Order closed"
+            "Order closed",
+            "Claim returned"
           ]
         },
         "date": {
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -232,7 +232,7 @@
             "Paged",
             "Declared lost",
             "Order closed",
-            "Claim returned"
+            "Claimed returned"
           ]
         },
         "date": {
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -2,6 +2,8 @@ package org.folio.rest.api;
 
 import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
 import static java.net.HttpURLConnection.HTTP_OK;
+import static java.nio.file.Files.readAllBytes;
+import static java.nio.file.Paths.get;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.folio.HttpStatus.HTTP_CREATED;
 import static org.folio.rest.api.StorageTestSuite.TENANT_ID;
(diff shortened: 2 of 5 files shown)


ramls/item.json
```diff
@@ -231,7 +231,8 @@
             "On order",
             "Paged",
             "Declared lost",
-            "Order closed"
+            "Order closed",
+            "Claim returned"
```
marcjohnson-kint: Are there any tests for item status in this module?

bohdan-suprun: I don't think it make sense, it is a pass through operation and RMB is responsible of it.

marcjohnson-kint: What do you mean by it is a passthrough?

Does that mean we don't test at least some of the various statuses in the API tests?

bohdan-suprun: I meant that the only think that validates the status is RMB.  

No, we have a test for `Available` and `Checked out` statuses.

marcjohnson-kint: Hmm, ok, that seems a bit odd to me. Now that the list is constrained, this should probably be tested.

marcjohnson-kint: Based upon scenario 6, my interpretation is that the item status becomes `Claimed returned` and there needs to be a date recorded for when this last occurred.

Please can you follow up with Emma about the requirements for this?

bohdan-suprun: Thanks, good point. Posted a comment.

bohdan-suprun: @marcjohnson-kint, I have added a test to verify that item can be created with all the allowed status values.

bohdan-suprun: Changed to `Claimed returned`. The date and time when item was last marked as claimed returned seems is not required for now. Date and time when loan has been marked as claimed returned will be on loan record.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1913,6 +1919,31 @@ public void cannotPostSynchronousBatchWithoutStatus() {
...
+      is(status.value()));
+
+    JsonObject itemInStorage = itemsClient.getById(createdItem.getId()).getJson();
+    assertThat(itemInStorage.getJsonObject("status").getString("name"),
+      is(status.value()));
+  }
+
+  @SuppressWarnings("unused")
+  private Status.Name[] getAllowedItemStatuses() {
+    return Status.Name.values();
```
marcjohnson-kint: I think we need to be cautious with this kind of testing. It demonstrates that the API respects the generated code, rather than the generated coding produces the expected behaviour.

bohdan-suprun: Do you suggest to hardcode all the allowed statuses instead?

marcjohnson-kint: I suggest we either hard code the values or we use the source of truth which is the schema (as I think you have done in the mod-inventory change)?

bohdan-suprun: Test uses schema as source of allowed statuses now.

PR #397, opened 2020-01-31
MODINVSTOR-431 Settings > Inventory > Mode of issuance. Misc. renaming and source update of existing values

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINVSTOR-431

- rename modes of issuance
- create migration script

### DIFF
--- a/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
+++ b/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
@@ -1,39 +1,23 @@
 START TRANSACTION;
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{name}',
-			'"integrating resource"')
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"integrating resource"')
 WHERE it.jsonb->>'name' = 'Integrating Resource';
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{name}',
-			'"multipart monograph"')
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"multipart monograph"')
 WHERE it.jsonb->>'name' = 'Sequential Monograph';
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{name}',
-			'"serial"')
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"serial"')
 WHERE it.jsonb->>'name' = 'Serial';
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{name}',
-			'"single unit"')
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"single unit"')
 WHERE it.jsonb->>'name' = 'Monograph';
 
-UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{name}',
-			'"unspecified"')
-WHERE it.jsonb->>'name' = 'Other';
-
-UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
-SET jsonb = jsonb_set(it.jsonb,
-			'{source}',
-			'"folio"')
-WHERE it.jsonb->>'name' = 'unspecified';
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance
+SET jsonb = jsonb || '{"name": "unspecified", "source": "folio"}'
+WHERE jsonb->>'name' = 'Other';
 
 END TRANSACTION;
--- a/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
+++ b/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
@@ -1,0 +1,39 @@
+START TRANSACTION;
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{name}',
+			'"integrating resource"')
+WHERE it.jsonb->>'name' = 'Integrating Resource';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{name}',
+			'"multipart monograph"')
+WHERE it.jsonb->>'name' = 'Sequential Monograph';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{name}',
+			'"serial"')
+WHERE it.jsonb->>'name' = 'Serial';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{name}',
+			'"single unit"')
+WHERE it.jsonb->>'name' = 'Monograph';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{name}',
+			'"unspecified"')
+WHERE it.jsonb->>'name' = 'Other';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb,
+			'{source}',
+			'"folio"')
+WHERE it.jsonb->>'name' = 'unspecified';
+
+END TRANSACTION;
--- a/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
+++ b/src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
@@ -2,22 +2,22 @@ START TRANSACTION;
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
 SET jsonb = jsonb_set(it.jsonb, '{name}', '"integrating resource"')
-WHERE it.jsonb->>'name' = 'Integrating Resource';
+WHERE it.jsonb->>'id' = '4fc0f4fe-06fd-490a-a078-c4da1754e03a';
 
 UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
 SET jsonb = jsonb_set(it.jsonb, '{name}', '"multipart monograph"')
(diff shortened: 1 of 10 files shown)


src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
```diff
@@ -0,0 +1,39 @@
+START TRANSACTION;
```
bohdan-suprun: Hi @julianladisch, @adamdickmeiss, @skoczko, do we need migration scripts for reference data? 

Are the reference data being updated/reloaded every time the module installed/upgraded?

src/test/java/org/folio/rest/api/ModesOfIssuanceMigrationScriptTest.java
```diff
@@ -0,0 +1,57 @@
...
+public class ModesOfIssuanceMigrationScriptTest extends MigrationTestBase {
+  private static final String MIGRATION_SCRIPT = loadScript("renameModesOfIssuance.sql");
+
+  @Before
+  public void beforeEach() {
+    StorageTestSuite.deleteAll(modesOfIssuanceUrl(""));
+  }
+
+  @Test
+  public void canMigrateCopyNumbersToSingleValue() throws Exception {
```
bohdan-suprun:
```suggestion
  public void canMigrateModeOfIssuens() throws Exception {
```

Svitlana-Zmiivska: renamed

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
marcjohnson-kint: Personally, I don't think we can assume that the reference records will be reloaded during every upgrade, as it might overwrite changes made by users of the tenant.

Alas, automatically migrating the existing record has similar issues.

bohdan-suprun: Ok, do you think we should have migration scripts anyway?

> Alas, automatically migrating the existing record has similar issues.

Partially, I think the script will be executed only once and when module version will be updated, then it should not.

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
```diff
@@ -0,0 +1,23 @@
...
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"integrating resource"')
+WHERE it.jsonb->>'name' = 'Integrating Resource';
```
bohdan-suprun: Can we use id here instead?

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
```diff
@@ -0,0 +1,23 @@
...
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance AS it
+SET jsonb = jsonb_set(it.jsonb, '{name}', '"single unit"')
+WHERE it.jsonb->>'name' = 'Monograph';
+
+UPDATE ${myuniversity}_${mymodule}.mode_of_issuance
+SET jsonb = jsonb || '{"name": "unspecified", "source": "folio"}'
+WHERE jsonb->>'name' = 'Other';
+
+END TRANSACTION;
```
bohdan-suprun: I don't think that we need to execute update in a single transaction. Partial migration is also acceptable, since the records are not related to each other.

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
bohdan-suprun: Maybe both, `id` and `name` to prevent value overwriting as described @marcjohnson-kint  in comment above.

src/test/java/org/folio/rest/api/ModesOfIssuanceMigrationScriptTest.java
```diff
@@ -0,0 +1,57 @@
...
+    expectModeOfIssuance(modesOfIssuance.get(3), "single unit", "rdamodeissue");
+    expectModeOfIssuance(modesOfIssuance.get(4), "unspecified", "folio");
+  }
+
+  private void expectModeOfIssuance(JsonObject modeOfIssuance, String name, String source) {
+    Assert.assertThat(modeOfIssuance.getString("name"), is(name));
+    Assert.assertThat(modeOfIssuance.getString("source"), is(source));
+  }
+
+  private static void createModeOfIssuance(String name, String source)
```
bohdan-suprun:
```suggestion
  private void createModeOfIssuance(String name, String source)
```

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
marcjohnson-kint: > Ok, do you think we should have migration scripts anyway?
> 
> > Alas, automatically migrating the existing record has similar issues.
> 
> Partially, I think the script will be executed only once and when module version will be updated, then it should not.

That sole update could still overwrite changes that a tenant has made.

I've raised questions about this on the JIRA issue. I think we need decisions about that area before we can decide if we should provide an automated data upgrade script.

marcjohnson-kint: Based upon [Charlotte's comment](https://issues.folio.org/browse/MODINVSTOR-431?focusedCommentId=70032&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70032) on the story, this work should include a data upgrade script that always overwrites the name of these records, whether they have changed or not. Meaning we should only match on ID.

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
marcjohnson-kint: Based upon [Charlotte's comment](https://issues.folio.org/browse/MODINVSTOR-431?focusedCommentId=70032&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-70032) on the story, this work should include a data upgrade script that always overwrites the name of these records, whether they have changed or not. Meaning we should only match on ID.

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
marcjohnson-kint: @bohdan-suprun Why is partial migration acceptable? If updates to these records fails, will the module upgrade fail?

bohdan-suprun: @marcjohnson-kint, those records are not related to each other, record `A` is not depend on record `B` and the case when record `A` has been migrated but `B` hasn't, won't cause any issues.

Transaction in this context means that migration must be applied to all the records or reverted at all if at least one failed. Without using transaction all previous successful updates will be committed (i.e. in this case each update statement will be executed in it's own transaction).

>  If updates to these records fails, will the module upgrade fail?

Yes, upgrade will fail anyway, but successful updates will be kept.

marcjohnson-kint: > @marcjohnson-kint, those records are not related to each other, record `A` is not depend on record `B` and the case when record `A` has been migrated but `B` hasn't, won't cause any issues.
> 
> Transaction in this context means that migration must be applied to all the records or reverted at all if at least one failed. Without using transaction all previous successful updates will be committed (i.e. in this case each update statement will be executed in it's own transaction).

What impact could a partial success / failure have on operations folks who need to back out a failed module upgrade?

> > If updates to these records fails, will the module upgrade fail?
> 
> Yes, upgrade will fail anyway, but successful updates will be kept.

As far as I am aware FOLIO has not established the expectations around partial data upgrades, based upon the conversations I've been part of, folks would prefer the data was not partially updated upon failure.

@skoczko Would that fit with your understanding?

@bohdan-suprun Is it going to be particularly expensive to use a transaction for these updates? Are you concerned about the performance overhead of that?

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
Svitlana-Zmiivska: Changed

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
Svitlana-Zmiivska: Changed

src/main/resources/templates/db_scripts/renameModesOfIssuance.sql
bohdan-suprun: > As far as I am aware FOLIO has not established the expectations around partial data upgrades, based upon the conversations I've been part of, folks would prefer the data was not partially updated upon failure.

Hmm... that's make sense to me then. We should maybe mention it somewhere - that there should be no partial upgrades. 

> @bohdan-suprun Is it going to be particularly expensive to use a transaction for these updates? Are you concerned about the performance overhead of that?

No, I don't think this affects performance, my main idea was - try to migrate as much records as we can vs revert everything if at least one update failed. 

What do you think @Svitlana-Zmiivska, @marcjohnson-kint?

marcjohnson-kint: > > As far as I am aware FOLIO has not established the expectations around partial data upgrades, based upon the conversations I've been part of, folks would prefer the data was not partially updated upon failure.
> 
> Hmm... that's make sense to me then. We should maybe mention it somewhere - that there should be no partial upgrades.

I agree, FOLIO needs to agree upon that expectation and document it. I have raised it as a separate topic for the Tech Leads (the current upgrade topic has grown rather large and unwieldy)
 
> > @bohdan-suprun Is it going to be particularly expensive to use a transaction for these updates? Are you concerned about the performance overhead of that?
> 
> No, I don't think this affects performance, my main idea was - try to migrate as much records as we can vs revert everything if at least one update failed.
> 
> What do you think @Svitlana-Zmiivska, @marcjohnson-kint?

I think we should update all of them or none, if there is limited impact on performance, until we know what the correct expectations are.

bohdan-suprun: Ok, thanks @marcjohnson-kint. @Svitlana-Zmiivska, I'm resolving the conversation, no changes required in this case.

Svitlana-Zmiivska:
@marcjohnson-kint @bohdan-suprun  Do you have any other comments? Or can this be merged?

bohdan-suprun:
> @marcjohnson-kint @bohdan-suprun Do you have any other comments? Or can this be merged?

Hi @Svitlana-Zmiivska, I don't have, maybe @marcjohnson-kint has.

marcjohnson-kint:
> > @marcjohnson-kint @bohdan-suprun Do you have any other comments? Or can this be merged?
> 
> Hi @Svitlana-Zmiivska, I don't have, maybe @marcjohnson-kint has.

I don't have any further comments. This pull request needs approval from a code owner and then it can be merged.

PR #398, opened 2020-02-11
MODINVSTOR-443: Move locations and location-units to sample-data

julianladisch:
They are not reference data, but locations of a sample institution.


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -104,17 +110,16 @@ public void postTenant(TenantAttributes ta, Map<String, String> headers,
...
       }
       tl.withKey(SAMPLE_KEY).withLead(SAMPLE_LEAD);
-      tl.add("instances", "instance-storage/instances");
-      tl.withIdContent();
-      tl.add("holdingsrecords", "holdings-storage/holdings");
-      tl.add("items", "item-storage/items");
-      tl.add("instance-relationships", "instance-storage/instance-relationships");
+      for (String p : samplePaths) {
+        tl.add(p);
+      }
```
marcjohnson-kint, line 122: The previous code appears to create an explicit mapping between the type of record and the path.

How does the new code do this, as it appears to only add the path?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -30,13 +30,9 @@
   private static final String REFERENCE_LEAD = "ref-data";
 
   private static final Logger log = LoggerFactory.getLogger(TenantRefAPI.class);
-  final String[] refPaths = new String[]{
+  static final String[] refPaths = new String[]{
```
marcjohnson-kint, line 33:
```suggestion
  static final String[] refPaths = new String[] {
```

marcjohnson-kint:
@julianladisch Are you intending on returning to this pull request?

marcjohnson-kint:
@julianladisch Can this be updated and merged, or should it be closed?

marcjohnson-kint:
@julianladisch Is this pull request going to be merged during 2020 Q3?

marcjohnson-kint:
I'm closing this PR as it appears to be inactive

PR #399, opened 2020-02-19
PgUtil.streamGet for instances, holdings, items storage MODINVSTOR-438

adamdickmeiss:
Note: using snapshot of RMB which is not even released.

The /instance-storage/instances is problematic and is not using
the "optimized" version no more. Better make an RMB change for that
rather than trying to fix it here.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "8.1",
+      "version": "8.2",
       "handlers": [
         {
           "methods": ["GET"],
@@ -46,7 +46,7 @@
     },
     {
       "id": "holdings-storage",
-      "version": "4.0",
+      "version": "4.1",
       "handlers": [
         {
           "methods": ["GET"],
@@ -88,7 +88,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.2",
+      "version": "7.3",
       "handlers": [
         {
           "methods": ["GET"],
--- a/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
@@ -15,7 +15,6 @@ import javax.ws.rs.core.Response;
 import org.folio.cql2pgjson.CQL2PgJSON;
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
-import org.folio.rest.jaxrs.model.HoldingsRecords;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.rest.jaxrs.resource.HoldingsStorage;
 import org.folio.rest.persist.PgExceptionUtil;
@@ -39,6 +38,7 @@ import io.vertx.core.logging.Logger;
 import io.vertx.core.logging.LoggerFactory;
 import io.vertx.ext.sql.SQLConnection;
 import io.vertx.ext.sql.UpdateResult;
+import io.vertx.ext.web.RoutingContext;
 
 /**
  *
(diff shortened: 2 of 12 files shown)


adamdickmeiss:
Wow .. builing fails in seconds.. This is for preview only!

adamdickmeiss:
Failing because RMB is not yet released.

src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
```diff
@@ -75,28 +76,33 @@ public void getInstanceStorageInstances(
...
+          "title", query, offset, limit,
+          okapiHeaders, vertxContext, GetInstanceStorageInstancesResponse.class, asyncResultHandler);
+      } catch (Exception e) {
+        log.error(e.getMessage(), e);
+        asyncResultHandler.handle(io.vertx.core.Future.succeededFuture(
+          GetInstanceStorageInstancesResponse.
+            respond500WithTextPlain(e.getMessage())));
+      }
+      return;
     }
```
marcjohnson-kint: Does this mean that for CQL searches including `title` the response will not be streamed? Or has `PgUtil.getWithOptimizedSql` been changed to stream the response?

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -90,71 +90,19 @@ public void deleteHoldingsStorageHoldings(String lang,
   @Override
   public void getHoldingsStorageHoldings(
     int offset, int limit, String query, String lang,
-    Map<String, String> okapiHeaders,
+    RoutingContext routingContext, Map<String, String> okapiHeaders,
```
marcjohnson-kint: Is this a change in the code generation that RAML Module Builder uses or is `routingContext` an optional parameter?

src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
adamdickmeiss: Correct. PgUtil.getWithOptimizedSql can not be changed as is. It would need new parameters. Created RMB-573 for that work

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
adamdickmeiss: The routingContext is parameter for paths in property generate_routing_context. This facility has been around for quite some time. See RMB-570

adamdickmeiss:
Thankx for reviewing this.. But I need https://github.com/folio-org/raml-module-builder/pull/614 merged first.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
marcjohnson-kint: Thanks, I wondered if there was a standard way of configuring this.

src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
marcjohnson-kint: Thank you for raising that additional issue.

In the meantime, I think we need to let folks know about this limitation. Particularly for folks working on [MODINVSTOR-439](https://issues.folio.org/browse/MODINVSTOR-439) and related issues.

Does the `keyword` index also count as including `title` or is that separate and can be streamed?

adamdickmeiss: Note.. At this point, only sort by title, is using the special path. So keyword search (other) has no relevance.

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "8.1",
+      "version": "8.2",
```
bohdan-suprun: We also need to update API version in item-storage.raml.

adamdickmeiss: OK. That's quite error prone to have to specify this in two places when that's not checked by some tool. Updated now - for the three RAML files.

adamdickmeiss:
Nice to see it passes with RMB snapshot is now that has streamGet.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: Thanks. Yeah, indeed it.

PR #401, opened 2020-02-21
MODINVSTOR-441 Preceding titles ability to add "unconnected" and "connected" title (create, edit)

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINVSTOR-441

Create new table for preceding/succeeding titles
Create new crud endpoints for preceding/succeeding titles

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -154,6 +154,33 @@
       ]
     },
     {
+      "id": "instance-preceding-succeeding-titles",
+      "version": "0.1",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/preceding-succeeding-titles",
+          "permissionsRequired": ["inventory-storage.preceding-succeeding-titles.collection.get"]
+        }, {
+          "methods": ["GET"],
+          "pathPattern": "/preceding-succeeding-titles/{id}",
+          "permissionsRequired": ["inventory-storage.preceding-succeeding-titles.item.get"]
+        }, {
+          "methods": ["POST"],
+          "pathPattern": "/preceding-succeeding-titles",
+          "permissionsRequired": ["inventory-storage.preceding-succeeding-titles.item.post"]
+        }, {
+          "methods": ["PUT"],
+          "pathPattern": "/preceding-succeeding-titles/{id}",
+          "permissionsRequired": ["inventory-storage.preceding-succeeding-titles.item.put"]
+        }, {
+          "methods": ["DELETE"],
+          "pathPattern": "/preceding-succeeding-titles/{id}",
+          "permissionsRequired": ["inventory-storage.preceding-succeeding-titles.item.delete"]
+        }
+      ]
+    },
+    {
       "id": "instance-storage-batch",
       "version": "0.2",
       "handlers": [
@@ -1344,6 +1371,31 @@
       "description": "delete individual instance-relationship-type from storage"
     },
     {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.collection.get",
+      "displayName": "inventory storage - get preceding-succeeding-title collection",
+      "description": "get preceding-succeeding-title collection from storage"
+    },
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.get",
+      "displayName": "inventory storage - get individual preceding-succeeding-title",
+      "description": "get individual preceding-succeeding-title from storage"
+    },
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.post",
+      "displayName": "inventory storage - create individual preceding-succeeding-title",
+      "description": "create individual preceding-succeeding-title in storage"
+    },
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.put",
+      "displayName": "inventory storage - modify preceding-succeeding-title",
+      "description": "modify preceding-succeeding-title in storage"
+    },
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.delete",
+      "displayName": "inventory storage - delete individual preceding-succeeding-title",
+      "description": "delete individual preceding-succeeding-title from storage"
+    },
+    {
       "permissionName": "inventory-storage.identifier-types.collection.get",
       "displayName": "inventory storage - get identifier-type collection",
       "description": "get identifier-type collection from storage"
@@ -2091,7 +2143,12 @@
         "inventory-storage.service-points-users.item.put",
         "inventory-storage.service-points-users.item.delete",
         "inventory-storage.hrid-settings.item.get",
-        "inventory-storage.hrid-settings.item.put"
+        "inventory-storage.hrid-settings.item.put",
+        "inventory-storage.preceding-succeeding-titles.collection.get",
+        "inventory-storage.preceding-succeeding-titles.item.get",
+        "inventory-storage.preceding-succeeding-titles.item.post",
+        "inventory-storage.preceding-succeeding-titles.item.put",
+        "inventory-storage.preceding-succeeding-titles.item.delete"
       ]
     }
   ],
--- a/ramls/instanceprecedingsucceedingtitle.json
+++ b/ramls/instanceprecedingsucceedingtitle.json
@@ -1,0 +1,60 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "A relationship of a given type between a super-instance and a sub-instance",
+  "type": "object",
+  "properties": {
+    "id": {
+      "description": "unique ID of the preceding-succeeding title",
+      "type": "string",
+      "$ref": "uuid.json"
+    },
+    "precedingInstanceId": {
+      "description": "ID (a UUID) of the preceding instance",
+      "type": "string",
+      "$ref": "uuid.json"
+    },
+    "succeedingInstanceId": {
+      "description": "ID (a UUID) of the succeeding instance",
+      "type": "string",
+      "$ref": "uuid.json"
+    },
+    "title": {
+      "type": "string",
+      "description": "The primary title (or label) associated with the resource"
+    },
+    "hrid": {
+      "type": "string",
+      "description": "The human readable ID, also called eye readable ID. A system-assigned sequential ID which maps to the Instance ID"
+    },
+    "identifiers": {
+      "type": "array",
+      "description": "An extensible set of name-value pairs of identifiers associated with the resource",
+      "minItems": 0,
+      "items": {
+        "type": "object",
+        "properties": {
+          "value": {
+            "type": "string",
+            "description": "Resource identifier value"
+          },
+          "identifierTypeId": {
+            "type": "string",
+            "description": "UUID of resource identifier type (e.g. ISBN, ISSN, LCCN, CODEN, Locally defined identifiers)",
+            "$ref": "uuid.json"
+          }
+        },
+        "additionalProperties": false,
+        "required": [
+          "value",
+          "identifierTypeId"
+        ]
+      }
+    },
+    "metadata": {
+      "type": "object",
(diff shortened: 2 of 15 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -1343,6 +1370,31 @@
...
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.put",
+      "displayName": "inventory storage - modify preceding-succeeding-title",
+      "description": "modify preceding-succeeding-title in storage"
+    },
+    {
+      "permissionName": "inventory-storage.preceding-succeeding-titles.item.delete",
+      "displayName": "inventory storage - delete individual preceding-succeeding-title",
+      "description": "delete individual preceding-succeeding-title from storage"
+    },
```
bohdan-suprun: We should add those permissions to `inventory-storage.all` as well.

ramls/instanceprecedingsucceedingtitle.json
```diff
@@ -0,0 +1,63 @@
+{
```
bohdan-suprun: Is it possible to name file with `-` separator, e.g. `instance-preceding-succeeding-title.json` or similar?

ramls/instanceprecedingsucceedingtitles.json
```diff
@@ -0,0 +1,23 @@
...
+      "description": "List of instance relationships",
+      "id": "precedingSucceedingTitles",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "$ref": "instanceprecedingsucceedingtitle.json"
+      }
+    },
+    "totalRecords": {
+      "type": "integer"
```
bohdan-suprun: Please add description.

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -0,0 +1,72 @@
...
+
+import javax.ws.rs.core.Response;
+import java.util.Map;
+
+public class PrecedingSucceedingTitleAPI implements org.folio.rest.jaxrs.resource.PrecedingSucceedingTitles {
+
+  private static final String PRECEDING_SUCCEEDING_TITLE_TABLE = "preceding_succeeding_title";
+
+  @Override
+  public void getPrecedingSucceedingTitles(int offset, int limit, String query, String lang,
```
bohdan-suprun: Please add `@Validate` annotation.

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -0,0 +1,72 @@
...
+    PrecedingSucceedingTitle entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    if (precedingOrSucceedingInstanceIdIsEmpty(entity)) {
+      handleError(asyncResultHandler);
+      return;
+    }
+    PgUtil.put(PRECEDING_SUCCEEDING_TITLE_TABLE, entity, precedingSucceedingTitleId, okapiHeaders, vertxContext,
+      PutPrecedingSucceedingTitlesByPrecedingSucceedingTitleIdResponse.class, asyncResultHandler);
+  }
+
+  private boolean precedingOrSucceedingInstanceIdIsEmpty(PrecedingSucceedingTitle entity) {
```
bohdan-suprun:
```suggestion
  private boolean precedingAndSucceedingInstanceEmpty(PrecedingSucceedingTitle entity) {
```

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+    PrecedingSucceedingTitle precedingSucceedingTitle = new PrecedingSucceedingTitle(
+      instance1Id, instance2Id, TITLE, HRID, identifiers);
+
+    Response response = createReferenceRecord(
+      precedingSucceedingTitleUrl("").getPath(), precedingSucceedingTitle);
+    assertThat(response.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    assertPrecedingSucceedingTitleIsCreated(response, instance1Id, instance2Id, TITLE, HRID, identifiers);
+    deletePrecedingSucceedingTitle(response.getJson().getString("id"));
```
bohdan-suprun: You can use `org.folio.rest.api.StorageTestSuite#deleteAll(tenantId, tableName)` to remove entire table in the `@BeforeClass` method instead.

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -0,0 +1,72 @@
...
+  public void getPrecedingSucceedingTitles(int offset, int limit, String query, String lang,
+    Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.get(PRECEDING_SUCCEEDING_TITLE_TABLE, PrecedingSucceedingTitle.class, PrecedingSucceedingTitles.class, query, offset, limit,
+      okapiHeaders, vertxContext, GetPrecedingSucceedingTitlesResponse.class, asyncResultHandler);
+  }
+
+  @Override
+  public void postPrecedingSucceedingTitles(String lang, PrecedingSucceedingTitle entity,
+    Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    if (precedingOrSucceedingInstanceIdIsEmpty(entity)) {
```
bohdan-suprun: Is it possible to add a constraint for this to Postgre as well?

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -0,0 +1,315 @@
...
+
+    JsonObject instance1Response = createInstance("Title One", INSTANCE_TYPE_ID_TEXT);
+    JsonObject instance2Response = createInstance("Title Two", INSTANCE_TYPE_ID_TEXT);
+    String instance1Id = instance1Response.getString(Instance.ID_KEY);
+    String instance2Id = instance2Response.getString(Instance.ID_KEY);
+
+    JsonArray identifiers = new JsonArray();
+    identifiers.add(identifier(UUID_ISBN, "9781473619777"));
+
+    PrecedingSucceedingTitle precedingSucceedingTitle = new PrecedingSucceedingTitle(
```
bohdan-suprun: Are we going to store title, HRID and identifiers for connected case as well? What should happen if a succeeding/preceding instance is got updated with some new identifiers?

Svitlana-Zmiivska: No. we aren't. Removed these fields from test

ramls/instanceprecedingsucceedingtitle.json
Svitlana-Zmiivska: Yes, it is. In my opinion it is more readable also. However, all other json files use this style. And I am not sure it is good idea to use another style only for preceding-succeeding-titles json file.

bohdan-suprun: Well, yes. Let's keep naming in the same way for now. 

(I think we should have a tech debt to fix all the names).

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+    PgUtil.put(PRECEDING_SUCCEEDING_TITLE_TABLE, entity, precedingSucceedingTitleId, okapiHeaders, vertxContext,
+      PutPrecedingSucceedingTitlesByPrecedingSucceedingTitleIdResponse.class, asyncResultHandler);
+  }
+
+  private boolean precedingAndSucceedingInstanceEmpty(PrecedingSucceedingTitle entity) {
+    return ObjectUtils.isEmpty(entity.getPrecedingInstanceId()) && ObjectUtils.isEmpty(entity.getSucceedingInstanceId());
+  }
+
+  private void handleError(Handler<AsyncResult<Response>> asyncResultHandler) {
```
bohdan-suprun:
```suggestion
  private void handlePrecedingAndSucceedingEmptyError(Handler<AsyncResult<Response>> asyncResultHandler) {
```

src/main/resources/templates/db_scripts/alterPrecedingSucceedingTitle.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE ${myuniversity}_${mymodule}.preceding_succeeding_title
```
bohdan-suprun: I think we need a drop constraint statement right before creation. Because subsequent execution of the script will fail with a 'constraint already exists' error.

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -0,0 +1,297 @@
...
+import org.folio.rest.support.Response;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeoutException;
+
+public class PrecedingSucceedingTitleTest extends ResourceTestBase {
```
bohdan-suprun: Could you please make sure that following edge case is covered as well:
* PUT/DELETE/GET fails with error when specified wrong id;

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -0,0 +1,297 @@
...
+      instance1Response.getString(Instance.ID_KEY), instance2Response.getString(Instance.ID_KEY),
+      TITLE, HRID, identifiers);
+    Response postResponse = createReferenceRecord(entityPath, precedingSucceedingTitle);
+    assertThat(postResponse.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+
+    String entityUUID = postResponse.getJson().getString("id");
+
+    String updateProperty = PrecedingSucceedingTitle.TITLE_KEY;
+
+    testGetPutDeletePost(entityPath, entityUUID, precedingSucceedingTitle, updateProperty);
```
bohdan-suprun: Hi @Svitlana-Zmiivska, I'm not sure that the `preceding-succeeding` API falls under ReferenceRecordAPI term.

Usually `ReferenceRecordAPI` means a very basic CRUD API that used as a storage for static data (e.g. instance relationship types, identifiers types, locations, etc.).

Svitlana-Zmiivska: changed

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -0,0 +1,344 @@
...
+    JsonObject instance1Response = createInstance("Title One", INSTANCE_TYPE_ID_TEXT);
+    JsonObject instance2Response = createInstance("Title Two", INSTANCE_TYPE_ID_TEXT);
+    String instance1Id = instance1Response.getString(Instance.ID_KEY);
+    String instance2Id = instance2Response.getString(Instance.ID_KEY);
+
+    PrecedingSucceedingTitle precedingSucceedingTitle = new PrecedingSucceedingTitle(
+      instance1Id, instance2Id, null, null, null);
+
+    Response response = createReferenceRecord(
+      precedingSucceedingTitleUrl("").getPath(), precedingSucceedingTitle);
```
bohdan-suprun: Does it make sense to create a client for precedingSucceeding API? similar that we have for items/holdings/etc.

PR #402, opened 2020-02-24
MODINVSTOR-444: Add indexes for item and holdings call number.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-444 - add indexes for call number searching (eye readable).

Add indexes:

#### Holdings:
* fullCallNumber = prefix + callNumber + suffix
* callNumberAndSuffix = callNumber + suffix

#### Item:
* fullCallNumber = effectiveCNC.prefix + effectiveCNC.callNumber + effectiveCNC.suffix
* callNumberAndSuffix = effectiveCNC.callNumber + effectiveCNC.suffix

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -595,14 +595,14 @@
           "multiFieldNames": "callNumber, callNumberSuffix",
           "tOps": "ADD",
           "caseSensitive": false,
-          "removeAccents": true
+          "removeAccents": false
         },
         {
           "fieldName": "fullCallNumber",
           "multiFieldNames": "callNumberPrefix, callNumber, callNumberSuffix",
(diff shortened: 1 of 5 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -589,13 +589,27 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
```
julianladisch: `"removeAccents": false` is correct because the accents are relevant for some libraries. Libraries that don't use accents in the call number are not affected by `"removeAccents": false`.
Can you change `"removeAccents"` from `true` to `false` for all call number fields?

bohdan-suprun: Sure, changed.

PR #403, opened 2020-02-25
MODINVSTOR-449: Set default value for instance.discoverySuppress=false

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-449: Create migration script to ensure that suppress from discovery is always yes or no.

### Changes:
* Set default value to `false` for `instance.discoverySuppress` property;
* Update API versions for `instance-storage/batch/sync`;
* Add migration script to set `discoverySuppress=false` if value for property is not set;

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -74,7 +74,7 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
   private static final String TAG_VALUE = "test-tag";
   private static final String STATUS_UPDATED_DATE_PROPERTY = "statusUpdatedDate";
   private static final Logger log = LoggerFactory.getLogger(InstanceStorageTest.class);
-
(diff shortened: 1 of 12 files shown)


src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -2297,6 +2297,74 @@ public void cannotCreateACollectionOfInstancesWithHRIDFailure() throws Exception
...
+
+    instancesArray.add(uprooted(uprootedId));
+    instancesArray.add(smallAngryPlanet(smallAngryPlanetId)
+      .put(DISCOVERY_SUPPRESS, true));
+
+    final JsonObject instanceCollection = new JsonObject().put(INSTANCES_KEY,
+      instancesArray);
+
+    final CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+    client.post(instancesStorageSyncUrl(""), instanceCollection, TENANT_ID,
```
Svitlana-Zmiivska: How about creating a client for instancesStorageSync API?

PR #404, opened 2020-02-27
MODINVSTOR-447 Migrate connected preceding / succeeding titles from instance relationships to the new model

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINVSTOR-447

Purpose:

In order to persist connected and unconnected preceding / succeeding titles in one place, the existed connected preceding / succeeding titles need to be migrated from instance_relationships to preceding_succeeding_titles table

### DIFF
--- a/src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
+++ b/src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
@@ -1,0 +1,8 @@
+INSERT INTO ${myuniversity}_${mymodule}.preceding_succeeding_title (id, jsonb)
+  SELECT id, jsonb_build_object(
+    'id', jsonb->'id',
+    'metadata', jsonb->'metadata',
+    'precedingInstanceId', jsonb->'superInstanceId',
+    'succeedingInstanceId', jsonb->'subInstanceId')
+  FROM ${myuniversity}_${mymodule}.instance_relationship as ir
+  WHERE ir.instanceRelationshipTypeId ='cde80cc2-0c8b-4672-82d4-721e51dcb990';
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -763,6 +763,11 @@
       "run": "after",
       "snippetPath": "renameModesOfIssuance.sql",
       "fromModuleVersion": "19.0.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "migratePrecedingSucceedingTitles.sql",
+      "fromModuleVersion": "19.0.0"
     }
   ]
 }
--- a/src/test/java/org/folio/rest/api/PrecedingSucceedingTitleMigrationScriptTest.java
+++ b/src/test/java/org/folio/rest/api/PrecedingSucceedingTitleMigrationScriptTest.java
@@ -1,0 +1,80 @@
+package org.folio.rest.api;
+
+import static org.folio.rest.api.ItemDamagedStatusAPITest.TEST_TENANT;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import io.vertx.core.json.JsonObject;
+import org.folio.rest.api.entities.InstanceRelationship;
+import org.folio.rest.support.IndividualResource;
+import org.junit.Before;
+import org.junit.Test;
+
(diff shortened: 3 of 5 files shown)


src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
```diff
@@ -0,0 +1,22 @@
...
+  SELECT id, jsonb_build_object(
+    'id', jsonb->'id',
+	'metadata', jsonb_build_object(
+		'createdDate', jsonb->'metadata'->'createdDate',
+		'updatedDate', jsonb->'metadata'->'updatedDate'),
+    'precedingInstanceId', jsonb->'superInstanceId',
+	'succeedingInstanceId', jsonb->'subInstanceId'),
+   creation_date, created_by, superInstanceId, subInstanceId
+  FROM ${myuniversity}_${mymodule}.instance_relationship as ir
+  WHERE ir.jsonb->>'instanceRelationshipTypeId'='cde80cc2-0c8b-4672-82d4-721e51dcb990';
```
bohdan-suprun: In this case FK index will be used.

```suggestion
  WHERE ir.instanceRelationshipTypeId = 'cde80cc2-0c8b-4672-82d4-721e51dcb990';
```

src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
```diff
@@ -0,0 +1,22 @@
...
+INSERT INTO ${myuniversity}_${mymodule}.preceding_succeeding_title (id, jsonb,
+	   creation_date, created_by, precedingInstanceId, succeedingInstanceId)
+  SELECT id, jsonb_build_object(
+    'id', jsonb->'id',
+	'metadata', jsonb_build_object(
+		'createdDate', jsonb->'metadata'->'createdDate',
+		'updatedDate', jsonb->'metadata'->'updatedDate'),
+    'precedingInstanceId', jsonb->'superInstanceId',
+	'succeedingInstanceId', jsonb->'subInstanceId'),
+   creation_date, created_by, superInstanceId, subInstanceId
```
bohdan-suprun:
```suggestion
   
```

I think we can omit them, as they are populated by triggers.

src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
```diff
@@ -0,0 +1,22 @@
+START TRANSACTION;
+
+INSERT INTO ${myuniversity}_${mymodule}.preceding_succeeding_title (id, jsonb,
+	   creation_date, created_by, precedingInstanceId, succeedingInstanceId)
+  SELECT id, jsonb_build_object(
+    'id', jsonb->'id',
+	'metadata', jsonb_build_object(
+		'createdDate', jsonb->'metadata'->'createdDate',
+		'updatedDate', jsonb->'metadata'->'updatedDate'),
```
bohdan-suprun: Can it be simplified to?

```suggestion
	'metadata', jsonb->'metadata'
```

src/main/resources/templates/db_scripts/migratePrecedingSucceedingTitles.sql
```diff
@@ -0,0 +1,22 @@
...
+	'succeedingInstanceId', jsonb->'subInstanceId'),
+   creation_date, created_by, superInstanceId, subInstanceId
+  FROM ${myuniversity}_${mymodule}.instance_relationship as ir
+  WHERE ir.jsonb->>'instanceRelationshipTypeId'='cde80cc2-0c8b-4672-82d4-721e51dcb990';
+
+DELETE FROM ${myuniversity}_${mymodule}.instance_relationship WHERE
+  jsonb->>'instanceRelationshipTypeId'='cde80cc2-0c8b-4672-82d4-721e51dcb990';
+
+DELETE FROM ${myuniversity}_${mymodule}.instance_relationship_type WHERE
+  id='cde80cc2-0c8b-4672-82d4-721e51dcb990';
```
bohdan-suprun: I think we should postpone removing old data, because UI still uses the old way. We should fill a separate jira to remove them after whole migration.

Svitlana-Zmiivska: It seems to me it will be better to remove old data in the separate story. 
I've created https://issues.folio.org/browse/MODINVSTOR-451

PR #405, opened 2020-02-27
MODINVSTOR-439: Create bulk download endpoint to save id UUIDs

kaladay:
This currently provides a working solution utilizing `PgUtil.streamGet()` from RMB-500.
I have concerns with the resource utilization (RAM, network, and time) resulting from the current RMB design and I believe that to properly meet the expectations of PO, functionality in `PgUtil.streamGet()` needs to be expanded to only fetch a single id per row and not the entire jsonb object per row (see [RMB-581](https://issues.folio.org/browse/RMB-581)).

Provides a new end-point: `/instance-bulk/ids`.

This will accept two special URL parameters:
- **type**: Currently only 'id' but can be expaned to include 'hrid'.
- ~**format**: Either 'raw' or 'base64'.~ (Removed due to base64 being larger than raw.)

This also accept other URL parameters:
- **query**: for filtering.
- **lang**: ~for filtering.~ (see [MODINVSTOR-198](https://issues.folio.org/browse/MODINVSTOR-198))

Adds new permission: 'inventory-storage.instance-bulk.ids.get'.


src/main/java/org/folio/rest/impl/InstanceBulkAPI.java
```diff
@@ -0,0 +1,66 @@
...
+        "ids", routingContext, okapiHeaders, vertxContext);
+    } catch (Exception e) {
+      LOG.error(e.getMessage(), e);
+      response.setStatusCode(HttpStatus.HTTP_INTERNAL_SERVER_ERROR.toInt());
+      response.putHeader(HttpHeaders.CONTENT_TYPE, "text/plain");
+      response.end(e.toString());
+    }
+  }
+
+  private Class<?> getItemClass(InstanceBulkIdsGetFormat format) {
```
julianladisch: We have instances, holdings and items, the method name is somewhat misleading. What about `getIDClass`?

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,458 @@
...
+
+  private Map<String, JsonObject> manyMoons(int total) {
+    return manyMoons(total, InstanceBulkIdsGetField.ID,
+      InstanceBulkIdsGetFormat.RAW);
+  }
+
+  private Map<String, JsonObject> manyMoons(int total,
+      InstanceBulkIdsGetField field, InstanceBulkIdsGetFormat format) {
+      Map<String, JsonObject> moons = new HashMap<String, JsonObject>();
+      Base64.Encoder encoder = Base64.getEncoder();
```
julianladisch: Can you fix the indentation?

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,458 @@
...
+            } else {
+              sequenceSet.completeExceptionally(r.cause());
+            }
+          });
+    });
+
+    try {
+      sequenceSet.get(2, SECONDS);
+    } catch (Exception e) {
+      fail(e.getMessage());
```
julianladisch: This does not swallow the stack trace:
```suggestion
      fail(e);
```

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,458 @@
...
+
+    int totalMoons = 20;
+    int expectedMatches = 11;
+    Map<String, JsonObject> moons = manyMoons(totalMoons,
+      InstanceBulkIdsGetField.ID,
+      InstanceBulkIdsGetFormat.RAW);
+    createManyMoons(moons);
+
+    String query = "query=keyword%20all%20%22Moon%20%231%2A%22";
+    URL getInstanceUrl = instanceBulkUrl("/ids?type=id&format=raw&" + query);
```
julianladisch: Can you use `urlEncode` for readability?
Example: https://github.com/folio-org/mod-inventory-storage/blob/v18.2.3/src/test/java/org/folio/rest/api/ContributorTypesTest.java#L32

    String query = urlEncode("keyword all \"Moon #1*\"");
    URL getInstanceUrl = instanceBulkUrl("/ids?type=id&format=raw&query=" + query);

julianladisch: The same for all other test methods with url encoded query strings.

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "8.1",
+      "version": "8.2",
```
bohdan-suprun: I don't think that we need to update neither "item-storage" nor "holdings-storage" nor "instances-storage" since nothing has updated in their schema.

ramls/instance-bulk.raml
```diff
@@ -0,0 +1,50 @@
...
+version: v0.1
+protocols: [ HTTP, HTTPS ]
+baseUri: http://localhost
+
+documentation:
+  - title: Inventory Storage Instance Bulk Download API
+    content: <b>API for downloading a bulk set of Instance IDs</b>
+
+types:
+  id: !include id.json
```
bohdan-suprun: I don't see where the type is used nor the schema itself.

ramls/instance-bulk.raml
```diff
@@ -0,0 +1,50 @@
...
+        searchable: {description: "with valid searchable fields", example: "name=aaa"},
+        validate
+      ]
+      queryParameters:
+        field:
+          description: Field
+          type: string
+          enum: [id]
+          default: id
+          required: false
```
bohdan-suprun: I'd remove this property at all for now, it can be added once required. What do you think?

ramls/instance-bulk.raml
```diff
@@ -0,0 +1,50 @@
...
+        format:
+          description: Format
+          type: string
+          enum: [raw, base64]
```
bohdan-suprun: I'd model this as content negotiation:
- /instance-bulk/ids - returns in raw format;
- /instance-bulk/ids.base64 - returns in base64 format;

What do you think?

src/main/java/org/folio/rest/impl/InstanceBulkAPI.java
```diff
@@ -0,0 +1,66 @@
...
+
+    try {
+      CQLWrapper wrapper = getCQL(query);
+      PgUtil.streamGet(INSTANCE_TABLE, getItemClass(format), wrapper, null,
+        "ids", routingContext, okapiHeaders, vertxContext);
+    } catch (Exception e) {
+      LOG.error(e.getMessage(), e);
+      response.setStatusCode(HttpStatus.HTTP_INTERNAL_SERVER_ERROR.toInt());
+      response.putHeader(HttpHeaders.CONTENT_TYPE, "text/plain");
+      response.end(e.toString());
```
bohdan-suprun: We should have a generated delegate class for this purpose:

```suggestion
      asyncResultHandler.handle(Future.succeededFuture(GetInstanceBulkIdsResponse
      .respond500WithTextPlain(e.getMessage());
```

ramls/instance-bulk.raml
julianladisch: The plan is to extend it for `field=hrid`. This is also an id therefore `/instance-bulk/ids` is a reasonable API path name.

ramls/instance-bulk.raml
julianladisch: I don't see an advantage of `/instance-bulk/ids.base64`.

src/main/java/org/folio/rest/support/InstanceIDBase64.java
```diff
@@ -0,0 +1,40 @@
...
+    id = "";
+  }
+
+  public InstanceIDBase64(String id) {
+    this.id = id;
+  }
+
+  public String getId() {
+    Base64.Encoder encoder = Base64.getEncoder();
+    return encoder.encodeToString(id.getBytes());
```
bohdan-suprun: Is it really gives us a benefit? 

With this approach:
* An UUID string is 72 bytes 
* Base64 encoded string is 96 bytes. 

It would work if use this approach: https://stackoverflow.com/questions/17893609/convert-uuid-to-byte-that-works-when-using-uuid-nameuuidfrombytesb, (convert mostSigBits and leastSigBits only) but i'm not sure if it will work for each client.

What do you think?

ramls/instance-bulk.raml
bohdan-suprun: The MODINVSTOR-439 states that HRID is out of scope now, so I don't see any benefit in adding a flag that is not used. Also, requirements is not clear enough, maybe it will be ok to have both ID + HRID in the response.

kaladay: I was thinking _Can take a query parameter to allow HRIDs in the future_ meant that I had to provide the initial infrastructure like this field query parameter, even if it only support `id` at this time.
As for _ID + HRID_, I just want to confirm, are you suggesting retrieving multiple columns per row (id column, hrid column)?

src/main/java/org/folio/rest/support/InstanceIDBase64.java
kaladay: I am thinking not. I never checked the size and only provided this based on the discussion in the ticket.
Though, I find that article interesting. I did compare the generated base64 UUID to one from command line and they had the same results (`echo -n "f31a36de-fcf8-44f9-87ef-a55d06ad21ae" | base64`). I would have though java to be more architecture independent and not impose byte order like this.

Removing this base64 support will make the tests and parameters a bit simpler.
I will remove the base64 support and therefore the _format_ query parameter.

descriptors/ModuleDescriptor-template.json
kaladay: These changes are an artifact of initially basing this off on another PR. I must have forgotten **not** to commit these changes.

ramls/instance-bulk.raml
kaladay: It was used up until here: https://github.com/folio-org/mod-inventory-storage/pull/405/commits/9e8151e67207a15c5c2ef7cfa8cd2f998704e3a5

I seem to have missed removing this one.

ramls/instance-bulk.raml
bohdan-suprun: Ok, let's leave it for now, since it is optional.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,400 @@
...
+    assertThat(ids.size(), is(expectedMatches));
+    assertThat(collection.getInteger("totalRecords"), is (expectedMatches));
+
+    for (int i = 0; i < ids.size(); i++) {
+      JsonObject idObject = ids.getJsonObject(i);
+      assertThat(moons.containsKey(idObject.getString("id")), is(true));
+    }
+  }
+
+  private void setInstanceSequence(long sequenceNumber) {
```
bohdan-suprun: We can remove this method. It is needed for some tests for HRID logic.

Maybe I'm missing something.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,400 @@
...
+  public void removeGeneratedEntities(TestContext context) {
+    final Async async = context.async();
+    List<CompletableFuture<Response>> cfs =
+      new ArrayList<CompletableFuture<Response>>();
+
+    instanceIdsToRemoveAfterTest.forEach(id -> cfs.add(client
+      .delete(instancesStorageUrl("/" + id), TENANT_ID)));
+    CompletableFuture.allOf(cfs.toArray(new CompletableFuture[cfs.size()]))
+      .thenAccept(v -> async.complete());
+  }
```
bohdan-suprun: The `TestBaseWithInventoryUtil` should remove the entire table before each test, so we can remove this method.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,400 @@
...
+
+      JsonArray langs = new JsonArray();
+      int l = 0;
+      if (i % 3 == 0) {
+        l = 2;
+      }
+      else if (i % 2 == 0) {
+        l = 1;
+      }
+      for (int j = 0; j < languages.get(l).length; j++) {
```
bohdan-suprun: Is it important for the test to have this randomization for language? Just think if we can simplify it a bit.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,400 @@
...
+    return moons;
+  }
+
+  public Buffer bufferedCheckURLs(TestContext context, URL url,
+      int codeExpected) {
+    String accept = "application/json";
+    return bufferedCheckURLs(context, url, codeExpected, accept);
+  }
+
+  private Buffer bufferedCheckURLs(TestContext context, URL url,
```
bohdan-suprun: Can it be replaced with the default `OkapiClient` and `ResponseHandler.any`:
```
client.get(url, TENANT_ID, ResponseHandler.any(getCompleted));

Response response = getCompleted.get(5, SECONDS);

assertThat(response.getStatusCode(), is(codeExpected));
assertThat(response.getContentType(), is(accept));
```

src/main/java/org/folio/rest/support/InstanceIDBase64.java
julianladisch: `echo "0 f31a36de-fcf8-44f9-87ef-a55d06ad21ae" | xxd -r | base64`
`8xo23vz4RPmH76VdBq0hrg==`
You need to call xxd -r to undo the hexdump. Then you can invoke base64 and will get the correct and shorter result as explained on the Jira issue.
The same bug is in the Java implementation. Do not use the hexdump String version, use the bits methods that are architecture and client independent.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
kaladay: The reason for these tests were to do an explicit stream test.
If the explicit buffered stream test is not needed, then the tests like `canGetInstanceBulkOfIdStream` could be removed entirely (being that they are otherwise redundant).

Is there a way to explicitly test that the stream is in fact streaming or is that unnecessary at this scope?
Should I remove this explicit stream entirely?

src/test/java/org/folio/rest/api/InstanceBulkTest.java
kaladay: It is not random, it is just using every threes, every evens, and every odds and rotating between them.

I originally wrote this when I was expecting to use the `lang` url parameter.
I wanted a simple and easy way to have some arbitrary but predictable language generator.

I was uncertain if I would still need to test languages via the query parameter after considering [MODINVSTOR-198](https://issues.folio.org/browse/MODINVSTOR-198).

Should I remove this entirely and the language tests for now?

src/test/java/org/folio/rest/api/InstanceBulkTest.java
julianladisch: Testing whether PgUtil.streamGet is implemented correctly is unnecessary at this scope. Such tests should go into RMB's [PgUtilIT](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/java/org/folio/rest/persist/PgUtilIT.java).

src/test/java/org/folio/rest/api/InstanceBulkTest.java
bohdan-suprun: I have not seen any logic related to this attribute, I think it is fine to remove/simplify this test.

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,267 @@
...
+    createManyMoons(moons);
+
+    String query = urlEncode("keyword all \"Planet #1*\"");
+    URL getInstanceUrl = instanceBulkUrl("/ids?type=id&query=" + query);
+
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+    client.get(getInstanceUrl, TENANT_ID, json(getCompleted));
+
+    Response response = getCompleted.get(5, SECONDS);
+    validateMoonsResponse(response, expectedMatches, moons);
```
bohdan-suprun: Can we check instead that response is empty just for batter readability?

src/test/java/org/folio/rest/api/InstanceBulkTest.java
```diff
@@ -0,0 +1,267 @@
...
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+
+    client.post(instancesStorageUrl(""), instanceToCreate,
+      TENANT_ID, json(getCompleted));
+
+    Response response = getCompleted.get(2, SECONDS);
+
+    assertThat(String.format("Create instance failed: %s", response.getBody()),
+      response.getStatusCode(), is(201));
+  }
```
bohdan-suprun: You can use `instancesClient.create(json)` instead.

marcjohnson-kint:
@kaladay  This pull request has been approved. What remains to be done before this can be merged?

kaladay:
@marcjohnson-kint It is done as far as I am concerned and ready to be merged.

PR #407, opened 2020-03-03
MODINVSTOR-413: Add "isbn" normalized index

julianladisch:
Requires RMB with https://github.com/folio-org/raml-module-builder/pull/618 = https://issues.folio.org/browse/RMB-575


marcjohnson-kint:
@adamdickmeiss @julianladisch Am I correct in my understanding that this change is not to be merged until it has been performance tested?

If so, please either close the pull request or mark it as do not merge.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -742,6 +747,11 @@
     }
   ],
   "scripts": [
+    {
+      "run": "before",
+      "snippetPath": "isbn.sql",
```
kaladay: Would it be better to rename the `isbn.sql` file and this line to be a bit more descriptive, such as isbnNormalize.sql (or something similar)?

julianladisch: This is the .sql file for the isbn field declared in line 439.
When we also add support for issn and isxn we may rename the file to isbn-issn-isxn.sql.

julianladisch:
Manual performance test using mod-inventory-storage's sample data:
CQL: `isbn=978-13*`
SQL: `WHERE to_tsvector('simple', normalize_isbns(jsonb->'identifiers')) @@ tsquery_phrase(normalize_digits('978-13*')) LIMIT 10 OFFSET 0`
Index usage analysis:
```
explain analyze select jsonb from instance WHERE to_tsvector('simple', normalize_isbns(jsonb->'identifiers')) @@ tsquery_phrase(normalize_digits('978-13*'));
                                                            QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------
 Bitmap Heap Scan on instance  (cost=20.08..35.66 rows=10 width=32) (actual time=0.021..0.023 rows=3 loops=1)
   Recheck Cond: (to_tsvector('simple'::regconfig, normalize_isbns((jsonb -> 'identifiers'::text))) @@ '''97813'':*'::tsquery)
   Heap Blocks: exact=2
   ->  Bitmap Index Scan on instance_isbn_idx_ft  (cost=0.00..20.07 rows=10 width=0) (actual time=0.015..0.015 rows=3 loops=1)
         Index Cond: (to_tsvector('simple'::regconfig, normalize_isbns((jsonb -> 'identifiers'::text))) @@ '''97813'':*'::tsquery)
 Planning time: 0.544 ms
 Execution time: 0.039 ms
```
The instance_isbn_idx_ft index is used and quickly finds the 3 instances where the isbn starts with 97813.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: I think it is a good idea to rename the file to describe what it is intended to do.

As I understand it, it defines a function for normalising ISBN values, maybe `create-normalize-isbns-function.sql`?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1268,6 +1268,31 @@ public void canSearchByTitleAndBarcodeWithMissingHoldingsAndItemsAndStillGetInst
...
+
+  @Test
+  public void canSearchByIsbn4() {
+    canSort("isbn = 9780* sortBy title", "Interesting Times", "Temeraire");
+  }
+
+  @Test
+  public void canSearchByIsbn5() {
+    canSort("isbn = 9-7-8-055-2167-543", "Interesting Times");
+  }
```
marcjohnson-kint: Please can the names of these tests describe the situation being demonstrated, rather than a numeric postfix.

For example, if it is searching by a value without hyphens, maybe it could be called `canSearchForISBNWithoutHyphens`?

What do you think?

julianladisch: Fixed.

src/main/resources/templates/db_scripts/isbn.sql
```diff
@@ -0,0 +1,8 @@
+-- Usage: SELECT normalize_isbns(jsonb->'identifiers') FROM instance
+-- This takes each ISBN, normalizes it using RMB's normalize_digits(text),
+-- and concatenates the results using a space as separator.
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.normalize_isbns(jsonb_array jsonb) RETURNS text AS $$
+  SELECT string_agg(${myuniversity}_${mymodule}.normalize_digits(value->>'value'), ' ')
+  FROM jsonb_array_elements($1)
```
bohdan-suprun: Can it be changed to:
```suggestion
  FROM jsonb_array_elements($1) as value
```

for batter readability?

julianladisch: Yes, name changed.

src/main/resources/templates/db_scripts/schema.json
julianladisch: Renamed to create-normalize_isbns-function.sql

PR #409, opened 2020-03-09
MODINVSTOR-450: Improve slow queries, migrating from Daisy to Edelweiss

kaladay:
This does not perform VACUUM, but a VACUUM before execution of update scripts
is highly recommended.

*EDIT:* Many of the scripts that were combined could not be combined due to versions that they need to execute on. This, unfortunately, prevents a lot of the performance gains.

Optimize the query using the following strategies:
- Disable auto-updating triggers.
- Combine similar queries (same set, based on WITH clause).
- Utilize function and trigger to perform if/then conditions.

Overall performances improvements using a test database showed:
- Before these changes: ~53 minutes.
- After these changes: ~32 minutes.
* This test case is not exhaustive.
  (~3.5 million items, ~1 million instances, most other tables empty.)

This does not optimize by disabling indexes, see notes below for reasons.

Explicitly cast manually created jsonb in `renameModesOfIssuance.sql`.
This avoids warnings and any possible problem implied by the warnings.

Disabling the auto-updating triggers is helpful because the disabled tasks do
not need to be performed during this manual update. The data is already assured.
If the data is not assured, this update will therefore do so already.
This saves notable time. See examples below.

Combining similar queries saves the expensive of running multiple loops.
This only works for cases when the WHERE clauses are identical.
Multiple jsonb columns are being updated.
The concat operator is useful in this regards.

Some of the queries can be reduced if IF/THEN conditions are allowed in SQL.
In the case of PGSQL functions, IF/THEN conditions are allowed.
Temporarily create a PGSQL function to effectively combine two similar queries
into a single query. This reduces the number of loops.

In early investigations, I attempted to disable indexes during the update
process. This process worked reasonably well and showed an ~50% reduction in
time spent. The problem is that the gains were inconsistent due to unclear
reasons and the indexes eventually do need to get updated (via a
background-process, a SELECT, or a manual REINDEX). The suspected cause of the
difference is that when the database is very dirty, queries perform 50% slower.
By running a VACUUM before calling the upgrade scripts, the database can be
put into a clean state that improves performance of all queries. This has
an initial cost of the time it takes to perform a vacuum. This vacuum may
take anywhere from 1 minute to 30 minutes!

For reference, the code for disabling/enabling indexes looks like:
```
UPDATE pg_index SET indisready = false
WHERE indrelid = '${myuniversity}_${mymodule}.item'::regclass::oid;

UPDATE pg_index SET indisready = true
WHERE indrelid = '${myuniversity}_${mymodule}.item'::regclass::oid;
```

Example With Triggers Enabled:
```
--------------------------------------------------------------------------------------------------------------------------------------
 Update on item it  (cost=2.27..3687419.57 rows=3500000 width=212) (actual time=723773.371..723773.371 rows=0 loops=1)
   ->  Hash Join  (cost=2.27..3687419.57 rows=3500000 width=212) (actual time=6.378..137812.680 rows=3500000 loops=1)
         Hash Cond: (it.holdingsrecordid = hr.id)
         ->  Seq Scan on item it  (cost=0.00..3534985.00 rows=3500000 width=1753) (actual time=4.070..71617.172 rows=3500000 loops=1)
         ->  Hash  (cost=2.12..2.12 rows=12 width=54) (actual time=1.183..1.183 rows=12 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 18kB
               ->  Seq Scan on holdings_record hr  (cost=0.00..2.12 rows=12 width=54) (actual time=1.138..1.160 rows=12 loops=1)
 Planning time: 29.087 ms
 Trigger set_id_in_jsonb: time=35607.609 calls=3500000
 Trigger set_item_md_json_trigger: time=48997.533 calls=3500000
 Trigger update_effective_location: time=14846.184 calls=3500000
 Trigger update_item_references: time=35915.253 calls=3500000
 Trigger update_item_status_date: time=31638.725 calls=3500000
 Execution time: 723777.654 ms
(14 rows)
Time: 723816.910 ms (12:03.817)
```

Example With Triggers Disabled:
```
--------------------------------------------------------------------------------------------------------------------------------------
 Update on item it  (cost=2.27..4609007.56 rows=4370069 width=212) (actual time=627148.629..627148.629 rows=0 loops=1)
   ->  Hash Join  (cost=2.27..4609007.56 rows=4370069 width=212) (actual time=0.531..159966.331 rows=3500000 loops=1)
         Hash Cond: (it.holdingsrecordid = hr.id)
         ->  Seq Scan on item it  (cost=0.00..4418679.69 rows=4370069 width=1753) (actual time=0.013..96735.732 rows=3500000 loops=1)
         ->  Hash  (cost=2.12..2.12 rows=12 width=54) (actual time=0.488..0.488 rows=12 loops=1)
               Buckets: 1024  Batches: 1  Memory Usage: 18kB
               ->  Seq Scan on holdings_record hr  (cost=0.00..2.12 rows=12 width=54) (actual time=0.477..0.483 rows=12 loops=1)
 Planning time: 0.270 ms
 Execution time: 627151.287 ms
(9 rows)
Time: 627151.896 ms (10:27.152)
```


kaladay:
I was so focused on the SQL itself, I overlooked the schema.json and versions.
Combining some of these SQL scripts may not simply be just about the WHERE clause.
There is a version issue as well.
I may need to avoid merging some of the SQL scripts as a result.

src/main/resources/templates/db_scripts/migrateItemCopyNumberToSingleValue.sql
```diff
@@ -1,12 +1,40 @@
...
-WHERE it.jsonb->>'copyNumbers' IS NOT NULL AND jsonb_array_length(it.jsonb->'copyNumbers') > 0;
-
--- Remove the copyNumbers property even if it has zero length.
-UPDATE ${myuniversity}_${mymodule}.item AS it
-  SET jsonb = it.jsonb - 'copyNumbers'
-WHERE it.jsonb->'copyNumbers' IS NOT NULL;
+
+UPDATE pg_trigger
+  SET tgenabled = 'D'
+WHERE tgrelid IN (SELECT oid FROM pg_class WHERE relname = 'item')
```
julianladisch:
```suggestion
WHERE tgrelid = '${myuniversity}_${mymodule}.item'::regclass
```
There may be other tenants and modules that also have an 'item' table, always restrict to the current schema. The same for all other tgrelid occurrences.

src/main/resources/templates/db_scripts/migrateItemCopyNumberToSingleValue.sql
```diff
@@ -1,12 +1,40 @@
...
+CREATE TRIGGER temp_upgrade_items_trigger
+  BEFORE UPDATE ON ${myuniversity}_${mymodule}.item
+  FOR EACH ROW EXECUTE PROCEDURE ${myuniversity}_${mymodule}.temp_upgrade_items();
+
+UPDATE ${myuniversity}_${mymodule}.item
+  SET id = id
+WHERE jsonb->'copyNumbers' IS NOT NULL;
+
+DROP TRIGGER temp_upgrade_items_trigger ON ${myuniversity}_${mymodule}.item;
+DROP FUNCTION ${myuniversity}_${mymodule}.temp_upgrade_items;
```
julianladisch: Why advantages has the plpgsql trigger function over a SQL CASE?
```
UPDATE ${myuniversity}_${mymodule}.item
SET jsonb = CASE WHEN jsonb_array_length(jsonb->'copyNumbers') > 0
                 THEN jsonb_set(jsonb - 'copyNumbers', '{copyNumber}', jsonb#>'{copyNumbers, 0}')
                 ELSE jsonb - 'copyNumbers'
            END
WHERE jsonb->'copyNumbers' IS NOT NULL;
```

kaladay: Thanks! I wasn't aware CASE was available outside of pgsql.
This greatly simplifies the work and allows me to avoid the undesirable function and trigger.

src/main/resources/templates/db_scripts/migrateItemCopyNumberToSingleValue.sql
kaladay: Thanks for catching this oversight, I was using `::regclass::oid` before and forgot to use it here (as well as the `${myuniversity}_${mymodule}`).
Updated.

marcjohnson-kint:
@kaladay I think I've likely lost the thread on this. Please could you help me and summarise where we are with this pull request.

kaladay:
@marc-johnson the situation is that after extensive review, the only optimization I was able to safely implement is the disabling of triggers for a given table during the update/migrate process for that table. I believe it is safe to do so because the triggers being disabled do not apply to the upgrade process. They perform extra redundant tasks that was already assured for existing data. The upgrade/migrate process is explicitly assuring the state of the new data.

There is also a case where I was able to safely combine two separate queries into a single query, saving resources (`migrateItemCopyNumberToSingleValue.sql`).

I think I also need to double check the changes in `populateEffectiveLocationForExistingItems.sql`.

marcjohnson-kint:
> @marc-johnson the situation is that after extensive review, the only optimization I was able to safely implement is the disabling of triggers for a given table during the update/migrate process for that table. 

Ok, thanks for clarifying where we are.

> I believe it is safe to do so because the triggers being disabled do not apply to the upgrade process. They perform extra redundant tasks that was already assured for existing data. The upgrade/migrate process is explicitly assuring the state of the new data.

Ah, so the belief is that the trigger being triggered is redundant overhead because the state is already correct and will not be changed during that upgrade script?

> There is also a case where I was able to safely combine two separate queries into a single query, saving resources (`migrateItemCopyNumberToSingleValue.sql`).
> 
> I think I also need to double check the changes in `populateEffectiveLocationForExistingItems.sql`.

Thanks, it sounds like there is a plan for moving this change forward.

kaladay:
> Ah, so the belief is that the trigger being triggered is redundant overhead because the state is already correct and will not be changed during that upgrade script?

Yes, that is correct.

The `populateEffectiveLocationForExistingItems.sql` is written the way it is because it needs to explicitly assure that `effectivelocationid` is still populated/updated with the triggers being disabled.

marcjohnson-kint:
@julianladisch Do you have any concerns with this change as it stands?

marcjohnson-kint:
@kaladay @julianladisch @bohdan-suprun  @Svitlana-Zmiivska  @skoczko What do folks want to do with this pull request?

As I understand it, changing existing migration scripts that have likely already been executed is unlikely to change the behaviour of the system as it is.

That leaves me to think that we need to instead, try to codify the learning from this investigation into guidance for future migration scripts.

What do folks think?

bohdan-suprun:
@marcjohnson-kint, I agree that we need to put our findings somewhere. And also agree that this unlikely will be executed during upcoming prod upgrades.

It would be also useful if we save the migration scripts as a template, so everyone can copy the template and specify just migration code, and operations with triggers will be already in the file.

marcjohnson-kint:
> It would be also useful if we save the migration scripts as a template, so everyone can copy the template and specify just migration code, and operations with triggers will be already in the file.

I'm ok with the idea of a template. I wonder how much it can generalised.

Does which triggers are ok to disable depend upon the specific migration or is it universal that some can be, all can be?

kaladay:
> Does which triggers are ok to disable depend upon the specific migration or is it universal that some can be, all can be?

I am thinking that the specific migration matters, but only if someone decides they want to migrate a live production system. (In that case, transaction locks on tables in migration to prevent problems.) I don't know if migrating a live system is a good idea, but I can imagine someone deciding to do such a thing.

If migration is to be exclusively treated as an effective downtime process, then I do not see any problems with a universal trigger disabling process.

marcjohnson-kint:
> > Does which triggers are ok to disable depend upon the specific migration or is it universal that some can be, all can be?
> 
> I am thinking that the specific migration matters, but only if someone decides they want to migrate a live production system. (In that case, transaction locks on tables in migration to prevent problems.) I don't know if migrating a live system is a good idea, but I can imagine someone deciding to do such a thing.

At the moment, a production upgrade `should` be conducted by activating the new module version `in situ` meaning that these scripts will run on a production system.

> If migration is to be exclusively treated as an effective downtime process, then I do not see any problems with a universal trigger disabling process.

Yes, I think upgrades are likely to be done within planned downtime. That said, folks don't want that downtime to be excessively long

marcjohnson-kint:
@bohdan-suprun Please could you suggest a minimal potential template, if you have something in mind?

bohdan-suprun:
@marcjohnson-kint, I think it can be like:

```sql
START TRANSACTION;

UPDATE pg_trigger
  SET tgenabled = 'D'
WHERE tgrelid = '${myuniversity}_${mymodule}.item'::regclass::oid
  AND tgisinternal IS FALSE
  AND tgenabled = 'O';

-- Your update for entire table goes here.
-- Note: If the update deals with holdingsRecordId or (and) item level call numbers/locations than 
-- some of triggers have not to be disabled.

UPDATE pg_trigger
  SET tgenabled = 'O'
WHERE tgrelid = '${myuniversity}_${mymodule}.item'::regclass::oid
  AND tgisinternal IS FALSE
  AND tgenabled = 'D';

END TRANSACTION;
```

marcjohnson-kint:
@bohdan-suprun What does that statement in the template do? Does it disable all triggers for a given table?

bohdan-suprun:
@marcjohnson-kint, the first one disables all the triggers and the second one enables them again.

@kaladay, am I correct understand?

kaladay:
@bohdan-suprun Yes. The first one disables all triggers in the state **O**, which is ("_O = trigger fires in “origin” and “local” modes, D = trigger is disabled, R = trigger fires in “replica” mode, A = trigger fires always."_).

We may want to consider saving the state somehow before disabling to ensure no local changes as well as hande all the possibilities for  `pg_trigger.tgenabled`. Internal database triggers should be left alone and this is handled via `tgisinternal`.

marcjohnson-kint:
@kaladay @jeremythuff @bohdan-suprun @julianladisch What should we do with this pull request?

marcjohnson-kint:
> This can be merged so that we can point people to this example.

I like the idea of an example.

My concern with merging this is that it changes existing upgrade scripts that should already have been run, which might be misleading?

Does that make sense?

julianladisch:
If someone upgrades from a mod-inventory-storage version older than 19.0.0 then migrateItemCopyNumberToSingleValue.sql and populateDiscoverySuppressIfNotSet.sql run.
If someone upgrades from a mod-inventory-storage version older than 17.1.0 then all scripts of this pull request run.
This is correct.
Those versions no longer exists in production systems but people may want to experiment with older version, for example migrating from Daisy to Fameflower or Goldenrod.

PR #410, opened 2020-03-11
MODINVSTOR-453 Set default value to false for staffSuppress property

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINVSTOR-453: 
Set default value to false for staffSuppress property

Changes:
Set default value to false for instance.staffSuppress property;
Update API versions for instance-storage/batch/sync;
Add migration script to set staffSuppress=false if value for property is not set;

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStaffSuppressMigrationScriptTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStaffSuppressMigrationScriptTest.java
@@ -2,21 +2,20 @@ package org.folio.rest.api;
 
 import org.folio.rest.jaxrs.model.Instance;
 
-public class InstanceDiscoverySuppressMigrationScriptTest
+public class InstanceStaffSuppressMigrationScriptTest
   extends InstanceDefaultValueMigrationScriptTest {
-
-  private static final String DISCOVERY_SUPPRESS = "discoverySuppress";
+  private static final String STAFF_SUPPRESS = "staffSuppress";
   private static final String MIGRATION_SCRIPT
-    = loadScript("populateDiscoverySuppressIfNotSet.sql");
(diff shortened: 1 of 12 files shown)


Svitlana-Zmiivska:
@adamdickmeiss  Need that suppress from staff has always yes or no value. That why I've set default value for instance.staffSuppress=false
It is need to filter Instance records by Staff suppress: No [UIIN-968](https://issues.folio.org/browse/UIIN-968)

This implementation is based on the similar story [MODINVSTOR-449](https://issues.folio.org/browse/MODINVSTOR-449) where this approach is used

julianladisch:
@Svitlana-Zmiivska The front-end uses CQL `staffSuppress="false"` but it should use CQL `staffSuppress = "*" not staffSuppress = "true"` instead as noted on UIIN-968 and explained on https://github.com/folio-org/raml-module-builder#cql-matching-undefined-or-empty-values .
This will also show records where staffSuppress is not defined when the users clicks "Staff suppress [x] no" in the front-end.

src/test/java/org/folio/rest/api/InstanceStaffSuppressMigrationScriptTest.java
```diff
@@ -0,0 +1,25 @@
+package org.folio.rest.api;
+
+import org.folio.rest.jaxrs.model.Instance;
+
+public class InstanceStaffSuppressMigrationScriptTest
```
bohdan-suprun: Would it be better to have parameterized tests instead of building hierarchy of classes?

Svitlana-Zmiivska:
Thank you @adamdickmeiss @julianladisch for your reviews.
Done on UI side https://issues.folio.org/browse/UIIN-968

PR #415, opened 2020-03-16
MODINVSTOR-457: Handle holdings record not found case for item insert.

bohdan-suprun:
Resolves:
* https://issues.folio.org/browse/MODINVSTOR-457 - item creation hangs if it refers to a non-existent holdings record.
* https://issues.folio.org/browse/MODINVSTOR-378 - Use PgUtil.post util with INSERT ... RETURNING jsonb from RMB in ItemStorageAPI#postItemStorageItems

Changes:
* Return 422 error code when user specified non-existent holdings record id.
* Remove unnecessary get-by-id after inserting an item (it was a temporary fix for RMB-476, but it is now resolved).


src/main/java/org/folio/rest/exceptions/NotFoundException.java
```diff
@@ -0,0 +1,7 @@
+package org.folio.rest.exceptions;
+
+public class NotFoundException extends RuntimeException {
```
marcjohnson-kint: Why do we need specialised exceptions for these cases?

PR #416, opened 2020-03-17
MODINVSTOR-458: Calculate effective call number components for item batch create

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-458 - EffecticeCallNumberComponents is not set for items in "/item-storage/batch/synchronous" API.

### Changes:
* Add logic to calculate effective call number components for batch item create.
* Move effective call number logic to separate service class.

### DIFF
--- a/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
@@ -1,8 +1,6 @@
 package org.folio.rest.impl;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
-import static org.folio.rest.jaxrs.resource.ItemStorageBatchSynchronous.PostItemStorageBatchSynchronousResponse.respond422WithApplicationJson;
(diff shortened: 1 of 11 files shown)


src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
```diff
@@ -41,15 +51,27 @@ public void postItemStorageBatchSynchronous(ItemsPost entity, Map<String, String
...
+          final ValidationException validationError = (ValidationException) error;
+
+          asyncResultHandler.handle(Future.succeededFuture(
+            respond422WithApplicationJson(validationError.getErrors())));
+        } else {
+          asyncResultHandler.handle(Future.succeededFuture(
+            respond500WithTextPlain(error.getMessage())));
+        }
+
+        return null;
```
marcjohnson-kint: Is this likely to be a common behaviour? 

I think it might be. Even if it isn't, I think it could be worth extracting it to a method that describes the intent.

bohdan-suprun: Yes it is, moved to a separate method.

PR #417, opened 2020-03-18
MODINVSTOR-451 Remove old data after the whole Preceding/Succeeding titles migration

Svitlana-Zmiivska:
https://issues.folio.org/browse/MODINVSTOR-451

**Purpose:** 

Remove old data after the whole Preceding/Succeeding titles migration:

**Approach:**

Remove preceding/succeeding data from:

- reference and sample data
- instance_relationship and instance_relationship_type tables

### DIFF
--- a/sample-data/instance-relationships/american-bar-association-preceding-title.json
+++ b/sample-data/instance-relationships/american-bar-association-preceding-title.json
@@ -1,6 +1,0 @@
-{
-  "id": "1e772f36-20ac-4429-955b-8cd6c7dd644a",
-  "superInstanceId": "69640328-788e-43fc-9c3c-af39e243f3b7",
-  "subInstanceId":   "00f10ab9-d845-4334-92d2-ff55862bf4f9",
-  "instanceRelationshipTypeId": "cde80cc2-0c8b-4672-82d4-721e51dcb990"
-}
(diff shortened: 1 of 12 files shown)


sample-data/instance-relationships/american-bar-association-preceding-title.json
```diff
@@ -1,6 +0,0 @@
-{
```
bohdan-suprun: Could you please create the same relation but for the new model?

src/main/resources/templates/db_scripts/removeOldPrecedingSucceedingTitles.sql
```diff
@@ -0,0 +1,9 @@
+START TRANSACTION;
+
+DELETE FROM ${myuniversity}_${mymodule}.instance_relationship
+WHERE instanceRelationshipTypeId ='cde80cc2-0c8b-4672-82d4-721e51dcb990';
```
marcjohnson-kint: Are any existing records migrated to the new structure before they are deleted?

Svitlana-Zmiivska: The migration has already been done https://github.com/folio-org/mod-inventory-storage/pull/404

marcjohnson-kint: Ok, thank you

PR #420, opened 2020-03-23
MODINVSTOR-461: Add Withdrawn status to allowed item statuses list.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-461 - Introduce withdrawn item status.

Add `Withdrawn` status to allowed item statuses list.


marcjohnson-kint:
@bohdan-suprun That does mean that this needs to be amended to turn on the preview builds functionality.

Please can you do so and check that a version is built from this pull request (and running in the environment). Please ask the DevOps folks if any of the instructions are not clear (and let me know).

PR #421, opened 2020-03-25
MODINVSTOR-463: Smart index recreation from 19.0.*/19.1.0 to >= 19.1.1

julianladisch:


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -745,7 +745,7 @@
     {
       "run": "before",
       "snippetPath": "populateRmbInternalIndex.sql",
-      "fromModuleVersion": "19.0.0"
+      "fromModuleVersion": "19.1.1"
```
marcjohnson-kint: Is this pull request intended to be merged into master?

If so, it is my understanding that there won't be a version `19.1.1` in master, and so I'm not sure what this configuration would do? Would it apply from 19.2.0 instead?

marcjohnson-kint: > This means that it only runs when upgrading from a version before 19.1.1. This is what we want. It will run when upgrading from 19.1.0 to any version (for example to 19.2.0 or to 20.0.0).

Thanks, that makes sense.

To me, given that the next mainline version of mod-inventory-storage is likely to be `19.2.0` (or `20.0.0`) then this should be `runs when upgrading from any version prior to 19.2.0` rather than 19.1.1 

Does that make sense?

(This is why I find it challenging to back port changes that include database upgrades in them, because they are coupled to a module version history - rather than a sequence)

julianladisch: If someone upgrades from 19.1.1 to 20.0.0 we don't need to run this script because 19.1.1 already filled rmb_internal_index with index information. People may upgrade from a b19.1 patch version to a 20.*.* mainline version.

marcjohnson-kint: Thanks @julianladisch 

As agreed, will take this topic into a separate conversation.

PR #424, opened 2020-03-27
MODINVSTOR-459: Smart index recreation for 18.2.1-18.2.3 only

julianladisch:
18.0.0-18.2.0 require more index recreations than 18.2.1-18.2.3 because of
left(..., 600) index truncation. Therefore don't suppress index recreation
for versions before 18.2.1.


julianladisch:
When testing the smart index recreation for 19.1.1 yesterday I found that MODINVSTOR-459 using 18.*.* is not correct. This was fixed in 19.1.1 (supporting 18.2.3 only) and needs to be fixed in master as well (this pull request).

PR #425, opened 2020-03-30
Add `illPolicy` link-field annotation for mod-graphql

MikeTaylor:
Needed for ReShare issue PR-611.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -46,7 +46,7 @@
     },
     {
       "id": "holdings-storage",
-      "version": "4.1",
+      "version": "4.2",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/holdings-storage.raml
+++ b/ramls/holdings-storage.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Holdings Storage
-version: v4.1
+version: v4.2
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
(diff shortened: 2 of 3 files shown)


MikeTaylor:
@nielserik Can you imagine any sane reason why this would have built in Jenkins after I made the first commit, but not after I made the second?

MikeTaylor:
Or why the *continuous-integration/jenkins/branch* build would fail when the *continuous-integration/jenkins/pr-merge* build succeeds?!

MikeTaylor:
I am going to close and re-open this PR, on the offchange that this was a Jenkins weirdness — though the log does complain about `illPolicy`, which is the structure I added, so I think that is a forlorn hope.

MikeTaylor:
Oh, I see closing and re-opening doesn't help. All right, then I will push a trivial change.

MikeTaylor:
No, same thing: the **continuous-integration/jenkins/branch** build fails but the **continuous-integration/jenkins/pr-merge** build succeeds.

bohdan-suprun:
@MikeTaylor, there is already java class with `IllPolicy` name. You should either rename the new property (e.g. `illPolicyHoldings`) or add `javaType` attribute for the new property, e.g. `"javaType": "org.folio.rest.jaxrs.model.IllPolicyHoldingsRecord"`, so that RAML will generate java class for it with different name.

MikeTaylor:
Thank you @bohdan-suprun, very helpful comments. I'll get on them.

Do you have any idea why the **continuous-integration/jenkins/pr-merge** check is passing?

bohdan-suprun:
> Thank you @bohdan-suprun, very helpful comments. I'll get on them.
> 
> Do you have any idea why the **continuous-integration/jenkins/pr-merge** check is passing?

No, I don't. Looks confusing to me.

MikeTaylor:
@bohdan-suprun Please re-review!

PR #427, opened 2020-04-01
MODINVSTOR-470: Update updateItemStatusDate.sql migration version.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-470 - Function not being migrated between Edelweiss and Fameflower.

Update `updateItemStatusDate.sql` migration version to `19.2.0`.

### DIFF
--- a/src/test/java/org/folio/rest/api/UpdateItemStatusDateFunctionMigrationTest.java
+++ b/src/test/java/org/folio/rest/api/UpdateItemStatusDateFunctionMigrationTest.java
@@ -26,7 +26,7 @@ import junitparams.JUnitParamsRunner;
 import junitparams.Parameters;
 
 @RunWith(JUnitParamsRunner.class)
-public class UpdateItemStatusDateMigrationTest extends MigrationTestBase {
+public class UpdateItemStatusDateFunctionMigrationTest extends MigrationTestBase {
   @Parameters({
     "18.2.3, 19.1.1",
     "18.2.3, 19.2.0",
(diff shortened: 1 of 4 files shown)


marcjohnson-kint:
Please can a test be written to demonstrate that this migration works when the previous function was in place

bohdan-suprun:
@marcjohnson-kint, the test has been added. 

### Approach:
1. Create an 'old' function which appends 'OLD ' prefix to `item.status.date` property;
1. Create an item and update status for it;
1. Verify there is 'OLD' prefix for the `date` property;
1. Migrate tenant to new version by sending `POST /_/tenant` with `module_from` and `module_to` versions;
1. Create and update item status;
1. Verify `item.status.date` has **no** 'OLD' prefix and within 1 seconds before now.

src/test/java/org/folio/rest/api/UpdateItemStatusDateFunctionMigrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.junit.runner.RunWith;
+
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+
+@RunWith(JUnitParamsRunner.class)
+public class UpdateItemStatusDateFunctionMigrationTest extends MigrationTestBase {
+  @Parameters({
+    "18.2.3, 19.1.1",
```
marcjohnson-kint, line 31: How does this case work when the `fromVersion` in the `schema.json` states `19.2.0`?

bohdan-suprun: That's confusing me too. It seems that a script won't be executed only when it's `fromModuleVersion` is less then `fromVersion`, otherwise it will be, even if it belongs to a 'future' release.

@julianladisch, am I correct understand?

bohdan-suprun: As agreed, I've added explanation for that case.

PR #428, opened 2020-04-01
MODOAIPMH-105: soft delete for items, holdings and instances

mrdpopov:
# Purpose

Implement soft delete support for instances, holdings and items. 

# Approach 

Introduce views with the name of the tables that will filter records that have the deleted date populated. Rename instances, holdings and items tables with suffix _log, add a column to store the deleted date. Instead of actually deleting the row in the corresponding /delete API, the deleted_date column will be populated for that row(MODOAIPMH-113 to address that). This is the least intrusive way as almost no changes to the existing code are needed; additionally, the soft deleted records are transparently hidden from API clients.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -313,7 +313,7 @@
     },
     {
       "tableName": "instance",
-      "fromModuleVersion": "17.1.0",
+      "fromModuleVersion": "19.2.0",
       "withMetadata": true,
       "foreignKeys": [
         {
@@ -564,7 +564,7 @@
     },
     {
       "tableName": "holdings_record",
-      "fromModuleVersion": "19.0.0",
+      "fromModuleVersion": "19.2.0",
       "withMetadata": true,
       "foreignKeys": [
         {
@@ -637,7 +637,7 @@
     {
       "tableName": "item",
       "withMetadata": true,
-      "fromModuleVersion": "19.0.0",
+      "fromModuleVersion": "19.2.0",
       "foreignKeys": [
         {
           "fieldName": "holdingsRecordId",
@@ -849,6 +849,11 @@
     {
       "run": "after",
       "snippetPath": "removeOldPrecedingSucceedingTitles.sql",
+      "fromModuleVersion": "19.2.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "softdelete/softDeleteForHoldingsInstancesItems.sql",
       "fromModuleVersion": "19.2.0"
     }
   ]
--- a/src/main/resources/templates/db_scripts/softdelete/softDeleteForHoldingsInstancesItems.sql
+++ b/src/main/resources/templates/db_scripts/softdelete/softDeleteForHoldingsInstancesItems.sql
@@ -1,5 +1,5 @@
--- ADD deleted_date field to the tables
--- create indexes on deleted date
+-- add deleted_date field to the tables
+-- create indexees on deleted date
 -- rename instance, holdings and item tables
 -- create views with the name of the corresponding tables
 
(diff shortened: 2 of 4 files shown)


marcjohnson-kint:
@mrdpopov  The pull request preview functionality is being piloted by the Core Functional team at present.

It is not intended for general use at the moment.

mrdpopov:
@marcjohnson-kint Thank you for the notice, I have updated the description. I would like to test these changes on CI, as there are issues with testing it on a local environment. Once it's resolved, I will update the description of the PR with more details.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -374,6 +374,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "deleted_date",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
         }
```
marcjohnson-kint, line 383: As I understand it, this section of schema.json creates indexes on JSONB properties, it does not create them on columns. 

I believe `deleted_date` is a database column, and so I don't think this will work as expected.

@julianladisch @adamdickmeiss  @skoczko maybe you could confirm my understanding?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -313,7 +313,7 @@
     },
     {
       "tableName": "instance",
-      "fromModuleVersion": "17.1.0",
+      "fromModuleVersion": "19.2.0-SNAPSHOT",
```
marcjohnson-kint, line 316: As I understand it, there is no need to change the `fromModuleVersion` of a table when an index is added.

@bohdan-suprun I believe you recently confirmed this?

mrdpopov:
@marcjohnson-kint Thank you for your comments. Please note, that this PR is in a very early stage of development and has not been updated for quite some time.

src/main/resources/templates/db_scripts/schema.json
mrdpopov: Yes, you are right, this creates a trigger for a json field.

marcjohnson-kint:
> @marcjohnson-kint Thank you for your comments. 

Would you prefer I didn't comment further on this pull request?

> Please note, that this PR is in a very early stage of development and has not been updated for quite some time.

If this pull request isn't under active development, I suggest it be closed until the work is closer to being ready for review and integration.

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: Yes, correct.

src/main/resources/templates/db_scripts/softdelete/softDeleteForHoldingsInstancesItems.sql
```diff
@@ -0,0 +1,20 @@
...
+
+ALTER TABLE ${myuniversity}_${mymodule}.instance ADD "deleted_date" TIMESTAMP;
+ALTER TABLE ${myuniversity}_${mymodule}.item ADD "deleted_date" TIMESTAMP;
+ALTER TABLE ${myuniversity}_${mymodule}.holdings_record ADD "deleted_date" TIMESTAMP;
+--
+CREATE INDEX instance_deleted_date__index ON ${myuniversity}_${mymodule}.instance (deleted_date);
+CREATE INDEX item_deleted_date__index ON ${myuniversity}_${mymodule}.item (deleted_date);
+CREATE INDEX holdings_record_deleted_date__index ON ${myuniversity}_${mymodule}.holdings_record (deleted_date);
+--
+ALTER TABLE ${myuniversity}_${mymodule}.instance RENAME TO instance_log;
```
bohdan-suprun, line 14: It seems to me that after renaming the table we won't be able to add/remove an index/FK/etc to these tables via RMB standard capabilities (because table name in schema.json does not match actual table name).

PR #430, opened 2020-04-02
MODINVSTOR-473: Add indexes for *suppress properties

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-473 - Add b-tree indexes for staffSuppress and discoverySuppress properties.

### Changes:
* Add index for `instance.staffSuppress` and `instance.discoverySuppress` properties;
* Add index for `item.discoverySuppress` property;
* Add index for `holdingsRecord.discoverySuppress` property;

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -412,6 +412,18 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": true
+        },
+        {
+          "fieldName": "staffSuppress",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
+        },
+        {
+          "fieldName": "discoverySuppress",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
         }
       ],
       "fullTextIndex": [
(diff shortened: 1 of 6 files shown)


marcjohnson-kint:
Are there already tests for searching with these properties?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -608,6 +620,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
```
marcjohnson-kint, line 624: Does the `fromModuleVersion` need updating for indexes, or is there creation automatic irrespective of the `fromModuleVersion`?

bohdan-suprun: I think @julianladisch mentioned that indexes will be created always regardless of `fromModuleVersion`.

bohdan-suprun:
> Are there already tests for searching with these properties?

@marcjohnson-kint, I've added tests.

src/main/resources/templates/db_scripts/schema.json
julianladisch: Yes, we have automatic index recreation. For indexes `fromModuleVersion` is no longer needed.

PR #432, opened 2020-04-03
MODINVSTOR-474: Add "invalidIsbn" normalized index.

julianladisch:


bohdan-suprun:
Merging to unblock UI work.

PR #435, opened 2020-04-20
Add identifierTypeObject virtual field into instance.json

MikeTaylor:
Fixes MODGQL-123.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,9 +6,11 @@ buildMvn {
   mvnDeploy = 'yes'
   runLintRamlCop = 'yes'
   doKubeDeploy = true
+  publishPreview = false
 
   doDocker = {
     buildJavaDocker {
+      publishPreview = false
       publishMaster = 'yes'
       healthChk = 'yes'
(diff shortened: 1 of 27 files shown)


MikeTaylor:
@adamdickmeiss Is RMB now running in sufficiently deterministic a way that I can be confident these passed tests means my schema change is OK?

julianladisch:
RMB has a deterministic sort order fix for GenerateRunner:
https://github.com/folio-org/raml-module-builder/pull/628
This hasn't been released yet, though.

Using javaType should be sufficient to prevent a name clash.

MikeTaylor:
@julianladisch I know I have the option of forcing the issue with `javaType`, but I don't want to introduce that if it's not needed in this case. So what I really want at the moment is just for RMB to reliably tell me whether or not it _is_ needed.

MikeTaylor:
I am going to interpret the radio silence as agreement that this is good to merge. If the merge _does_ cause problems, that can be taken as an incentive to get https://github.com/folio-org/raml-module-builder/pull/628 deployed soonest, to prevent a recurrence!

PR #436, opened 2020-04-27
MODINVSTOR-483: Remove gen_random_uuid(), it fails in pgpool replication

julianladisch:
See https://issues.folio.org/browse/FOLIO-2447 : "Remove gen_random_uuid(), it fails in pgpool replication"

### DIFF
--- a/src/main/resources/templates/db_scripts/hridSettings.sql
+++ b/src/main/resources/templates/db_scripts/hridSettings.sql
@@ -2,19 +2,12 @@
 ALTER TABLE ${myuniversity}_${mymodule}.${table.tableName}
   ADD COLUMN IF NOT EXISTS
     lock boolean DEFAULT true UNIQUE CHECK(lock=true);
--- Derive id from tenant, module and table, setting UUID version to 4 and
--- UUID variant byte to 8.
--- Don't use pgcrypto gen_random_uuid() because pg-pool cannot replicate
--- random numbers: https://www.pgpool.net/docs/latest/en/html/restrictions.html
 INSERT INTO ${myuniversity}_${mymodule}.${table.tableName}
-  SELECT id, jsonb_build_object(
-    'id', id,
+  SELECT 'a501f2a8-5b31-48b2-874d-2191e48db8cd', jsonb_build_object(
     'instances', jsonb_build_object('prefix', 'in', 'startNumber', 1),
     'holdings', jsonb_build_object('prefix', 'ho', 'startNumber', 1),
     'items', jsonb_build_object('prefix', 'it', 'startNumber', 1)
   )
-  FROM (SELECT overlay(overlay(md5('${myuniversity}_${mymodule}.${table.tableName}')
-                 placing '4' from 13) placing '8' from 17)::uuid AS id) AS alias
   ON CONFLICT DO NOTHING;
 
 -- create initial sequences for HRID generation
--- a/src/main/resources/templates/db_scripts/hridSettings.sql
+++ b/src/main/resources/templates/db_scripts/hridSettings.sql
@@ -1,9 +1,11 @@
-CREATE EXTENSION IF NOT EXISTS "pgcrypto";
-
--- ensure that there will be one row only
+-- Ensure that there will be one row only.
 ALTER TABLE ${myuniversity}_${mymodule}.${table.tableName}
   ADD COLUMN IF NOT EXISTS
     lock boolean DEFAULT true UNIQUE CHECK(lock=true);
+-- Derive id from tenant, module and table, setting UUID version to 4 and
+-- UUID variant byte to 8.
+-- Don't use pgcrypto gen_random_uuid() because pg-pool cannot replicate
+-- random numbers: https://www.pgpool.net/docs/latest/en/html/restrictions.html
 INSERT INTO ${myuniversity}_${mymodule}.${table.tableName}
   SELECT id, jsonb_build_object(
     'id', id,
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/hridSettings.sql
```diff
@@ -1,17 +1,20 @@
-CREATE EXTENSION IF NOT EXISTS "pgcrypto";
-
--- ensure that there will be one row only
+-- Ensure that there will be one row only.
 ALTER TABLE ${myuniversity}_${mymodule}.${table.tableName}
   ADD COLUMN IF NOT EXISTS
     lock boolean DEFAULT true UNIQUE CHECK(lock=true);
+-- Derive id from tenant, module and table, setting UUID version to 4 and
+-- UUID variant byte to 8.
```
marcjohnson-kint, line 6: Does this mean that the database generated UUID is a custom design specific to FOLIO?

src/main/resources/templates/db_scripts/hridSettings.sql
```diff
@@ -1,17 +1,20 @@
...
 INSERT INTO ${myuniversity}_${mymodule}.${table.tableName}
   SELECT id, jsonb_build_object(
     'id', id,
     'instances', jsonb_build_object('prefix', 'in', 'startNumber', 1),
     'holdings', jsonb_build_object('prefix', 'ho', 'startNumber', 1),
     'items', jsonb_build_object('prefix', 'it', 'startNumber', 1)
   )
-  FROM (SELECT gen_random_uuid() AS id) AS alias
+  FROM (SELECT overlay(overlay(md5('${myuniversity}_${mymodule}.${table.tableName}')
+                 placing '4' from 13) placing '8' from 17)::uuid AS id) AS alias
```
marcjohnson-kint, line 17: Alternatively, could we remove the `id` column entirely?

@mreno-EBSCO Why does this configuration need a UUID `id` column?

mreno-EBSCO: I copied the pattern from another script, so I don't know if a UUID is required for this locking SQL to work. The pattern I copied it from generated invalid versioned UUIDs, so I changed it to the pgcrypto built in.

My problem with this is that I wasn't aware of a FOLIO requirement that module SQL must support pgpool nor can I find any indication it is a requirement in the wiki, dev site or github. All I see are vague references to pgpool here and there. I think that this requirement, if it is indeed a requirement, should be more explicit somewhere so that failure is not something that we have to wait until deployed to an environment that uses this software to discover. Then again, perhaps it is explicitly stated as a requirement somewhere that my searches have not found... 🤷‍♂️

src/main/resources/templates/db_scripts/hridSettings.sql
julianladisch: The pattern for microservices is that the client provides the id.
This doesn't work when a database trigger or some internal process creates new records. Then the database needs to come up with some id. One possibility is to use the md5 trick because a md5 hash and a uuid have the same number of bytes.
Another possibility is to simply use a hard-coded uuid in hridSetings.sql.

src/main/resources/templates/db_scripts/hridSettings.sql
julianladisch: Removing the `id` column entirely is possible but requires several code changes:
The table `hrid_settings` is generated by a table entry in schema.json and we use some RMB methods to access it.
Therefore I suggest to simply provide some id without using gen_random_uuid().

src/main/resources/templates/db_scripts/hridSettings.sql
marcjohnson-kint: @julianladisch 

> The pattern for microservices is that the client provides the id.

Unfortunately, that is not possible in this case because these records are required to be created without any involvement from a client.

> Another possibility is to simply use a hard-coded uuid in hridSetings.sql.

I really like this idea. I much prefer that to trying to invent our own UUID generation scheme.

src/main/resources/templates/db_scripts/hridSettings.sql
marcjohnson-kint: @mreno-EBSCO 

> My problem with this is that I wasn't aware of a FOLIO requirement that module SQL must support pgpool nor can I find any indication it is a requirement in the wiki, dev site or github.

You are not alone. As far as I'm aware, there is no official requirement for this.

> I think that this requirement, if it is indeed a requirement, should be more explicit somewhere

I totally agree

marcjohnson-kint: @julianladisch 

> Removing the id column entirely is possible but requires several code changes:
The table hrid_settings is generated by a table entry in schema.json and we use some RMB methods to access it.

That is a good point. I was merely coming up with ideas. I agree this is not viable.

> Therefore I suggest to simply provide some id without using gen_random_uuid().

I agree, and I like the suggestion you made in the other comment, to use hard coded generated UUIDs.

craigmcnally:
So if I understand correctly, this record is essentially system data that must always be there, right?  Acquisitions ran into a similar situation with batch-groups - we needed a "default" batch group record to always be present, even if the reference/sample data wasn't loaded.  

Here you handle this directly in the database/sql scripts.  Acquisitions took a slightly different approach.  The batch-group record was created and committed to source control just like reference/sample data, and the tenant API was updated to always load this record, regardless of whether the sample/reference data was being loaded.

As it turns out, the same solution of using a "hardcoded" UUID for the record was used in the acquisitions case as well.

I'm not saying one way is better than the other - but wanted to provide this as an alternative approach.  I think it might make sense to come up with a standard way of doing this for consistency across FOLIO.  This PR probably isn't the place for that discussion/debate though.

PR #440, opened 2020-05-01
MODINVSTOR-468: RMB upgrade for hitcount (totalRecords) RMB-591

MikhailFokanov:
https://issues.folio.org/browse/RMB-591 made these changes (https://github.com/folio-org/raml-module-builder/pull/634, https://github.com/folio-org/raml-module-builder/pull/637):
* Always return some estimated totalRecords value for optimized get sql: PgUtil.getWithOptimizedSql.
* No longer use SQL FUNCTION count_estimate_default. It may take long for full text queries when running `SELECT COUNT(*) ... LIMIT 1000`. Instead take the estimation of Postgres' EXPLAIN. Only if that estimation is < 4000 run `SELECT COUNT(*) ... LIMIT 1000`.
* If it is the last page (because > 0 but less than LIMIT records returned) we know for sure the accurate result count and return it as totalRecords.

This pull request uses RMB 29.5.0-SNAPSHOT in mod-inventory-storage master to test it in the folio-perf environment as discussed on https://issues.folio.org/browse/MODINVSTOR-468 .

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -88,7 +88,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <ramlfiles_path>${basedir}/ramls/</ramlfiles_path>
-    <raml-module-builder-version>29.4.0</raml-module-builder-version>
+    <raml-module-builder-version>29.5.0-SNAPSHOT</raml-module-builder-version>
     <generate_routing_context>/instance-storage/instances,/holdings-storage/holdings,/item-storage/items,/instance-bulk/ids</generate_routing_context>
     <argLine />
   </properties>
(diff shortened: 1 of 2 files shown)


MikhailFokanov:
It will be green after merge of https://github.com/folio-org/raml-module-builder/pull/637

pom.xml
```diff
@@ -88,7 +88,7 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <ramlfiles_path>${basedir}/ramls/</ramlfiles_path>
-    <raml-module-builder-version>29.4.0</raml-module-builder-version>
+    <raml-module-builder-version>29.5.0-SNAPSHOT</raml-module-builder-version>
```
marcjohnson-kint, line 91: @MikhailFokanov  @julianladisch  @adamdickmeiss Are we intending to run mod-inventory-storage master on a SNAPSHOT version of RAML Module Builder temporarily or permanently?

MikhailFokanov: This is temporary for testing. I will remove it after the build will be green. I should have posted a comment.

marcjohnson-kint: > I should have posted a comment.

That's ok.

> This is temporary for testing. I will remove it after the build will be green. 

After which build goes green, master or this pull request?

@skoczko mentioned a week or so ago that master would need to run a SNAPSHOT build temporarily in order for the performance tests to run against the new record count changes.

Is that no longer the plan?

julianladisch: I've merged this to master to run the performance test as discussed on https://issues.folio.org/browse/MODINVSTOR-468 .

PR #441, opened 2020-05-04
Modinvstor 487 update to vertx pg client

adamdickmeiss:
Updates for anticipated API of RMB with vertx-pg-client .. Work of RMB-246


src/test/java/org/folio/rest/api/AuditDeleteTest.java
```diff
@@ -127,26 +130,40 @@ private Object getRecordIdFromAuditTable(String tableName)
...
+          Object obj = row.getValue(i);
+          if (obj instanceof java.util.UUID) {
+            ar.add(obj.toString());
+          } else {
+            ar.add(obj);
+          }
+        }
+        list.add(ar);
+      }
+      result.complete(list);
```
marcjohnson-kint, line 156: Could this be extracted out into a separate method, both to explain what the intent is and to reduce the noise in the test?

src/test/java/org/folio/rest/api/MigrationTestBase.java
```diff
@@ -113,9 +116,24 @@ UpdateResult unsetJsonbProperty(String tableName, UUID id, String propertyName)
...
+          JsonArray ar = new JsonArray();
+          for (int i = 0; i < row.size(); i++) {
+            Object obj = row.getValue(i);
+            if (obj instanceof java.util.UUID) {
+              ar.add(obj.toString());
+            } else {
+              ar.add(obj);
+            }
+          }
+          list.add(ar);
```
marcjohnson-kint, line 134: Is this basically the same code as above?

src/test/java/org/folio/rest/api/MigrationTestBase.java
```diff
@@ -113,9 +116,24 @@ UpdateResult unsetJsonbProperty(String tableName, UUID id, String propertyName)
...
+          Row row = iterator.next();
+          JsonArray ar = new JsonArray();
+          for (int i = 0; i < row.size(); i++) {
+            Object obj = row.getValue(i);
+            if (obj instanceof java.util.UUID) {
+              ar.add(obj.toString());
+            } else {
+              ar.add(obj);
+            }
+          }
```
marcjohnson-kint, line 133: What is this intended to do?

What happens if there is more than one UUID value in a row, are both included in the list?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -1542,7 +1542,8 @@ public void cannotCreateAHoldingsWithHRIDFailure()
     final Response response = createCompleted.get(5, TimeUnit.SECONDS);
 
     assertThat(response.getStatusCode(), is(HttpURLConnection.HTTP_INTERNAL_ERROR));
-    assertThat(response.getBody(), isMaximumSequenceValueError("hrid_holdings_seq"));
+    // vertx-pg-client does not return code in message (ONLY message)
+    assertThat(response.getBody(), containsString("hrid_holdings_seq"));
```
marcjohnson-kint, line 1546: Does this mean that the only way to tell if this has happened now is by comparing the messages?

julianladisch: Yes.

src/test/java/org/folio/rest/api/AuditDeleteTest.java
adamdickmeiss: Yep. Will do

src/test/java/org/folio/rest/api/MigrationTestBase.java
adamdickmeiss: Most likely, will refactor

src/test/java/org/folio/rest/api/MigrationTestBase.java
adamdickmeiss: Probably bad things.. I guess the UUID will only bw at a specific column (which we are interested in).

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
marcjohnson-kint: @julianladisch Thanks

julianladisch: RMB fix is in code review: https://issues.folio.org/browse/RMB-622 = https://github.com/folio-org/raml-module-builder/pull/664

adamdickmeiss: And part of RMB 30.0.0 now

src/test/java/org/folio/rest/api/MigrationTestBase.java
adamdickmeiss: The issue is that the response is more typed than Json ... We have to manually map types to Srring.. Such as UUID in this case. The position in the array is "fixed" (columnwise).

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
adamdickmeiss: This is now "fixed" in PR https://github.com/folio-org/mod-inventory-storage/pull/445

src/test/java/org/folio/rest/api/MigrationTestBase.java
adamdickmeiss: Done

src/test/java/org/folio/rest/api/AuditDeleteTest.java
adamdickmeiss: Done

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -94,4 +100,28 @@ void assertGetNotFound(URL url) {
       throw new RuntimeException(e);
     }
   }
+
+  /**
+   * Map RowSet to JsonArray(s). One array per row.
+   * @param rowSet
+   * @return
+   */
+  List<JsonArray> rowSetToJsonArrays(RowSet<Row> rowSet) {
```
julianladisch, line 109: Can you add static?
Can you move the static method into a better place, for example test/java/org.folio.rest.support.db.RowSetUtil?

adamdickmeiss: ok

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
julianladisch: Can you make the change not here but in the isMaximumSequenceValueError method? If we get the SQLSTATUS code back we only need to update isMaximumSequenceValueError.

src/test/java/org/folio/rest/api/MigrationTestBase.java
marcjohnson-kint: What do you mean by the position in the array is fixed? The code above seems to interpret every index in the array, unless this has since changed.

PR #443, opened 2020-05-14
Modinvstor 479: Index - Filter Instance records by Date updated

jeremythuff:
Resolves [MODINVSTOR-479](https://issues.folio.org/browse/MODINVSTOR-479) by adding and index rule to the schema.json.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -428,6 +428,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "metadata.updatedDate",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
         }
       ],
       "fullTextIndex": [
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -564,9 +564,9 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
 
     JsonObject metadata = ir.getJson().getJsonObject("metadata");
 
-    String query = urlEncode(String.format("%s.updatedDate>=\"%s\"", METADATA_KEY, metadata.getString("updatedDate")));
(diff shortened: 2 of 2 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -428,6 +428,12 @@
           "tOps": "ADD",
           "caseSensitive": false,
           "removeAccents": false
+        },
+        {
+          "fieldName": "metadata.updatedDate",
```
bohdan-suprun, line 433: Hi @julianladisch, @adamdickmeiss what type of index has to be created for following CQL `metadata.updateDate >='2020-05-12'` (i.e. for `>=` operator)?

bohdan-suprun:
Hi @jeremythuff, could you please update the title of PR

src/main/resources/templates/db_scripts/schema.json
julianladisch: A b-tree index ("index" section). This is correct.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -541,6 +541,43 @@ public void canSearchUsingKeywordIndexAll()
...
+    client.get(instancesStorageUrl(String.format("?query=%s", query)), StorageTestSuite.TENANT_ID,
+        ResponseHandler.json(getCompleted));
+
+    Response response = getCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject responseBody = response.getJson();
+
+    JsonArray allInstances = responseBody.getJsonArray("instances");
+
+    assertThat(allInstances.size(), is(1));
```
marcjohnson-kint, line 578: Could two instances be created and filter so that only one of them is returned in the results?

With the way that the CQL searching works, if an index name is incorrect, it is effectively ignored, which I think means this test could easily produce a false positive (inclusion of the instance because it is the only one in the system)

jeremythuff: I believe this test is creating two instances and then only returning the second from the search. Is that not the case?

marcjohnson-kint: My apologies, indeed it is. Please could an assertion be added to check that the expected instance is included in the results

PR #444, opened 2020-05-14
API for OAI-PMH in inventory: Return instances, holdings and items fields by request.

mrdpopov:
# Purpose

Exposing API for OAI-PMH in inventory: Return instances, holdings and items fields by request.

# Approach

As the data returned may be rather large, selectStream() from PostgresClient has been used  to stream the data from the database using custom sql query. 

**Test creation is in progress.**
This resolves: MODINVSTOR-492


mrdpopov:
@adamdickmeiss Tests that call the new API have been added, but it seems that they are not counted towards the code coverage. Can you suggest how to resolve this?

PR #458, opened 2020-06-24
MODINVSTOR-499: Add Index for Instances Full-text Subjects Search

julianladisch:
The JSON array characters ["", ""] are ignored by full text search,
no need to strip them.

The generated query for subjects=a* uses the new index instance_subjects_idx_ft:

```
postgres=# explain analyze select count(*) from instance WHERE to_tsvector('simple', f_unaccent(instance.jsonb->>'subjects')) @@ tsquery_phrase(f_unaccent('a*'));
                                                               QUERY PLAN
-----------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=35.68..35.69 rows=1 width=8) (actual time=0.029..0.029 rows=1 loops=1)
   ->  Bitmap Heap Scan on instance  (cost=20.08..35.66 rows=10 width=0) (actual time=0.022..0.027 rows=5 loops=1)
         Recheck Cond: (to_tsvector('simple'::regconfig, f_unaccent((jsonb ->> 'subjects'::text))) @@ '''a'':*'::tsquery)
         Heap Blocks: exact=4
         ->  Bitmap Index Scan on instance_subjects_idx_ft  (cost=0.00..20.07 rows=10 width=0) (actual time=0.016..0.016 rows=5 loops=1)
               Index Cond: (to_tsvector('simple'::regconfig, f_unaccent((jsonb ->> 'subjects'::text))) @@ '''a'':*'::tsquery)
```

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -375,7 +375,7 @@
         },
         {
           "fieldName": "subjects",
-          "tOps": "ADD",
+          "tOps": "DELETE",
           "caseSensitive": false,
           "removeAccents": true
         }
@@ -466,6 +466,12 @@
           "fieldName": "isbn",
           "sqlExpression" : "normalize_isbns(jsonb->'identifiers')",
           "sqlExpressionQuery": "normalize_digits($)"
+        },
+        {
+          "fieldName": "subjects",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": true
         }
       ]
     },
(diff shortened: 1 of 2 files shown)


julianladisch:
No.
The outcome of such a test in mod-inventory-storage is the same regardless of the existence of a supporting index.
For performance tests we have https://github.com/folio-org/folio-perf-test

marcjohnson-kint:
> No.

Please could one be added.

> The outcome of such a test in mod-inventory-storage/src/test does not change when adding or removing a supporting index, the number of test records is too small to see any execution time difference.

That is true. I wasn't intending for the test to detect performance issues. 

Given the increased attention on specific searches that are done by the reference UI, I've started asked for a test that demonstrates that search (and the preferred CQL) to be added when addressing these kinds of concerns.

Does that make sense?

PR #464, opened 2020-07-02
MODINVSTOR-527: Update indexes and pmh view function to improve perfo…

mrdpopov:
## Purpose

Update indexes to improve performance of pmh view function


src/main/resources/templates/db_scripts/oaipmh/createOaiPmhViewFunction.sql
```diff
@@ -18,20 +18,22 @@ create or replace function ${myuniversity}_${mymodule}.getElectronicAccessName(v
 $$
 select jsonb_agg(distinct e)
```
julianladisch, line 19: schema.json says:
```
 {
      "run": "after",
      "fromModuleVersion": "19.2.0"
    },
    {
      "run": "after",
      "snippetPath": "removeOldPrecedingSucceedingTitles.sql",
      "fromModuleVersion": "19.2.0"
    }
```
These .sql files will not be executed when upgrading from 19.2.0 to the current version. Please update fromModuleVersion.

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
```diff
@@ -1,19 +1,18 @@
...
 
 drop function if exists ${myuniversity}_${mymodule}.pmh_view_function(timestamptz, timestamptz, bool, bool);
-
 drop function if exists ${myuniversity}_${mymodule}.strToTimestamp(text);
-
 drop function if exists ${myuniversity}_${mymodule}.dateOrMin(timestamptz);
-
 drop function if exists ${myuniversity}_${mymodule}.dateOrMax(timestamptz);
-
 drop function if exists ${myuniversity}_${mymodule}.getElectronicAccessName(val jsonb);
```
julianladisch, line 18: createOaiPmhViewFunction.sql uses `create or replace function` therefore `drop function` is not needed.
If it were needed (for example when argument types have changed) it should be moved immediately before the `create or replace function` in createOaiPmhViewFunction.sql for readability.

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
```diff
@@ -1,19 +1,18 @@
...
+drop index if exists ${myuniversity}_${mymodule}.audit_instance_createddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.audit_holdings_record_createddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.audit_item_createddate_idx;
+
+drop index if exists ${myuniversity}_${mymodule}.instance_pmh_metadata_updateddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.item_pmh_metadata_updateddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.holdings_record_pmh_metadata_updateddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.audit_instance_pmh_createddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.audit_holdings_record_pmh_createddate_idx;
+drop index if exists ${myuniversity}_${mymodule}.audit_item_pmh_createddate_idx;
```
julianladisch, line 12: Can you move the `drop index` line to the corresponding `create index` line in createOaiPmhViewFunction.sql?

For example:
```
drop index if exists ${myuniversity}_${mymodule}.item_metadata_updateddate_idx;
drop index if exists ${myuniversity}_${mymodule}.item_pmh_metadata_updateddate_idx;
create index if not exists item_pmh_metadata_updateddate_idx on ...

drop index if exists ${myuniversity}_${mymodule}.holdings_record_metadata_updateddate_idx;
drop index if exists ${myuniversity}_${mymodule}.holdings_record_pmh_metadata_updateddate_idx;
create index if not exists holdings_record_pmh_metadata_updateddate_idx on ...
```
Grouping `drop index` and `create index` increases readability.

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
mrdpopov: Do you think revert_createOaiPmhViewFunction.sql  is not needed at all if create or replace is used in our indexes and function?

src/main/resources/templates/db_scripts/oaipmh/createOaiPmhViewFunction.sql
mrdpopov: This will be release in the next version of mod-inventory-storage, 19.2.4, is it ok to set 12.2.4 in from module version?

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
mrdpopov: I agree, but indexes depend of functions, so they can't be deleted before functions are deleted. That's why create and drop are separated.

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
mrdpopov: Revert script has been removed.

src/main/resources/templates/db_scripts/oaipmh/revert_createOaiPmhViewFunction.sql
mrdpopov: Done.

src/main/resources/templates/db_scripts/oaipmh/createOaiPmhViewFunction.sql
marcjohnson-kint: @mrdpopov Personally, I think the `fromVersion` in the mainline should be `19.3.0` and `19.2.4` in the bug fix back port.

@julianladisch is that a sensible approach?

marcjohnson-kint: > These .sql files will not be executed when upgrading from 19.2.0 to the current version. Please update fromModuleVersion.

Is this because `fromModuleVersion` in this case means apply this script when the module is upgraded from something before this version to either that version or newer? 

And hence, an upgrade from 19.2.x to a version with this code in would not execute this script?

mrdpopov: I cant upgrade to the unreleased 19.0.4 now

mrdpopov:
@julianladisch @marcjohnson-kint Do you think this is ready to be merged?

marcjohnson-kint:
> @julianladisch @marcjohnson-kint Do you think this is ready to be merged? I have put 19.2.2 in the frommoduleversion. Will this trigger schema upgrade?

I don't think 19.2.2 makes sense, because that version has already been released. I think `fromModuleVersion` doesn't mean apply this change when coming from this version or earlier. I think it means apply this change when going to this version or later.

I think we need @julianladisch advice on from module version. As my questions suggest, I don't understand this area enough to know what the right value is, I'm only reflecting my understanding.

src/main/resources/templates/db_scripts/oaipmh/createOaiPmhViewFunction.sql
julianladisch: Using 19.2.4 in both mainline and bug fix is correct.
When upgrading from 19.2.3 to 19.3.1 it should run.
When upgrading from 19.2.4 to 19.3.0 it should not run again because it has run in 19.2.4.

marcjohnson-kint: > Using 19.2.4 in both mainline and bug fix is correct.

Why would we want a version referred to on the mainline that should never be released from the mainline?

Is this to avoid it being executed twice if the bug fix release has previously been installed?

julianladisch: > It looks like I cant put the unreleased 19.2.4 now in frommoduleversion because the build will fail.

Branch MODINVSTOR-527 with
```
     {
       "run": "after",
      "fromModuleVersion": "19.2.4"
     },
```
successfully builds.

julianladisch: >Is this to avoid it being executed twice if the bug fix release has previously been installed?

Yes.

marcjohnson-kint: Thanks @julianladisch for explaining this

PR #472, opened 2020-07-15
MODINVSTOR-536: Separate filtering from response generation in oai-pmh view

oleksandr-yatsenko1:
Updated RAML description and implementation for intances view. Now it consists from two routines: 
1) "updated instances view" and 2) "instances view" itself. The order of invocatrion is the same as mentioned above.


oleksandr-yatsenko1:
Pull request will be abandoned due unneeded commits inclusion.

PR #473, opened 2020-07-15
MODINVSTOR-536: Separate filtering from response generation in oai-pmh view

oleksandr-yatsenko1:
MODINVSTOR-536: Separate filtering from response generation in oai-pmh view


mrdpopov:
Please upgrade version in shema.json, something like this:
    {
      "run": "after",
      "fromModuleVersion": "19.2.5"
    },

oleksandr-yatsenko1:
> Please upgrade version in shema.json, something like this:
> {
> "run": "after",
> "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
> "fromModuleVersion": "19.2.5"
> },

Done

ramls/oai-pmh-view.raml
```diff
@@ -20,14 +21,14 @@ resourceTypes:
   collection-get: !include raml-util/rtypes/collection-get.raml
 
 /oai-pmh-view:
-  /instances:
-    displayName: Stream API to get data for OAI-PMH
+  /updatedInstanceIds:
+    displayName: Stream API to get instances ids for OAI-PMH
```
marcjohnson-kint, line 25: Was it considered that the existing API could have remained in place, and the updated ID specific variant added as a new endpoint, in order to avoid the break in compatibility?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1057,10 +1057,14 @@
     },
     {
       "id": "oaipmhview",
-      "version": "1.0",
+      "version": "2.0",
```
marcjohnson-kint, line 1060: Why is this change compatibility breaking? Is it because the `oai-pmh-view/instances` endpoint has become a POST rather than a GET?

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -80,6 +77,35 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
     }
   }
 
+  @Override
+  public void getOaiPmhViewUpdatedInstanceIds(String
+                                                startDate, String endDate, boolean deletedRecordSupport,
+                                              boolean skipSuppressedFromDiscoveryRecords, @Pattern(regexp = "[a-zA-Z]{2}") String lang, RoutingContext routingContext,
+                                              Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
marcjohnson-kint, line 84: Please use the standard line continuation and 2 space indentation

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -80,6 +77,35 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
...
+      .params()));
+
+    Tuple params = createPostgresParams(startDate, endDate, deletedRecordSupport, skipSuppressedFromDiscoveryRecords);
+    log.debug("postgres params:", params);
+    getCommonInstanceProcessing(SQL_UPDATED_INSTANCES, params, routingContext, okapiHeaders, asyncResultHandler, vertxContext,
+      "Select from oai pmh updated instances view completed successfully");
+  }
+
+  @Override
+  public void postOaiPmhViewInstances(boolean skipSuppressedFromDiscoveryRecords, List<String> entity, RoutingContext routingContext, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
```
marcjohnson-kint, line 96: Please apply standard line continuation and 2 space indentation

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -93,7 +119,7 @@ private void respondWithError(Throwable t, Handler<AsyncResult<Response>> asyncR
   }
 
   private Tuple createPostgresParams(String startDate, String endDate, boolean deletedRecordSupport,
-      boolean skipSuppressedFromDiscoveryRecords) {
+                                     boolean skipSuppressedFromDiscoveryRecords) {
```
marcjohnson-kint, line 122: Please apply standard line continuation and 2 space indentation

Horizontal alignment is harder to maintain and uses more valuable horizontal space

descriptors/ModuleDescriptor-template.json
mrdpopov: Because the old endpoint is working differently in this version of API

oleksandr-yatsenko1: I fixed version to emphasize that changes made just extend existing particular API related to instances retrievement.

marcjohnson-kint: @oleksandr-yatsenko1 What do you mean you fixed the version? 

Unless, the existing endpoint is to be maintained, as @mrdpopov suggests, then it is a compatibility breaking change.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: Done

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: Done

descriptors/ModuleDescriptor-template.json
mrdpopov: The existing endpoint does not need to be maintained, au contraire it needs to be changed. It is only used in mod-oai-pmh.

marcjohnson-kint: @mrdpopov Indeed, and given it is changing in an incompatible way, it is a breaking compatibility change. I only wanted to confirm my understanding. Thank you for doing so.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -922,7 +922,7 @@
     {
       "run": "after",
       "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
-      "fromModuleVersion": "19.2.4"
+      "fromModuleVersion": "19.2.5"
```
bohdan-suprun, line 925: Should it be 19.3.0?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -118,54 +137,75 @@ public void testDeletedRecordSupport() throws InterruptedException, TimeoutExcep
...
     assertThat(data.size(), is(0));
   }
 
+  private List<JsonObject> getOiaPmhViewInstances(Map<String, String> queryParams) throws InterruptedException, ExecutionException, TimeoutException {
+    // Get updated instances ids
+    List<JsonObject> updatedInstanceData = requestOaiPmhViewUpdatedInstanceIds(queryParams);
+
+    // Extract instances ids
+    UUID[] instanceIds = updatedInstanceData.stream()
+      .map(e -> UUID.fromString(e.getString("instanceid")))
```
bohdan-suprun, line 157:
```suggestion
      .map(json -> UUID.fromString(json.getString("instanceid")))
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -1057,10 +1057,14 @@
...
       "handlers" : [
         {
           "methods" : [ "GET" ],
+          "pathPattern" : "/oai-pmh-view/updatedInstanceIds",
+          "permissionsRequired": ["inventory-storage.oai-pmh-view.updatedinstanceids.collection.get"]
+        }, {
+          "methods" : [ "POST" ],
```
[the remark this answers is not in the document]
mrdpopov: It accepts list of params in body

ramls/oai-pmh-view.raml
```diff
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Inventory Storage OAI-PMH view API
-version: v1.0
+version: v1.1
```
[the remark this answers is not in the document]
mrdpopov: to 2.0 version

ramls/oai-pmh-view.raml
```diff
@@ -49,4 +50,25 @@ resourceTypes:
...
+      body:
+        application/json:
+          description: List of instances Ids
+          type: string[]
+      queryParameters:
+        skipSuppressedFromDiscoveryRecords:
+          description: skip discovery suppressed records
+          type: boolean
+          required: false
+          default: true
```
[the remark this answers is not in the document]
mrdpopov: There should be one get, one post

ramls/oaipmh/oaipmhinstanceids.json
```diff
@@ -0,0 +1,25 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Response schema for oai-pmh updated instances ids view",
+  "type": "object",
+  "properties": {
+    "instanceId": {
+      "description": "OAI-PMH updated instances ids",
+      "type": "string"
+    },
+    "instanceUpdatedDate": {
+      "description": "The last updated date or deleted date of an instance or it's items and holdings",
```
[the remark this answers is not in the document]
mrdpopov: According to the business logic, we call those updateddate

src/main/resources/templates/db_scripts/schema.json
mrdpopov: Maybe, we need to make a new release, but I doubt it will be 19.3.0

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
bohdan-suprun: What I'm trying to say is that the POST in this case behaves as a GET method - used to read records by some criteria, right?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: hmm, but the PR is created to `master`, and it is already of 19.3.0 version.

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
mrdpopov: Yes, it's behaving like get. IT's made post because it accepts list of ids in the body.

src/main/resources/templates/db_scripts/schema.json
mrdpopov: I think it will be a hotfix release, nevertheless this change is needed to be in master too.

marcjohnson-kint: There was a recent workaround where a bug fix version was used in a `fromModuleVersion` in master in order to avoid the same script being executed twice if the tenant had already upgraded to a bug fix version already containing the change.

Personally, I don't think this is a good regular pattern, I think the mainline code should refer to versions that are / could be present on the mainline.

Given that this change breaks compatibility, the module version should be changed to 20.0.0 and I think that is what should be present here as well.

If this breaking change is to be back ported (which I strongly disagree with, however I'm aware that exceptional permission has been granted for this) then I think the `fromModuleVersion` should be changed in the bug fix branch.

That said, if this script is destructive or non-repeatable, and folks want to apply the same workaround, then this version should be the version *after* the bug fix version that contained it. 

@julianladisch Is that a correct understand of how these versions work and how the workaround works?

marcjohnson-kint:
Please change the module version to 20.0.0 to indicate that there has been a compatibility breaking change.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1057,10 +1057,14 @@
...
           "pathPattern" : "/oai-pmh-view/instances",
           "permissionsRequired": ["inventory-storage.oai-pmh-view.instances.collection.get"]
```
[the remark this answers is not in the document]
marcjohnson-kint: Please remember to change the contents of the all permission set when this change is made.

The permissions granted to the OAI-PMH module will also likely need to change as well.

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
oleksandr-yatsenko1: Done

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: Done

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
oleksandr-yatsenko1: Because it can carry list of instance's ids and their amout can me signoificant, that's why POST method was chosen. This will avoid facing limitation of query params in case with GET method.

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
oleksandr-yatsenko1: I have fixed permissions in both "mod-inventory-storage" and "mod-oai-pmh" modules.

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
oleksandr-yatsenko1: > This suggests that you're going to supply filter params as a POST body? But in a fact it is GET method?

Do you mean that despite the fact that this is a POST entry it uses "query parameter"? Of cource we can  pass it in body's payload. But for sake of simplyfiing payload processing we use single query parameter here. Does it make sence?

ramls/oaipmh/oaipmhinstanceids.json
```diff
@@ -0,0 +1,25 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Response schema for oai-pmh updated instances ids view",
+  "type": "object",
+  "properties": {
+    "instanceId": {
+      "description": "OAI-PMH updated instances ids",
+      "type": "string"
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Done

ramls/oaipmh/oaipmhinstanceids.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "instanceUpdatedDate": {
+      "description": "The last updated date or deleted date of an instance or it's items and holdings",
+      "type": "string",
+      "format": "date-time"
+    },
+    "suppressDiscovery": {
+        "description": "Indicates if instance should be suppressed from discovery",
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Good point, thanks.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -64,7 +62,7 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
             respondWithError(completed.cause(), asyncResultHandler);
             return;
           }
-          log.info("Select from oai pmh view completed successfully");
+          log.info(logMessage);
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: We don't have much of console output messages and this one is just refactored from former code. So if we need to desrease amount of logging possibly this can be performed over the whole module? How do you think?

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -119,4 +145,22 @@ private Tuple createPostgresParams(String startDate, String endDate, boolean del
...
+      if (ArrayUtils.isNotEmpty(instancesIds)) {
+        tuple.addUUIDArray(instancesIds);
+      } else {
+        tuple.addValue(new String[0]);
+      }
+
+      tuple.addBoolean(skipSuppressedFromDiscoveryRecords);
+
+    } catch (Exception e) {
+      throw new IllegalArgumentException(e);
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Not sure I understant what do you mean? I'll slightly update this part of code though.

src/main/resources/templates/db_scripts/schema.json
oleksandr-yatsenko1: As I understand correctly this is the question of duscussion now which version to leave.

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
oleksandr-yatsenko1: Done

oleksandr-yatsenko1:
> Please change the module version to 20.0.0 to indicate that there has been a compatibility breaking change.

Sure, I will fix this. Thank you for clarifying, Marc.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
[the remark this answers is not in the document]
bohdan-suprun: I mean that the exception is wrapped and re-thrown as unchecked exception, however this unchecked exception seems newer caught by calling code.

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
bohdan-suprun: > Yes, it's behaving like get. IT's made post because it accepts list of ids in the body.
If it was agreed with an SA I'm fine as well.

> But for sake of simplyfiing payload processing we use single query parameter here. Does it make sence?

I'd prefer specifying everything as either query params or as payload.

ramls/oaipmh/oaipmhinstanceids.json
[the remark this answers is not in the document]
bohdan-suprun: Ah, good point. Do you mean that this schema describes response only?

Usually we specify format to better explain API, so API consumers now what data format is expected for a property.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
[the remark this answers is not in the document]
mrdpopov: The "old" api method was wreapped in try/catch, please add same to the new api methods

ramls/oai-pmh-view.raml
```diff
@@ -49,4 +50,25 @@ resourceTypes:
...
+    post:
+      description: Stream instances view data for oai-pmh
+      is:
+        [validate]
+      body:
+        application/json:
+          description: List of instances Ids
+          type: string[]
+      queryParameters:
+        skipSuppressedFromDiscoveryRecords:
```
marcjohnson-kint, line 64: Why is this needed in the POST to fetch the specific records? Would the initial query (that determines the instances and remembers the IDs) reflect whether to include or exclude instances that are suppressed from discovery?

ramls/oaipmh/oaipmhinstanceids.json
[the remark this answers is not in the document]
marcjohnson-kint: Indeed, the primary intent of the API definition is to describe to the potential clients. That FOLIO sometimes also uses it to drive behaviour is a secondary concern from my perspective.

descriptors/ModuleDescriptor-template.json
```diff
@@ -1057,12 +1057,21 @@
...
+        },
+        {
+          "methods" : [ "GET" ],
+          "pathPattern" : "/oai-pmh-view/updatedInstanceIds",
+          "permissionsRequired": ["inventory-storage.oai-pmh-view.updatedinstanceids.collection.get"]
+        }, {
+          "methods" : [ "POST" ],
+          "pathPattern" : "/oai-pmh-view/enrichedInstances",
```
marcjohnson-kint, line 1073: What are enriched instances?

ramls/oai-pmh-view.raml
```diff
@@ -49,4 +51,53 @@ resourceTypes:
...
+  /enrichedInstances:
+    displayName: Stream API to get instances with items and holdings for OAI-PMH
+    post:
+      description: Stream instances view data for oai-pmh
+      is:
+        [validate]
+      body:
+        application/json:
+          description: List of instances Ids
+          type: string[]
```
marcjohnson-kint, line 92: Usually a schema is used to define a JSON request or response. What does `string[]` mean in the context of a JSON response?

ramls/oai-pmh-view.raml
```diff
@@ -10,6 +10,8 @@ documentation:
 
 types:
   oaipmhInstances: !include oaipmh/oaipmhinstances.json
+  oaipmhUpdatedInstanceIds: !include oaipmh/oaipmhupdatedinstanceids.json
+  oaipmhEnrichedInstances: !include oaipmh/oaipmhenrichedinstances.json
```
marcjohnson-kint, line 14: This type does not appear to be referenced anywhere in the schema

ramls/oaipmh/oaipmhupdatedinstanceids.json
```diff
@@ -0,0 +1,25 @@
...
+    "instanceId": {
+      "description": "OAI-PMH updated instances ids",
+      "type": "string"
+    },
+    "instanceUpdatedDate": {
+      "description": "The last updated date or deleted date of an instance or it's items and holdings",
+      "type": "string",
+      "format": "date-time"
+    },
+    "suppressDiscovery": {
```
marcjohnson-kint, line 15: Similar to one of @bohdan-suprun earlier comments, could this be `suppressFromDiscovery` instead, or is it like this for consistency with something else?

descriptors/ModuleDescriptor-template.json
oleksandr-yatsenko1: It just contains instance data (items with holdings as in former and currently retained API entry, for backward compatibility purposes) but without some atttributes on the top level (like "lastUpdatedDate" and "deleted"). Indeed possibly it can be named differently, but I belive to so as not not confuse with former get@"/oai-pmh-view/instances" possibly it would be named differently. Or maye taking into account another method the name can be retaned - post@"/oai-pmh-view/instances". How do you think, which approach will be better?

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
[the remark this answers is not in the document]
oleksandr-yatsenko1: This is legacy part of code, I just eliminated code duplication. I'm not sure if this particular part of logging is important or not, but it looks it was there before even for production. So how do you think, do we need to change the level of its logging still?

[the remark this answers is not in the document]
bohdan-suprun: yes, please change to `debug` level

ramls/oai-pmh-view.raml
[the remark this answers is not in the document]
oleksandr-yatsenko1: As for parameter passing here, I removed single query parameter and moved it into body payload. Because there is particular data type was introduced in RAML, which carry all passed data ("instances ids" and "suppressed from discovery" flag). Thank you for raising this qurestion.

ramls/oaipmh/oaipmhinstanceids.json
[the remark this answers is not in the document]
oleksandr-yatsenko1: In the last changes I added for all newly introduced schemas reference to uuid defirnition. But taking into account that uuid definition located in different path and to allow relative reference I was updated uuid definiion itself in term of valid schema definition. Now it can be referenced from everywhere by relative path specifying.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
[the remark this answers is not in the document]
oleksandr-yatsenko1: I changed it to debug level. Thank you for noticing.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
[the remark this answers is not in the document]
oleksandr-yatsenko1: I fixed this by moving parameter preparation into sage scope of code. Thank you for noticing.

src/main/resources/templates/db_scripts/schema.json
oleksandr-yatsenko1: I left the version 19.2.5 because latest changes made with mind of backward compatible. To all change shouldn't lead to breakdown of former functionality.

oleksandr-yatsenko1:
> Please change the module version to 20.0.0 to indicate that there has been a compatibility breaking change.

As I wrote above, latesr changes are backward compatible, so it look like we can retain current 19th major version number.

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: I havce checked and added a new type definitoon for POST "instances ids" with "skip suppressed from discovery" flag. Currently it seems like all the defined types are used.

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: You are right that skipSuppressedFromDiscoveryRecords from the initial request (updatedInstanceIds) will be used to determine what ids should be passed, but in this query we need to know what items and holdings should be included in the response (based on skipSuppressedFromDiscoveryRecords ). The latest changes passes "skipSuppressedFromDiscovery" in couple with "instance ids" in POST's body payload. And as it was mentioned above it just affect if items/holdings will be excluded from result depending on their "discoverySuppress" value. And such a flag affect th whole bunch of instance ids.

ramls/oaipmh/oaipmhinstanceids.json
```diff
@@ -0,0 +1,25 @@
...
+    "instanceId": {
+      "description": "OAI-PMH updated instances ids",
+      "type": "string"
+    },
+    "instanceUpdatedDate": {
+      "description": "The last updated date or deleted date of an instance or it's items and holdings",
+      "type": "string",
+      "format": "date-time"
+    },
+    "suppressDiscovery": {
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Initially that name "suppressDiscovery" came from data view itself. I mean it was part of json field "itemsAndHoldingsFields". Taking into  account backward compatibility approach we  adhere to, we tried to made as less chages as possible. That's why legacy attributes used too. If you think that this name is confusing of cource we can change it. How pdo you think, is it important to do so?

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: I introduced a new type for POST payload which carry bith "instance ids" and "skipSuppressedFromDiscovery" data. So now there is only payload passed with parameters.

ramls/oaipmh/oaipmhupdatedinstanceids.json
oleksandr-yatsenko1: As I mentioned above this name came from data view and for the sake of minimizing any changes this name was retained. If it looks confusing I think we can fix this. So the last word is for you, let me know about your final decision and I will apply it.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: If the name is intended to keep consistency with existing decisions, I suggest it be left as is.

ramls/oai-pmh-view.raml
marcjohnson-kint: Please could the descriptions of both properties make the distinction explicit that this only applies to the holdings and items that are fetched with the instance, rather than the instance itself.

ramls/oaipmh/oaipmhupdatedinstanceids.json
marcjohnson-kint: I think if it is named this way for consistency that is ok. I would have preferred the name to read better, that opportunity is likely now lost to us for the time being.

ramls/oai-pmh-view.raml
```diff
@@ -49,4 +52,48 @@ resourceTypes:
...
+      description: Stream updated instances ids for oai-pmh
+      is:
+        [validate]
+      queryParameters:
+        startDate:
+          description: start date
+          type: string
+          required: false
+        endDate:
+          description: start date
```
julianladisch, line 71: end date

julianladisch, line 71: Can you extend the description of the startDate and endDate parameters to mention what they do (filter instances? filter holdings? filter items?) and which field is used for filtering (metadata.createdDate?, metadata.updatedDate?)? What date formats are accepted (2020-01-01, 2020-01-01T00:00:00, ...)? Does endDate=2020-01-01 means the same as endDate=2020-01-01T23:59:59.999 and includes 2020-01-01T12:00:00?

ramls/oaipmh/oaipmhupdatedinstanceids.json
```diff
@@ -0,0 +1,26 @@
...
+  "description": "Response schema for oai-pmh updated instances ids view",
+  "type": "object",
+  "properties": {
+    "instanceId": {
+      "description": "OAI-PMH updated instances ids",
+      "type": "string",
+      "$ref": "../uuid.json"
+    },
+    "instanceUpdatedDate": {
+      "description": "The last updated date or deleted date of an instance or it's items and holdings",
```
julianladisch, line 12: The property name "instanceUpdatedDate" is misleading when the description says that it may be the updated date of an item.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -80,6 +94,28 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
     }
   }
 
+  @Override
+  public void postOaiPmhViewEnrichedInstances(OaipmhInstanceIds entity, RoutingContext routingContext,
```
julianladisch, line 98: `@Validate` is missing. The .raml file says `is: [validate]`.

ramls/oaipmh/oaipmhupdatedinstanceids.json
oleksandr-yatsenko1: I will fix it. Thank you for noticing.

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: Done

src/main/resources/templates/db_scripts/schema.json
julianladisch: > hmm, but the PR is created to `master`, and it is already of 19.3.0 version.

master is on 19.3.0-SNAPSHOT.

FOLIO by default uses the next feature version as the *-SNAPSHOT development version for master: https://dev.folio.org/guidelines/release-procedures/#prepare-and-perform-the-source-release

When doing the next release the maintainer needs to adjust the version: If there are only bugfixes use v19.2.5. If there are new features use v19.3.0. If there are breaking changes use v20.0.0.

The 19.3.0-SNAPSHOT on master does not reflect whether the commits that were made after the last release are bugfixes, features or breaking changes. This is postponed until the next version gets actually released.

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: Done

ramls/oaipmh/oaipmhupdatedinstanceids.json
oleksandr-yatsenko1: How do you think, will the name "oaiRecordUpdatedDate" be more appropriate for this field?

src/main/resources/templates/db_scripts/schema.json
julianladisch: `fromModuleVersion` is the version where the script has been introduced: https://github.com/folio-org/raml-module-builder#the-post-tenant-api
19.2.5 is correct if we release this pull request as 19.2.5.

marcjohnson-kint: > The 19.3.0-SNAPSHOT on master does not reflect whether the commits that were made after the last release are bugfixes, features or breaking changes. This is postponed until the next version gets actually released.

As the next mainline version is always intended to be the next minor version, I tend to consider compatible feature changes and bug fixes as not affecting the version. 

I tend to try to identify compatibility breaking changes during the pull request and change the next version appropriately within that pull request. Otherwise what happens is that SNAPSHOT versions of a module version are published that do not reflect the intent of the changes.

This is less problematic for modules, mostly only misleading, it can be significant for libraries.

As a lead maintainer, I want the defined version to be accurate to what should be next released when I come to make the formal release. Otherwise, I have to know enough about all of the changes in order to evaluate that myself, which can be difficult and time consuming when that is 3 months of changes made by multiple teams autonomously.

I do tend to do some checks to spot unacknowledged compatibility breaking changes. It isn't unusual for those to have been made without the version changing.

julianladisch: This pull request is no longer a breaking change. But it adds two new API endpoints: `oai-pmh-view/updatedInstanceIds`, `/oai-pmh-view/enrichedInstances`. This indicates that the release version should be 19.3.0 to signal that it is a feature version. I understand the hot-fix releases should be restricted to bug fixes, but this pull request qualifies as an exceptions to this rule because it has very low risk.

marcjohnson-kint: > 19.2.5 is correct if we release this pull request as 19.2.5.

This pull request is intended to be merged to the mainline. To me, that means that this change is actively intended for 19.3.0.

If folks decide to also port these changes to 19.2.5 then that will be a separate pull request, with different intentions. (I won't repeat my concerns about that approach here)

This is why I personally think the `fromVersion` in this pull request should be `19.3.0` unless we are aiming to apply a similar workaround to avoid double application of scripts when upgrading from a bug fix release.

ramls/oai-pmh-view.raml
oleksandr-yatsenko1: Thank you I will fix it.

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: > This indicates that the release version should be 19.3.0 to signal that it is a feature version. I understand the hot-fix releases should be restricted to bug fixes, but this pull request qualifies as an exceptions to this rule because it has very low risk.

I believe the mainline version is already 19.3.0, meaning that this pull request does not need to actively change the version.

ramls/oai-pmh-view.raml
julianladisch: Not on GitHub yet.

ramls/oaipmh/oaipmhupdatedinstanceids.json
julianladisch: Yes, "oaiRecordUpdatedDate" is fine, for me "updatedDate" is also sufficient because it is in the oaipmhupdatedinstanceids.json object, no need to repeat the object name in the object's properties.

oleksandr-yatsenko1: Thank you Julian, I'll use this name - "updatedDate".

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: It looks genetared by raml OaiPmhView inteface has this method annotated with @Validate. Is it not sufficient to allow implementation to have this annotation too?
image

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -80,6 +94,28 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
...
+
+    UUID[] ids = entity.getInstanceids().stream().map(UUID::fromString).toArray(UUID[]::new);
+
+    getCommonInstanceProcessing(SQL_INSTANCES,
+      () -> createPostgresParams(ids, entity.getSkipSuppressedFromDiscoveryRecords()),
+      routingContext, okapiHeaders, asyncResultHandler, vertxContext,
+      "Select from oai pmh instances view completed successfully");
+  }
+
+  @Override
```
julianladisch, line 109: Can you add `@Validate`? The .raml file says `is: [validate]`.

oleksandr-yatsenko1: Done

src/main/resources/templates/db_scripts/schema.json
oleksandr-yatsenko1: Thank you for clarification, so the version 19.2.5 will be retained which doesn't forbid to include thesechanged to the future 19.3.0 from main buildline.

ramls/oaipmh/oaipmhenrichedinstances.json
```diff
@@ -0,0 +1,22 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Response schema for oai-pmh enriched instances view",
+  "type": "object",
+  "properties": {
+    "instanceid": {
```
bohdan-suprun, line 6:
```suggestion
    "instanceId": {
```

ramls/oaipmh/oaipmhenrichedinstances.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "OAI-PMH enriched instance id",
+      "type": "string",
+      "$ref": "../uuid.json"
+    },
+    "itemsandholdingsfields": {
```
bohdan-suprun, line 11:
```suggestion
    "itemsAndHoldingsFields": {
```

ramls/oaipmh/oaipmhenrichedinstances.json
```diff
@@ -0,0 +1,22 @@
...
+      "description": "Aggregated item and holding fields",
+      "type": "object",
+      "$ref": "oaipmhinstance.json"
+    }
+  },
+  "required": [
+    "instanceid",
+    "itemsandholdingsfields"
```
bohdan-suprun, line 19:
```suggestion
    "instanceId",
    "itemsAndHoldingsFields"
```

ramls/oaipmh/oaipmhinstanceids.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Define format of request data with instances ids",
+  "type": "object",
+  "properties": {
+    "instanceids": {
```
bohdan-suprun, line 6:
```suggestion
    "instanceIds": {
```

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -36,23 +42,33 @@
...
+      routingContext, okapiHeaders, asyncResultHandler, vertxContext,
+      "Select from oai pmh view completed successfully");
+  }
+
+  private void getCommonInstanceProcessing(String sql, Supplier<Tuple> paramsSupplier, RoutingContext routingContext, Map<String, String> okapiHeaders,
+      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext, String logMessage) {
 
     try {
-      Tuple params = createPostgresParams(startDate, endDate, deletedRecordSupport, skipSuppressedFromDiscoveryRecords);
+      log.debug("request params:", Iterables.toString(routingContext.request().params()));
```
bohdan-suprun, line 64:
```suggestion
      if (log.isDebugEnabled()) {
         log.debug("request params:", Iterables.toString(routingContext.request().params()));
      }
```

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -36,23 +42,33 @@
...
 
-    log.debug("request params:", Iterables.toString(routingContext.request()
-      .params()));
+    getCommonInstanceProcessing(SQL,
+      () -> createPostgresParams(startDate, endDate, deletedRecordSupport, skipSuppressedFromDiscoveryRecords),
+      routingContext, okapiHeaders, asyncResultHandler, vertxContext,
+      "Select from oai pmh view completed successfully");
+  }
+
+  private void getCommonInstanceProcessing(String sql, Supplier<Tuple> paramsSupplier, RoutingContext routingContext, Map<String, String> okapiHeaders,
```
bohdan-suprun, line 60: Maybe rename the method?
```suggestion
  private void fetchInstancesByQuery(String sql, Supplier<Tuple> paramsSupplier, RoutingContext routingContext, Map<String, String> okapiHeaders,
```

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -80,6 +96,30 @@ public void getOaiPmhViewInstances(String startDate, String endDate, boolean del
     }
   }
 
+  @Validate
+  @Override
+  public void postOaiPmhViewEnrichedInstances(OaipmhInstanceIds entity, RoutingContext routingContext,
+      Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+
+    UUID[] ids = entity.getInstanceids().stream().map(UUID::fromString).toArray(UUID[]::new);
```
bohdan-suprun, line 104: Maybe directly use `String[]` here? assuming that the payload might be quite big.

oleksandr-yatsenko1: Usaing of String array allows us to eliminate extra conversion, because later array if UUIDs will be passed as array of Strings in payload. But internally UUID type is used to be sure that passed data has UUID type, ans not just any strings. Possibly this is not a big overhead, I mean conversion from String to UUID and then back to String for payload, because data consistency is on the other side. And the by fact, we deal with UUID therefore I believe we need to use this type of data inside the code.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: It looks like you mean that even internally used method needs to be more self-explanatoryform semantic point of view. In this case taking into account thta this "common processing" part of code intended to return no only instance relatd data but also items and holdings and all these entities related to "oai records" conceptually. So possibly the name like "fetchRecordsByQuery" would be more appropriate in this case. I intentionally skipped part of name like "OaiPmh" ("getOaiPmh....") because the code itself belong to module contained this in name.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
bohdan-suprun: Ok, let's leave as it is.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
bohdan-suprun: Yes, `fetchRecordsByQuery` works for me as well.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
marcjohnson-kint: What happens if a client provides a ID that is not a valid UUID?

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: Possibly it would be event better to avoud double ckecking of debug enabling and just use parametrized messages, which in turns uses lazy message formatting, I mean after detecting that debug leven is enabed.

ramls/oaipmh/oaipmhinstanceids.json
oleksandr-yatsenko1: Done

ramls/oaipmh/oaipmhenrichedinstances.json
oleksandr-yatsenko1: Done

ramls/oaipmh/oaipmhenrichedinstances.json
oleksandr-yatsenko1: Done

ramls/oaipmh/oaipmhenrichedinstances.json
oleksandr-yatsenko1: Done

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
oleksandr-yatsenko1: Thank you.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
julianladisch: RMB only looks at the implementation, not the interface, to decide whether to do validation or not.

src/main/java/org/folio/rest/impl/OaiPmhViewInstancesAPI.java
```diff
@@ -36,23 +42,33 @@
...
+      "Select from oai pmh view completed successfully");
+  }
+
+  private void fetchRecordsByQuery(String sql, Supplier<Tuple> paramsSupplier, RoutingContext routingContext, Map<String, String> okapiHeaders,
+      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext, String logMessage) {
 
     try {
-      Tuple params = createPostgresParams(startDate, endDate, deletedRecordSupport, skipSuppressedFromDiscoveryRecords);
-      log.debug("postgres params:", params);
+      log.debug("request params: {}", Iterables.toString(routingContext.request().params()));
```
bohdan-suprun, line 64:
```suggestion
      if (log.isDebugEnabled()) {
        log.debug("request params: {}", Iterables.toString(routingContext.request().params()));
      }
```

oleksandr-yatsenko1: Ok, will retain parametrized message.

PR #475, opened 2020-07-20
MODINVSTOR-480: Add tests for eye readable call number search.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-480 - Holdings record. Call number, eye readable. Search on exact data in Call number data element

### Changes:

Add tests to verify jira scenarios can be done with the chosen approach.

### DIFF
--- a/src/test/java/org/folio/rest/api/HoldingsStorageTest.java
+++ b/src/test/java/org/folio/rest/api/HoldingsStorageTest.java
@@ -31,6 +31,8 @@ import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.folio.HttpStatus.HTTP_CREATED;
+import static org.folio.HttpStatus.HTTP_UNPROCESSABLE_ENTITY;
 import static org.folio.rest.api.StorageTestSuite.TENANT_ID;
 import static org.folio.rest.support.HttpResponseMatchers.*;
 import static org.folio.rest.support.JsonObjectMatchers.hasSoleMessageContaining;
@@ -111,7 +113,7 @@ public class HoldingsStorageTest extends TestBaseWithInventoryUtil {
     assertThat(holdingFromGet.getString("permanentLocationId"), is(mainLibraryLocationId.toString()));
     assertThat(holdingFromGet.getString("hrid"), is("ho00000000001"));
 
-    List<String> tags = holdingFromGet.getJsonObject("tags").getJsonArray("tagList").getList();
+    List<String> tags = getTags(holdingFromGet);
 
     assertThat(tags.size(), is(1));
     assertThat(tags, hasItem(TAG_VALUE));
@@ -152,7 +154,7 @@ public class HoldingsStorageTest extends TestBaseWithInventoryUtil {
     assertThat(holdingFromGet.getString("instanceId"), is(instanceId.toString()));
     assertThat(holdingFromGet.getString("permanentLocationId"), is(mainLibraryLocationId.toString()));
 
-    List<String> tags = holdingFromGet.getJsonObject("tags").getJsonArray("tagList").getList();
+    List<String> tags = getTags(holdingFromGet);
 
     assertThat(tags.size(), is(1));
     assertThat(tags, hasItem(TAG_VALUE));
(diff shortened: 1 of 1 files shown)


src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -1969,6 +1969,77 @@ public void canSearchByDiscoverySuppressProperty() throws Exception {
...
+        .forInstance(instance.getId())
+        .withCallNumber("GE77 .F73 ")
+        .withCallNumberSuffix("2014 Curriculum Materials Collection")
+        .withPermanentLocation(mainLibraryLocationId));
+
+    final List<UUID> foundHoldings = searchByCallNumberEyeReadable("GE77 .F73 2014");
+
+    assertThat(foundHoldings.size(), is(2));
+    assertThat(foundHoldings, hasItems(firstHoldingsToMatch.getId(),
+      secondHoldingsToMatch.getId()));
```
marcjohnson-kint, line 2001: This would be the case prior to the agreed changes (to include call number alone) to the definition of readable call number?

Is there a similar test that searches for only `GE77 .F73 ` and gets a single match?

bohdan-suprun: I only covered jira scenarios. 

> Is there a similar test that searches for only GE77 .F73 and gets a single match?

In this case nothing is expected to be returned, I think, because it matches only part of the call number.

marcjohnson-kint: My understanding is that you suggested to Charlotte that a way to get the behaviour asked for, which is to be able to match the call number alone (and not the whole call number) was to include it as a separate clause in the CQL query.

Did I misunderstand your suggestion?

marcjohnson-kint: If my understanding is correct, then a search for a readable call number of only the call number part should match, without the need for any right truncation.

bohdan-suprun: Yes, that's correct. 

As I understand it is not the usual right truncation, but ability to search with call number only, ignoring suffix, but at the same time we can not search by part of the call number only, unless user explicitly specified the * sign.

bohdan-suprun: Are you suggesting to split out the case `GE77 .F73 ` + `2014 Curriculum Materials Collection` into separate scenario?

marcjohnson-kint: As I understand it, the intent of this change is so that a readable call number for only the call number part without any truncation will match that holdings record.

For example, if there is a call number of `AB` and a suffix of `CD`, a search of `AB` (without explicit truncation) should match that holdings records.

Does that fit with your understanding?

bohdan-suprun: Yes, that how I understand the requirement as well. 

And the 3rd example here is call number `A` and suffix `BCD` and search for `AB` should not return this 3rd record.

marcjohnson-kint: > Are you suggesting to split out the case GE77 .F73 + 2014 Curriculum Materials Collection into separate scenario?

> It is expected that the third holdings record does not matches search criteria. Because it's call number is GE77 .F73 vs GE77 .F73 2014 expected.

I don't think the test defined here matches the intent of the scenario in the story (it also doesn't match that scenario, which I can understand because I find it confusing) because both records match with a concatenated call number and suffix which is what we had before these issues were asked for.

I think the simplest test, using the examples from the story, that matches that intent is a holdings record with a call number of `GE77 .F73 2014` and a suffix of `Curriculum Materials Collection` should be a match for a readable call number search for `GE77 .F73 2014` without any explicit truncation.

Does that fit with your understanding?

marcjohnson-kint: Having re-read the test I think I might have missed one of the holdings records being created during the test, and it does demonstrate the intended behaviour. Apologies :-(

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -1969,6 +1969,77 @@ public void canSearchByDiscoverySuppressProperty() throws Exception {
...
+      new HoldingRequestBuilder()
+        .forInstance(instance.getId())
+        .withCallNumber("GE77 .F73 ")
+        .withCallNumberSuffix("2014 Curriculum Materials Collection")
+        .withPermanentLocation(mainLibraryLocationId));
+
+    holdingsClient.create(
+      new HoldingRequestBuilder()
+        .forInstance(instance.getId())
+        .withCallNumber("GE77 .F74 ")
```
marcjohnson-kint, line 2032: Is this intentionally meant to be `.F74` instead of `F73`?

How does the inclusion of this record help describe the scenario better?

bohdan-suprun: It demonstrates that filtering really works vs just returns everything from DB. It was added to demonstrate that this record is being excluded from the search result.

marcjohnson-kint: Ok

PR #478, opened 2020-07-27
MODINVSTOR-503: Add new 'Aged to lost' item status.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-503 - Add 'Aged to lost' status to allowed item statuses list

### Changes:

* Add `Aged to lost` item status;
* Update `item-storage` API version;
* Update `item-batch-sync` API version;

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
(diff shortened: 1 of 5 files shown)


bohdan-suprun:
> Are there any tests that try to create an item with a given status?

Yes, see `src/test/java/org/folio/rest/api/ItemStorageTest.java`

PR #481, opened 2020-07-30
Hotfix/release 19.3.1

oleksandr-yatsenko1:
This change fixes version of SQL schema for "oai pmh view" fucntionality and added dropping of legacy functions before their creation.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -928,7 +928,7 @@
     {
       "run": "after",
       "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
-      "fromModuleVersion": "19.2.5"
+      "fromModuleVersion": "19.3.0"
     },
     {
       "run": "after",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,24 @@
+## 19.3.0 2020-07-28
+
+* Add 'Aged to lost' status to allowed item statuses list (MODINVSTOR-503)
+* Solving issue with GBV by making migration for preceeding and succeeding titles to be possible to run several times without failing (MODINVSTOR-541)
+* Separate filtering from response generation in oai-pmh view (MODINVSTOR-536)
+* Item record. Effective call number (item), eye readable. Search on exact data in Call number data element (MODINVSTOR-481)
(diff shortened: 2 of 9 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -928,7 +928,7 @@
     {
       "run": "after",
       "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
-      "fromModuleVersion": "19.2.5"
+      "fromModuleVersion": "19.3.0"
```
marcjohnson-kint, line 931: Given that 19.2.5 didn't trigger the execution of this script. Should this be 19.3.1 rather than 19.3.0?

If 19.2.5 being less than 19.3.0 is what caused the script not to execute, then could 19.3.0 being less than 19.3.1 cause it not to execute this time?

@julianladisch Maybe you can help me understand how this works again, because I'm not sure I'm following it.

julianladisch: Yes, it should be 19.3.1 to enforce running the .sql file when upgrading from any version below 19.3.1.

julianladisch:
I can reproduce the issue https://issues.folio.org/browse/MODINVSTOR-547 by running the mod-inventory-storage jars of version 19.2.4 (tenant install) and 19.3.0 (tenant upgrade).
This pull request fixes it and upgrade from 19.2.4 to 19.3.1 runs successfully.
It should be `"fromModuleVersion": "19.3.1"` in this pull request because it will be released as 19.3.1.
`"fromModuleVersion": "19.2.5"` in the 19.3.0 release does not cause any problem because there was no 19.2.5 release.

marcjohnson-kint:
> I can reproduce the issue https://issues.folio.org/browse/MODINVSTOR-547 by running the mod-inventory-storage jars of version 19.2.4 (tenant install) and 19.3.0 (tenant upgrade).

> This pull request fixes it and upgrade from 19.2.4 to 19.3.1 runs successfully.
> It should be `"fromModuleVersion": "19.3.1"` in this pull request because it will be released as 19.3.1.

Because some environments might have already upgraded to 19.3.0 and we want to be sure that this remedial script is triggered?

Otherwise, 19.2.5 should also be ok, because it is still greater than 19.2.4?

julianladisch:
Yes.

PR #482, opened 2020-07-30
Ported fixes of release 19.3.0 issue for mainline.

oleksandr-yatsenko1:
This change fixes version of SQL schema for "oai pmh view" fucntionality and added dropping of legacy functions before their creation for the mainline.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -928,7 +928,7 @@
     {
       "run": "after",
       "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
-      "fromModuleVersion": "19.3.0"
+      "fromModuleVersion": "19.3.1"
     },
     {
       "run": "after",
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -928,7 +928,7 @@
     {
       "run": "after",
       "snippetPath": "oaipmh/createOaiPmhViewFunction.sql",
-      "fromModuleVersion": "19.2.5"
+      "fromModuleVersion": "19.3.0"
```
marcjohnson-kint, line 931: Given that the next mainline release is going to be 19.4.0 or 20.0.0, should this be 19.4.0 rather than 19.3.0?

julianladisch: We should use 19.3.1 because all versions >= 19.3.1 already have run the .sql file.
Otherwise upgrading from 19.3.1 to 19.4.0 will re-run the file - this does not cause any harm because it is unneccessary.

marcjohnson-kint: Ok. I dislike this workaround but I understand and accept it. Thanks @julianladisch for the explanation

PR #483, opened 2020-07-30
MODINVSTOR-548: Return 500/400, not 201, on upgrade failure

julianladisch:
TenantRefAPI:

Don't swallow HTTP error status >= 300, for example 400,
pass them.

Also fix the success HTTP status (200, 201) by using
the value returned from RMB.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -83,7 +83,7 @@ public class TenantRefAPI extends TenantAPI {
     log.info("postTenant");
     Vertx vertx = cntxt.owner();
     super.postTenant(ta, headers, res -> {
-      if (res.failed()) {
+      if (res.failed() || res.result().getStatus() >= 300) {
         hndlr.handle(res);
         return;
       }
@@ -127,8 +127,7 @@ public class TenantRefAPI extends TenantAPI {
             .respond500WithTextPlain(res1.cause().getLocalizedMessage())));
           return;
         }
-        hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
-          .respond201WithApplicationJson("")));
+        hndlr.handle(res);  // HTTP status: 200 for upgrade, 201 for install
       });
     }, cntxt);
   }
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -78,13 +78,13 @@ public class TenantRefAPI extends TenantAPI {
   }
 
   @Override
-  public void postTenant(TenantAttributes ta, Map<String, String> headers,
-    Handler<AsyncResult<Response>> hndlr, Context cntxt) {
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers,
+    Handler<AsyncResult<Response>> handler, Context context) {
     log.info("postTenant");
-    Vertx vertx = cntxt.owner();
-    super.postTenant(ta, headers, res -> {
-      if (res.failed() || res.result().getStatus() >= 300) {
-        hndlr.handle(res);
+    Vertx vertx = context.owner();
+    super.postTenant(tenantAttributes, headers, databaseInitResult -> {
+      if (databaseInitResult.failed() || databaseInitResult.result().getStatus() >= 300) {
+        handler.handle(databaseInitResult);
         return;
       }
       try {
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -83,7 +83,7 @@ public void postTenant(TenantAttributes ta, Map<String, String> headers,
     log.info("postTenant");
     Vertx vertx = cntxt.owner();
     super.postTenant(ta, headers, res -> {
-      if (res.failed()) {
+      if (res.failed() || res.result().getStatus() >= 300) {
```
marcjohnson-kint, line 86: Does a similar change need to be applied to any module that uses this facility?

Could this be handled by instead returning a failed result from the process, which would mean the client code does not need to change. 

It isn't clear from this change what status might be produced from the RAML Module Builder code. Please can you help me understand why a 4XX would be the expected response under these circumstances?

julianladisch: RestVerticle calls postTenant. RMB's postTenant returns a Response where the status is 400 and the body contains detailed information about the SQL statement that failed.
When a module uses postTenant it calls RMB's postTenant and either passes on the error Response, the success response, or creates an own error response.
If RMB's postTenant were returning a failed AsycResult it cannot pass any Response information, only an Exception.
This is the reason for this design decision.
The expected post tenant HTTP status values are 200 for successful upgrade, 201 for successful installation, 204 for skipped installation because it already exists, 400 for failed installation/upgrade, 422 for validation error, 500 for internal server error.

julianladisch: > Does a similar change need to be applied to any module that uses this facility?

Yes, see https://issues.folio.org/browse/RMB-686 "Fix documentation about postTenant error handling" which includes a note in the upgrading notes.

marcjohnson-kint: @julianladisch Thank you for explaining the design decision. It it is a pity that information could be conveyed in a more explicit manner to the client code.

It is likely outside of the scope of this issue, I think 400 is a potentially inappropriate response, as this suggests the fault likes with the client initiating the upgrade, which in this case it was not.

marcjohnson-kint, line 86: This check will also return any status codes over 300, meaning redirection statuses will be considered failures and handled as such. Is that the intended behaviour?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -127,8 +127,7 @@ public void postTenant(TenantAttributes ta, Map<String, String> headers,
             .respond500WithTextPlain(res1.cause().getLocalizedMessage())));
           return;
         }
-        hndlr.handle(io.vertx.core.Future.succeededFuture(PostTenantResponse
-          .respond201WithApplicationJson("")));
+        hndlr.handle(res);  // HTTP status: 200 for upgrade, 201 for install
```
marcjohnson-kint, line 130: Please could `res` and `res1` be renamed to better describe the processes these are results for to make it easier to understand these statements without needing to refer to the rest of the code.

Given the intent of them I think it also might make sense to use an `else` branch to make it explicit this is two parts of the same decision.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: Yes.
A redirection means that a different API needs to be called to start the installation/upgrade. As no tables have been installed/upgraded no reference/sample data should be inserted.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: This intentionally uses no else branch. A [guard clause](https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html) is used for the error handling making the success path stand out for readability.

julianladisch: The variables have been renamed.

marcjohnson-kint: Thank you

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: I think the description of this pull request and JIRA issue could be changed to explain that.

> A redirection means that a different API needs to be called to start the installation/upgrade. 

Is there an existing example of this? Should Okapi follow that redirection.

> As no tables have been installed/upgraded no reference/sample data should be inserted.

Agreed, the upgrade should fail fast.

julianladisch:
Obsoleted by https://issues.folio.org/browse/RMB-687 shipped with RMB v30.2.5.

PR #487, opened 2020-08-18
MODINVSTOR-543 Holdings Statements Staff Notes

rladdusaw:
Resolves [MODINVSTOR-543](https://issues.folio.org/browse/MODINVSTOR-543)

Added a new data element to the holdings statements on the holdings record. This is a private note that is meant to only be displayed to staff. 

This PR currently doesn't include any security to prevent someone from viewing the staff notes, but the issue seems to imply that would be handled in data export/harvest and the public/private designation is there to help make that determination.


bohdan-suprun (approved):

marcjohnson-kint:
@rladdusaw Can this be merged?

PR #492, opened 2020-08-24
MODINVSTOR-518: Transform OAI-PMH view API in Inventory into general API

oleksandr-yatsenko1:
Initial implementation of new Inventory records API.


ramls/url.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "URL definition",
+  "type": "string",
+  "pattern": "^(?:http(s)?:\\/\\/)?[\\w.-]+(?:\\.[\\w\\.-]+)+[\\w\\-\\._~:/?#[\\]@!\\$&'\\(\\)\\*\\+,;=.]+$",
```
julianladisch, line 5: I suggest to completely remove url.json. If there is a need for a validation pattern please create a separate Jira and a separate pull request because it needs further discussion of the requirements.

The regular expression pattern doesn't compile. `#[\\]` needs to be replaced by `#\\[\\]`.

After fixing the pattern there are several issues:

It takes 10 seconds to validate strings like `http://....................................Ø` because of [regular expression denial of service (ReDoS)](https://en.wikipedia.org/wiki/ReDoS); this is a security issue.

The pattern rejects these perfectly valid URLs:
* https://münster.de
* https://en.wikipedia.org/wiki/Ø
* https://en.wikipedia.org/wiki/%C3%98

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+              "description": "Aggregated electronic access from holdings",
+              "type": "array",
+              "items": [
+                {
+                  "type": "object",
+                  "properties": {
+                    "uri": {
+                      "description": "Electronic access URI",
+                      "type": "string",
+                      "$ref": "../url.json"
```
julianladisch, line 172: The underlying holdingsrecord.json uri doesn't have such a validation pattern: https://github.com/folio-org/mod-inventory-storage/blob/v19.3.1/ramls/holdingsrecord.json#L57-L60
How can you ensure that an uri that fails the validation pattern is not returned in inventory-hierarchy/inventory-instance-records.json?

I suggest to completely remove `"$ref": "../url.json"`. If there is a need for validation a separate Jira and a separate pull request should be opened to discuss the requirements.

ramls/examples/inventory-items-and-holdings.json
```diff
@@ -0,0 +1,48 @@
...
+            "name": "Related resource",
+            "linkText": "item link text",
+            "publicNote": "item public note",
+            "relationshipId": "5bfe1b7b-f151-4501-8cfa-23b321d5cd1e",
+            "materialsSpecification": "item mast"
+          }
+        ]
+      }
+    ],
+    "instanceid": "1b74ab75-9f41-4837-8662-a1d99118008d"
```
julianladisch, line 46: `instancedid` is a typo, should be `instanceId`.

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+                    "materialsSpecification"
+                  ]
+                }
+              ]
+            },
+            "notes": {
+              "description": "Notes about action, copy, binding etc.",
+              "type": "array",
+              "items": [
+                {
```
julianladisch, line 212: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+            },
+            "digitizationPolicy": {
+              "description": "Records information regarding digitization aspects",
+              "type": "string"
+            },
+            "holdingsStatements": {
+              "description": "Notes about action, copy, binding etc.",
+              "type": "array",
+              "items": [
+                {
```
julianladisch, line 240: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+                    "type": "string"
+                  }
+                }
+              ]
+            },
+            "holdingsStatementsForIndexes": {
+              "description": "Holdings record indexes statements",
+              "type": "array",
+              "items": [
+                {
```
julianladisch, line 256: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+                    "type": "string"
+                  }
+                }
+              ]
+            },
+            "holdingsStatementsForSupplements": {
+              "description": "Holdings record supplements statements",
+              "type": "array",
+              "items": [
+                {
```
julianladisch, line 272: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+            },
+            "receivingHistory": {
+              "description": "Receiving history of holdings record",
+              "type": "object",
+              "properties": {
+                "entries": {
+                  "description": "Entries of receiving history",
+                  "type": "array",
+                  "items": [
+                      {
```
julianladisch, line 300: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+            },
+            "tags": {
+              "description": "Arbitrary tags associated with this holding",
+              "type": "object",
+              "properties": {
+                "tagList": {
+                  "description": "List of tags",
+                  "type": "array",
+                  "items": [
+                    {
```
julianladisch, line 326: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,557 @@
...
+                    }
+                  ]
+                }
+              }
+            },
+            "statisticalCodes": {
+              "description": "Holdings record statistical codes",
+              "type": "array",
+              "items": [
+                {
```
julianladisch, line 340: This fails the build and should be changed to

              "items": {
                "type": "object",
                "properties": {

ramls/url.json
oleksandr-yatsenko1: > I suggest to completely remove url.json
Good point julian, it looks like validation of URLs for all the cases is not strighforward task, so it will be better to avoid such vaidation and remains with just string parameter, I agree. Especially since validation isn't a first proirity behavior here.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: > I suggest to completely remove "$ref": "../url.json"

I will remote it, thank you for noticing

ramls/examples/inventory-items-and-holdings.json
oleksandr-yatsenko1: I didn't see this typo, possibly was fixed.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: You are right, it looks like I was wrong defined arrays of objects for some entries. I'll fix it, thanks.

mrdpopov (commented):
Maybe it makes sence to name the new SQL file according to raml file?

src/main/resources/templates/db_scripts/common/createRecordsViewFunction.sql
```diff
@@ -0,0 +1,324 @@
...
+CREATE INDEX IF NOT EXISTS audit_holdings_record_pmh_createddate_idx ON ${myuniversity}_${mymodule}.audit_holdings_record ((strToTimestamp(jsonb -> 'record' ->> 'updatedDate')));
+CREATE INDEX IF NOT EXISTS audit_item_pmh_createddate_idx ON ${myuniversity}_${mymodule}.audit_item ((strToTimestamp(jsonb -> 'record' ->> 'updatedDate')));
+
+-- Creates function returned instances with records updates/deleted/added in specified period o time
+CREATE OR REPLACE function ${myuniversity}_${mymodule}.get_updated_instance_ids_view(startDate                          timestamptz,
+                                                                                     endDate                            timestamptz,
+                                                                                     deletedRecordSupport               bool DEFAULT TRUE,
+                                                                                     skipSuppressedFromDiscoveryRecords bool DEFAULT TRUE,
+                                                                                     allowTieredUpdateDate              bool DEFAULT TRUE)
```
mrdpopov, line 86: I think this name  'allowTieredUpdateDate 'is not intuitive enough. Can you name it to something that better describes its purpose?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1074,6 +1074,21 @@
           "permissionsRequired": ["inventory-storage.oai-pmh-view.enrichedinstances.collection.post"]
         }
       ]
+    },
+    {
+      "id": "inventory-hierarchy",
+      "version": "1.0",
```
bohdan-suprun, line 1080: We usually use `0.1` version for new interfaces.

ramls/inventory-hierarchy.raml
```diff
@@ -0,0 +1,70 @@
+#%RAML 1.0
+title: Inventory Hierarchy API
+version: v1.10
```
bohdan-suprun, line 3: Please update version here to match module descriptor.

src/main/resources/templates/db_scripts/common/createRecordsViewFunction.sql
```diff
@@ -0,0 +1,324 @@
+-- Creates auxiliary functions
```
bohdan-suprun, line 1: This script won't be executed unless you specify it in `schema.json`

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,813 @@
...
+                    "name",
+                    "campusName",
+                    "libraryName",
+                    "institutionName"
+                  ]
+                }
+              },
+              "required": [
+                "permanentLocation",
+                "temporaryLocation"
```
marcjohnson-kint, line 108: How can temporary location be required when it is not required for holdings in general?

ramls/inventory-hierarchy/inventory-instance-records.json
```diff
@@ -0,0 +1,813 @@
...
+      "$ref": "../uuid.json"
+    },
+    "source": {
+      "description": "Source of metadata and format of the underlying record to the instance record",
+      "type": "string"
+    },
+    "holdings": {
+      "type": "array",
+      "description": "Holdings record fields",
+      "items": {
```
marcjohnson-kint, line 18: Is this representation of holdings actively different to the main holdings definition?

ramls/examples/inventory-items-and-holdings.json
julianladisch: It is still wrong. `id` starts with a lower i, it should be a capital I: `Id`
This still fails the build.

ramls/examples/inventory-items-and-holdings.json
```diff
@@ -0,0 +1,48 @@
+{
+  "instanceId": "1b74ab75-9f41-4837-8662-a1d99118008d",
+  "itemsAndHoldingsFields": {
+    "items": [
```
julianladisch, line 4: "itemsAndHoldingsFields" has 3 required fields according to inventory-instance-records.json: instanceId, holdings, items.
This example fails the build because there is no holdings property.

descriptors/ModuleDescriptor-template.json
oleksandr-yatsenko1: Will fix it, thank you.

ramls/inventory-hierarchy.raml
oleksandr-yatsenko1: Done

src/main/resources/templates/db_scripts/common/createRecordsViewFunction.sql
oleksandr-yatsenko1: Good point, I'll add script entry there. Only question related to specification of the version, can we set next release version which is expected to be an 19.3.2 ?

marcjohnson-kint: > Only question related to specification of the version, can we set next release version which is expected to be an 19.3.2 ?

Why would we want to use a bug fix version for this? As I understand it, this change isn't intended to be back ported and so will be released from the mainline as either a new minor or major version.

ramls/inventory-hierarchy/inventory-items-and-holdings.json
```diff
@@ -0,0 +1,804 @@
...
+                  "name",
+                  "campusName",
+                  "libraryName",
+                  "institutionName"
+                ]
+              }
+            },
+            "required": [
+              "permanentLocation",
+              "temporaryLocation"
```
marcjohnson-kint, line 99: How can this be required when the temporary location for a holdings record is not required?

ramls/inventory-hierarchy/inventory-items-and-holdings.json
```diff
@@ -0,0 +1,804 @@
...
+                "type": "string"
+              },
+              "callNumber": {
+                "description": "Call Number identifier assigned to a holding",
+                "type": "string"
+              }
+            },
+            "required": [
+              "prefix",
+              "suffix",
```
marcjohnson-kint, line 130: How are these required when a call number does not have to have either a prefix or a suffix?

ramls/inventory-hierarchy/inventory-items-and-holdings.json
```diff
@@ -0,0 +1,804 @@
...
+                "id",
+                "code",
+                "name",
+                "statisticalCodeType",
+                "source"
+              ]
+            }
+          }
+        },
+        "required": [
```
marcjohnson-kint, line 764: Have these schema be generated in some way? The required properties defined here don't seem to match up to the properties that are required in the underlying records and so I don't understand how these expectations can be fulfilled.

oleksandr-yatsenko1: I'll check requirted fields from mod-inventory-storage's json schema definitions anf fix this. Thank you.

ramls/examples/inventory-items-and-holdings.json
oleksandr-yatsenko1: I fixed to use camel-case wverywhere.

src/main/resources/templates/db_scripts/common/createRecordsViewFunction.sql
oleksandr-yatsenko1: Do you mean, that we need to specify 19.4.0 version for script here?

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: I checked and fixed reirement for payload's fields.

ramls/inventory-hierarchy/inventory-instance-records.json
oleksandr-yatsenko1: I think current represetnation of holdings contains combined parts of original ones. The same for items too. Because not all parts needed and some of them needs to be presented in extended forms, I mean "names" instead of id references to other wntities, for example.

oleksandr-yatsenko1:
> There needs to be automated tests for the new APIs

Tests were added too, for requited elements in particular for each of levels.

ramls/examples/inventory-items-and-holdings.json
oleksandr-yatsenko1: I removed redundant level of hierarchy, i.e. mentioned "itemsAndHodlingsFields".

ramls/inventory-hierarchy/inventory-items-and-holdings.json
oleksandr-yatsenko1: I checked and updated requirements for payload elements.

ramls/inventory-hierarchy/inventory-items-and-holdings.json
oleksandr-yatsenko1: I checked and updated requirements for payload elements.

mrdpopov (approved):

MikhailFokanov (approved):

mrdpopov:
I suggest merging it since this is a dependency to many other stories.

PR #495, opened 2020-08-26
MODDICORE-80 Add fullTextIndex on electronicAccess.uri field for holdings record

KaterynaSenchenko:


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -679,6 +679,12 @@
           "fieldName": "fullCallNumberNormalized",
           "sqlExpression" : "normalize_holdings_full_call_number(jsonb)",
           "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
+        },
+        {
+          "fieldName": "electronicAccess",
+          "tOps": "ADD",
+          "arraySubfield": "uri",
+          "arrayModifiers": ["uri"]
```
julianladisch, line 687: This looks good. There should be a unit tests for the intended `/@` query syntax because a `/@` query will fail if the entry in schema.json is missing or wrong.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -679,6 +679,12 @@
           "fieldName": "fullCallNumberNormalized",
           "sqlExpression" : "normalize_holdings_full_call_number(jsonb)",
           "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
+        },
+        {
+          "fieldName": "electronicAccess",
```
bohdan-suprun, line 684: Does it make sense to add `Uri` at the end, because the object is quite complex, and there are some other options that potentially can be used for search? 
```suggestion
          "fieldName": "electronicAccessUri",
```

marcjohnson-kint (commented):
Changes (both the commits and the pull request) to a module should be associated with a JIRA issue for that project

KaterynaSenchenko:
Thank you for the review, but I will close this PR for now. Additional changes are required in data-import-processing-core to build queries using modifiers. MODDICORE-80 bug was fixed in a different way. I will reopen the PR with the relevant JIRA number when it is relevant again.

PR #502, opened 2020-09-04
[MODINVSTOR-542] Holdings Source: add to Accordion Administrative data in details, edit

jsavell:
Resolves https://issues.folio.org/browse/MODINVSTOR-542


src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,193 @@
...
+        } catch (Exception e) {
+          log.error(e.getMessage(), e);
+          String message = messages.getMessage(lang, MessageConsts.InternalServerError);
+          if (e.getCause() instanceof CQLParseException) {
+            message = " CQL parse error " + e.getLocalizedMessage();
+            asyncResultHandler.handle(succeededFuture(GetHoldingsRecordsSourcesResponse
+              .respond500WithTextPlain(message)));
+          }
+        }
+      });
```
bohdan-suprun, line 72: Is there any reason why standard `org.folio.rest.persist.PgUtil#get(...)` method can not be used?

src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,193 @@
...
+                asyncResultHandler.handle(Future.succeededFuture(PostHoldingsRecordsSourcesResponse
+                  .respond400WithTextPlain(msg)));
+              }
+            });
+        } catch (Exception e) {
+          log.error(e.getMessage(), e);
+          asyncResultHandler.handle(Future.succeededFuture(PostHoldingsRecordsSourcesResponse
+            .respond500WithTextPlain(messages.getMessage(lang, MessageConsts.InternalServerError))));
+        }
+      });
```
bohdan-suprun, line 107: Is there any reason why standard `org.folio.rest.persist.PgUtil#post(...)` method can not be used?

src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,193 @@
...
+              }
+              asyncResultHandler.handle(succeededFuture(DeleteHoldingsRecordsSourcesByIdResponse
+                .respond204()));
+            });
+        } catch (Exception e) {
+          log.error(e.getMessage(), e);
+          asyncResultHandler.handle(succeededFuture(DeleteHoldingsRecordsSourcesByIdResponse
+            .respond500WithTextPlain(messages.getMessage(lang, MessageConsts.InternalServerError))));
+        }
+      });
```
bohdan-suprun, line 150: Is there any reason why standard org.folio.rest.persist.PgUtil#deleteById(...) method can not be used?

src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,193 @@
...
+                asyncResultHandler.handle(succeededFuture(PutHoldingsRecordsSourcesByIdResponse
+                    .respond400WithTextPlain(msg)));
+              }
+            });
+        } catch (Exception e) {
+          log.error(e.getMessage(), e);
+          asyncResultHandler.handle(succeededFuture(PutHoldingsRecordsSourcesByIdResponse
+            .respond500WithTextPlain(messages.getMessage(lang, MessageConsts.InternalServerError))));
+        }
+      });
```
bohdan-suprun, line 186: Is there any reason why standard org.folio.rest.persist.PgUtil#put(...) method can not be used?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -633,6 +644,11 @@
           "fieldName": "illPolicyId",
           "targetTable": "ill_policy",
           "tOps": "ADD"
+        },
+        {
+          "fieldName": "holdingsRecordsSourceId",
```
bohdan-suprun, line 649: The property has `sourceId` name in schema 

```suggestion
          "fieldName": "sourceId",
```

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.folio.rest.support.Response;
+import org.folio.rest.support.ResponseHandler;
+import org.folio.rest.support.http.ResourceClient;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+public class HoldingsSourceTest extends TestBase {
```
bohdan-suprun, line 27: Please add this tests to `org.folio.rest.api.StorageTestSuite`, so that the test cases will be included into coverage.

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,116 @@
...
+    assertThat(response.getStatusCode(), is(422));
+    JsonArray errors = response.getJson().getJsonArray("errors");
+    assertThat(errors.size(), is(1));
+
+    JsonObject firstError = errors.getJsonObject(0);
+    assertThat(firstError.getString("message"), containsString("must match"));
+    assertThat(firstError.getJsonArray("parameters").getJsonObject(0).getString("key"),
+      is("id"));
+  }
+}
```
bohdan-suprun, line 116: Could you please add some more test cases:
* Can replace a source;
* can not replace not existent source; 
* can get multiple sources by query;
* can remove a source.
* Can not create a source with existing name;
* `SourceId` is saved for a holdings record.

ramls/holdings-records-source.raml
```diff
@@ -0,0 +1,47 @@
...
+  pageable: !include raml-util/traits/pageable.raml
+  searchable: !include raml-util/traits/searchable.raml
+  language: !include raml-util/traits/language.raml
+  validate: !include raml-util/traits/validation.raml
+
+resourceTypes:
+  collection: !include raml-util/rtypes/collection.raml
+  collection-item: !include raml-util/rtypes/item-collection.raml
+
+/holdings-records-sources:
```
marcjohnson-kint, line 26: This path does not match with the path in the module descriptor /holdings-records-source, which suggests it won't work when proxied via Okapi

Personally, I would be tempted to go with `/holdings-sources` or `/holdings-record-sources` rather than records, as I think the focus is on the source rather than the holdings record. What do you think?

ramls/examples/holdingsrecordssource.json
```diff
@@ -0,0 +1,4 @@
+{
```
marcjohnson-kint, line 1: @bohdan-suprun  @jsavell What do you think about using hyphens in the file names for schema and examples?

ramls/holdingsrecord.json
```diff
@@ -316,6 +316,11 @@
       "type": "object",
       "$ref": "raml-util/schemas/metadata.schema",
       "readonly": true
+    },
+    "sourceId": {
+      "description": "UUID of the holdings records source",
```
marcjohnson-kint, line 321:
```suggestion
      "description": "(A reference to) the source of a holdings record",
```

ramls/holdingsrecordssource.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "A holdings records source",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "The system assigned unique ID of the holdings records source; UUID",
```
marcjohnson-kint, line 8: Does it have to be system assigned, what happens if the client provides it on creation?

ramls/holdingsrecordssources.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "A collection of holdings records sources",
+  "type": "object",
+  "properties": {
+    "holdingsRecordsSources": {
```
marcjohnson-kint, line 6: Personally, I would be tempted to name this `sources` (or maybe `holdingsSources`). What do you think?

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,116 @@
...
+
+  private static ResourceClient holdingsSourceClient;
+
+  @BeforeClass
+  public static void beforeAll() {
+    holdingsSourceClient = ResourceClient.forHoldingsSource(client);
+  }
+
+  @Test
+  public void canCreateHoldingSource()
```
marcjohnson-kint, line 37:
```suggestion
  public void canCreateHoldingsSource()
```

marcjohnson-kint (commented):
As I understand it, the two sources: `FOLIO` and `MARC` are intended to be provided as reference data, please could those be added.

They also shouldn't be able to be deleted. Please disallow this and add tests to demonstrate it.

It also shouldn't be possible to delete a source that is being referenced, please add a test to demonstrate that.

marcjohnson-kint (commented):
The front end story suggests there is also a `source` for each source (confusing, I know). There does not appear to be a property for that at the moment.

ramls/examples/holdingsrecordssource.json
bohdan-suprun: I like this idea.

descriptors/ModuleDescriptor-template.json
```diff
@@ -46,7 +46,7 @@
     },
     {
       "id": "holdings-storage",
-      "version": "4.3",
+      "version": "4.4",
```
bohdan-suprun, line 49: Please update the RAML file as well.

ramls/examples/holdingsrecordssource.json
jsavell: Looking at the other json files in the directory, there are some underscored, some hyphenated, but the vast majority have no separator, which is the 'convention' these new files follow.

jsavell:
@marcjohnson-kint @bohdan-suprun 

I think I've corrected everything that was firmly requested and most of the wish list items.

Please let me know if there are more required changes.

src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,153 @@
...
+                      respond500WithTextPlain(e.getMessage())));
+                  }
+                } else {
+                  log.error("Holdings Records Sources with source of folio or marc can not be deleted");
+                  asyncResultHandler.handle(succeededFuture(GetHoldingsSourcesResponse
+                    .respond400WithTextPlain("Holdings Records Sources with source of folio or marc can not be deleted")));
+                }
+              } else {
+                log.error(reply.cause().getMessage(), reply.cause());
+                asyncResultHandler.handle(succeededFuture(GetHoldingsSourcesResponse
```
bohdan-suprun, line 126:
```suggestion
                asyncResultHandler.handle(succeededFuture(DeleteHoldingsSourcesByIdResponse
```

ramls/examples/holdingsrecordssource.json
marcjohnson-kint: I don't think there is a convention at the moment. 

I think we should start to introduce one. This pull request seems to be as good a place as any to do so. My preference would be for hyphens.

reference-data/holdings-sources/Marc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "id": "036ee84a-6afd-4c3c-9ad3-4a12ab875f59",
+  "name": "MARC",
+  "source": "marc"
```
marcjohnson-kint, line 4: Scenarios 6 and 7 of [UIIN-1229](https://issues.folio.org/browse/UIIN-1229) state that the two options for this property are `folio` and `local`.

Why does this record have a different `source` value to those?

ramls/holdingsrecordssource.json
```diff
@@ -0,0 +1,23 @@
...
+      "$ref": "uuid.json"
+    },
+    "name": {
+      "type": "string",
+      "description": "Name of the holdings records source"
+    },
+    "source": {
+      "type": "string",
+      "description": "The holdings records source"
+    }
```
marcjohnson-kint, line 18: Scenarios 6 and 7 of [UIIN-1229](https://issues.folio.org/browse/UIIN-1229) state that the two options for this property are `folio` and `local`.

I think we should consider defining an enum for this property to disallow any other values.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -594,6 +594,17 @@
         }
       ]
     },
+    {
+      "tableName": "holdings_records_source",
+      "withMetadata": false,
```
marcjohnson-kint, line 599: Why would we not want the change metadata for these records?

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,308 @@
...
+    assertThat(getResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+
+    JsonObject sourceFromGet = getResponse.getJson();
+
+    assertThat(sourceFromGet.getString("id"), is(sourceId.toString()));
+    assertThat(sourceFromGet.getString("name"), is("test source"));
+  }
+
+  @Test
+  public void canNotCreateDuplicateHoldingsSource()
```
marcjohnson-kint, line 64:
```suggestion
  public void cannotCreateHoldingsSourceWithDuplicateId()
```

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,308 @@
...
+    Response response = holdingsSourceClient.attemptToCreate(
+        new JsonObject()
+            .put("id", sourceId.toString())
+            .put("name", "new source with duplicate id")
+    	);
+    assertThat(response.getStatusCode(), is(422));
+  }
+
+  @Test
+  public void canNotCreateDuplicateHoldingsSourceName()
```
marcjohnson-kint, line 85:
```suggestion
  public void cannotCreateHoldingsSourceWithDuplicateName()
```

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,308 @@
...
+  public void canCreateHoldingsSource()
+    throws MalformedURLException, InterruptedException,
+    ExecutionException, TimeoutException {
+
+    UUID sourceId = UUID.randomUUID();
+
+    JsonObject source = holdingsSourceClient.create(
+      new JsonObject()
+      .put("id", sourceId.toString())
+      .put("name", "test source")
```
marcjohnson-kint, line 47: I'd be tempted to introduce a builder to reduce the low level detail of the structure in every test.

I'd be more than happy if that was using lombok to reduce the need for much code.

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,308 @@
...
+    instanceToCreate.put("title", "Test Instance");
+    instanceToCreate.put("instanceTypeId", "535e3160-763a-42f9-b0c0-d8ed7df6e2a2");
+
+    instancesClient.create(instanceToCreate);
+
+    holdingsSourceClient.create(
+              new JsonObject()
+              .put("id", sourceId.toString())
+              .put("name", "associated source")
+            ).getJson();
```
marcjohnson-kint, line 265: Why is the indentation and formatting of this is different to the other similar examples in this class?

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
```diff
@@ -0,0 +1,308 @@
...
+	    	    ).getJson();
+
+	    IndividualResource holdingsResponse = holdingsClient.create(new HoldingRequestBuilder()
+	      .withId(UUID.randomUUID())
+	      .forInstance(instanceId)
+	      .withPermanentLocation(mainLibraryLocationId)
+	      .withSource(sourceId));
+
+	    assertThat(holdingsResponse.getJson().getString("sourceId"), is(sourceId.toString()));
+  }
```
marcjohnson-kint, line 307: Personally, I think this test could be part of the holdings tests rather than these tests

reference-data/holdings-sources/Marc.json
jsavell: I've never heard of UIIN-1229.

I added the reference data because I thought you asked me to:

"As I understand it, the two sources: FOLIO and MARC are intended to be provided as reference data, please could those be added."

ramls/holdingsrecordssource.json
jsavell: Is that compatible with Story point 4?

https://issues.folio.org/browse/MODINVSTOR-542

"
    When the holdings record is being maintained with an underlying MARC record in the MFHD format
    Then the value of the source element is MARC
"

src/main/java/org/folio/rest/impl/HoldingsRecordsSourceAPI.java
```diff
@@ -0,0 +1,153 @@
...
+    Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    vertxContext.runOnContext(v -> {
+      try {
+        String tenantId = TenantTool.tenantId(okapiHeaders);
+        PostgresClient pgClient = PostgresClient.getInstance(vertxContext.owner(), tenantId);
+        pgClient.getById(REFERENCE_TABLE,
+            id, HoldingsRecordsSource.class,
+            reply -> {
+              if (reply.succeeded()) {
+                String source = reply.result().getSource();
```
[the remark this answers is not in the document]
jsavell: It does result in a null value for the source, but not an NPE due to the null check on line 83.

It *does* cause an NPE without the null check.

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
jsavell: I activated the .editorconfig from the git repo and this is the indentation it wants to do.

jsavell: `# default to 2 space indentation
[*]
trim_trailing_whitespace = true
insert_final_newline = true
max_line_length = 80
indent_style = space
indent_size = 2`

https://github.com/folio-org/mod-inventory-storage/blob/master/.editorconfig

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
jsavell: Does this need to happen in order to merge the PR?

src/main/resources/templates/db_scripts/schema.json
jsavell: I don't know what that property does and that code change predates me. I'm happy to change its value if you're telling me it needs to be changed.

reference-data/holdings-sources/Marc.json
marcjohnson-kint: I understand that it is confusing and that maybe I should have consolidated the requirements from the two front-end stories before the back end work started. 

[MODINVSTOR-542](https://issues.folio.org/browse/MODINVSTOR-542) is intended to provide API changes that meet the needs of [UIIN-1123](https://issues.folio.org/browse/UIIN-1123), [UIIN-1229](https://issues.folio.org/browse/UIIN-1229) and [UIIN-1134](https://issues.folio.org/browse/UIIN-1134)

As I understand it, the holdings sources that are needed are `FOLIO` and `MARC`. I believe both of those should have `source` properties of `folio` and be uneditable, whereas holdings sources created by the UI should have a source property of `local` and be editable.

Does that make sense?

Would it be helpful if I consolidated the requirements from these UI stories into the backend story? I don't tend to do that because it leads to a lot of duplication of requirements.

ramls/holdingsrecordssource.json
marcjohnson-kint: Yes, as I understand it. I believe the `source element` in that scenario is the `source` for the holdings record, rather than the `source` for the holdings source itself. Which I fully admit is really confusing and frustrating :-/

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Yes please, I believe it is needed in order to present the `last updated` column in the settings UI.

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
marcjohnson-kint: No. I think it would simplify the tests and make them more maintainable. It is only a suggestion, and if you want to ignore it or defer it, that is ok.

src/test/java/org/folio/rest/api/HoldingsSourceTest.java
marcjohnson-kint: Hmm, that is very strange, because most of the indentation in this file is using four space indentation as far as I can tell (or double indentation). 

I decided not to raise it, as it was consistent.

reference-data/holdings-sources/Marc.json
jsavell: I think I just came onto this issue late in its life cycle, so I don't have the best frame of reference.

ramls/examples/holdingsrecordssource.json
jsavell: Are you asking me to rename all the files in the directory that aren't already hyphenated as part of this PR?

marcjohnson-kint: Yes, that is what I'm suggesting. It was a suggestion, if you want to defer or disregard it, that is ok.

reference-data/holdings-sources/Marc.json
marcjohnson-kint: This isn't the first time this kind of issue has been confusing. There is a mismatch between the front end and back end work, and typically the requirements are very heavily biased towards the front end.

It hasn't come up in a while, so I'd likely forgotten how painful it can be. I've added a topic to the teams retrospective for folks to reflect upon it.

marcjohnson-kint: Also, please wait before actioning my comments about whether holdings sources should be editable. I need to get those requirements clarified.

descriptors/ModuleDescriptor-template.json
```diff
@@ -2260,6 +2312,11 @@
         "inventory-storage.oai-pmh-view.instances.collection.get",
         "inventory-storage.oai-pmh-view.updatedinstanceids.collection.get",
         "inventory-storage.oai-pmh-view.enrichedinstances.collection.post",
+        "inventory-storage.holdings-records-source.collection.get",
+        "inventory-storage.holdings-records-source.get",
+        "inventory-storage.holdings-records-source.post",
+        "inventory-storage.holdings-records-source.put",
+        "inventory-storage.holdings-records-source.delete",
```
bohdan-suprun, line 2319:
```suggestion
        "inventory-storage.holdings-sources.collection.get",
        "inventory-storage.holdings-sources.get",
        "inventory-storage.holdings-sources.post",
        "inventory-storage.holdings-sources.put",
        "inventory-storage.holdings-sources.delete",
```

bohdan-suprun (approved):

ramls/holdingsrecordssource.json
marcjohnson-kint, line 18: Should this be constrained by an enum to only allow `folio` and `local`, as I think is described in my comment on the issue that Charlotte agreed to?

jsavell: Is there an example of this approach? I'm not sure where to put the enum class within the package structure.

jsavell: I think I figured it out. Enum is a type available within the RAML spec.

ramls/holdingsrecordssource.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "A holdings records source",
+  "type": "object",
+  "properties": {
```
marcjohnson-kint, line 5: I think this record needs the `metadata` property in order for the change metadata to be recorded and available to the client.

An example of this property can be found in the holdings record.

jsavell: Added the metadata property.

PR #506, opened 2020-09-14
MODINVSTOR-535: Fix instance-relationships API permissions.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-535 - Incorrect required permissions for instance relationship endpoints

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -88,7 +88,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.4",
+      "version": "7.5",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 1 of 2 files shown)


descriptors/ModuleDescriptor-template.json
```diff
@@ -88,7 +88,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.4",
+      "version": "7.5",
```
marcjohnson-kint, line 91: Why has the interface version changed? The permissions are part of the implementation, as far as I can tell, the interface itself has not changed.

bohdan-suprun: Make sense, I'll revert this change.

marcjohnson-kint (approved):

PR #508, opened 2020-09-16
MODINVSTOR-573: Return 422 when service point user does not exist.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-573 - Unable to upgrade inventory-storage on rancher env

### DIFF
--- a/ramls/service-points-user.raml
+++ b/ramls/service-points-user.raml
@@ -10,7 +10,7 @@ documentation:
 
 types:
   servicePointsUser: !include servicepointsuser.json
-  servicepointsusers: !include servicepointsusers.json
+  servicePointsUsers: !include servicepointsusers.json
   errors: !include raml-util/schemas/errors.schema
 
 traits:
@@ -28,7 +28,7 @@ resourceTypes:
     collection:
       exampleCollection: !include examples/servicepointsusers.json
       exampleItem: !include examples/servicepointsuser.json
-      schemaCollection: servicepointsusers
+      schemaCollection: servicePointsUsers
       schemaItem: servicePointsUser
   get:
     is: [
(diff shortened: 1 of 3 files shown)


ramls/service-points-user.raml
```diff
@@ -10,7 +10,7 @@ documentation:
 
 types:
   servicePointsUser: !include servicepointsuser.json
-  servicepointsusers: !include servicepointsusers.json
+  servicePointsUsers: !include servicepointsusers.json
```
marcjohnson-kint, line 13: Why has the case of this name been changed?

src/test/java/org/folio/rest/api/ServicePointsUserTest.java
```diff
@@ -164,20 +157,36 @@ public void canUpdateServicePointUser() throws MalformedURLException,
...
     createServicePointUser(spuId, UUID.randomUUID(), null, null);
     Response getResponse = getServicePointUserById(spuId);
     assertThat(getResponse.getStatusCode(), is(200));
     deleteServicePointUserById(spuId);
     Response getResponseAgain = getServicePointUserById(spuId);
     assertThat(getResponseAgain.getStatusCode(), is(404));
+  }
+
+  @Test
+  public void cannotCreateServicePointUserIfUserAlreadyExist() throws Exception {
```
marcjohnson-kint, line 176:
```suggestion
  public void cannotCreateServicePointUserIfUserAlreadyExists() throws Exception {
```

marcjohnson-kint (approved):

ramls/service-points-user.raml
bohdan-suprun: In order to have the class generated with correct camel style.

marcjohnson-kint: Ah, I didn't know that.

PR #509, opened 2020-09-18
MODINVSTOR-575 Increase request timeout from 5 s to 10 s in unit tests

julianladisch:
In addition three catch blocks are fixed to no longer swallow
stack traces but pass on the (possibly checked) exception wrapped
in a RuntimeException.

### DIFF
--- a/src/test/java/org/folio/rest/api/LocationsTest.java
+++ b/src/test/java/org/folio/rest/api/LocationsTest.java
@@ -1,5 +1,6 @@
 package org.folio.rest.api;
 
+import static org.folio.rest.api.StorageTestSuite.TIMEOUT;
 import static org.folio.rest.support.http.InterfaceUrls.holdingsStorageUrl;
 import static org.folio.rest.support.http.InterfaceUrls.instancesStorageUrl;
 import static org.folio.rest.support.http.InterfaceUrls.itemsStorageUrl;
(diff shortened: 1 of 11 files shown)


src/test/java/org/folio/rest/api/LocationsTest.java
```diff
@@ -200,7 +201,7 @@ public void canListLocations()
     CompletableFuture<Response> getCompleted = new CompletableFuture<>();
     send(locationsStorageUrl("/"), HttpMethod.GET,
       null, SUPPORTED_CONTENT_TYPE_JSON_DEF, ResponseHandler.json(getCompleted));
-    Response getResponse = getCompleted.get(5, TimeUnit.SECONDS);
+    Response getResponse = getCompleted.get(TIMEOUT, TimeUnit.SECONDS);
```
bohdan-suprun, line 204: It is worth introducing a test utility method, e.g.:

``` java
public static <T> T get(CompletableFuture<T> future) throws Exception {
  return future != null ? future.get(10, SECONDS) : null;
}
```

adamdickmeiss: I don't think it's worth. The plain API is simple enough. And perhaps the constant might be useful for other waiting things.

adamdickmeiss (approved):

src/test/java/org/folio/rest/api/LocationsTest.java
julianladisch: The requested get method has been added as a test utility method.

julianladisch: I don't see a need for the null test. The NullPointerException will be either in get(CompletableFuture) or when calling getResponse.get...:

    Response getResponse = get(getCompleted);
    assertThat(getResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));

bohdan-suprun (approved):

nielserik (approved):

PR #511, opened 2020-09-24
capitalize WillNotLend ill reference data

jsavell:
Resolves https://issues.folio.org/browse/MODINV-314

### DIFF
--- a/reference-data/ill-policies/WillNotLend.json
+++ b/reference-data/ill-policies/WillNotLend.json
@@ -1,6 +1,6 @@
 {
   "id": "b0f97013-87f5-4bab-87f2-ac4a5191b489",
-  "name": "will not lend",
+  "name": "Will not lend",
   "source": "folio"
 }
 
--- a/src/main/resources/templates/db_scripts/updateIllPolicyWillNotLend.sql
+++ b/src/main/resources/templates/db_scripts/updateIllPolicyWillNotLend.sql
@@ -1,0 +1,1 @@
+UPDATE ${myuniversity}_${mymodule}.ill_policy SET jsonb = jsonb_set(jsonb, '{name}', '"Will not lend"') where id='b0f97013-87f5-4bab-87f2-ac4a5191b489';
--- a/src/test/java/org/folio/rest/api/IllMigrationScriptTest.java
+++ b/src/test/java/org/folio/rest/api/IllMigrationScriptTest.java
@@ -11,7 +11,7 @@ import org.junit.Test;
 
 import io.vertx.core.json.JsonObject;
 
-public class illMigrationScriptTest extends MigrationTestBase {
+public class IllMigrationScriptTest extends MigrationTestBase {
   private static final String MIGRATION_SCRIPT = loadScript("updateIllPolicyWillNotLend.sql");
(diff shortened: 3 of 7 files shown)


reference-data/ill-policies/WillNotLend.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "b0f97013-87f5-4bab-87f2-ac4a5191b489",
-  "name": "will not lend",
+  "name": "Will not lend",
```
marcjohnson-kint, line 3: Does this need to change in existing environments as well? If so, then there needs to be an upgrade script as well.

Please check with Charlotte

jsavell: Yes, I found in an issue comment that an upgrade script has been requested. I'm working on it now.

jsavell: I added an upgrade script and I think I've triggered it properly.

src/main/resources/templates/db_scripts/updateIllPolicyWillNotLend.sql
```diff
@@ -0,0 +1 @@
+UPDATE ${myuniversity}_${mymodule}.ill_policy SET jsonb = jsonb_set(jsonb, '{name}', '"Will not lend"') where id='b0f97013-87f5-4bab-87f2-ac4a5191b489';
```
bohdan-suprun, line 1: Great work! We usually create a test for a migration, you can use `org.folio.rest.api.ModesOfIssuanceMigrationScriptTest#canMigrateModeOfIssuance` as an example for your case.

jsavell: Test added.

src/test/java/org/folio/rest/api/IllMigrationScriptTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class IllMigrationScriptTest extends MigrationTestBase {
+  private static final String MIGRATION_SCRIPT = loadScript("updateIllPolicyWillNotLend.sql");
+  private static final String WILL_NOT_LEND_ID = "b0f97013-87f5-4bab-87f2-ac4a5191b489";
+  private static final String WILL_NOT_LEND_UPDATED_NAME = "Will not lend";
+
+  @Test
+  public void canMigrateIllPolicies() throws Exception {
+    Map<String, JsonObject> initialIllPolicies = new HashMap<String, JsonObject>();
+    illPoliciesClient.getAll().forEach(policy -> {
+      initialIllPolicies.put(policy.getString("id"), policy);
```
bohdan-suprun, line 23: I think, the policy should already have a proper name because it is being loaded as ref data.

I'd suggest to manually update it with the old name before executing the migration script.

bohdan-suprun (approved):
Looks good, please take a look at my comment in the test before merging the PR.

src/test/java/org/folio/rest/api/IllMigrationScriptTest.java
jsavell: I have confirmed that the tests still pass when starting from '[w]ill not lend' in the reference data.

PR #512, opened 2020-09-28
Add support for using inventory as shared index

nielserik:
- add matchKey property to Instance
 - add enumeration element "Unknown" for Item.status

This pull request suggests two changes to Inventory Storage in order to use Inventory for shared indexes, with 1)  being the most important: 

1)
It's suggested to extend the Instance schema of Inventory storage with a unique, optional property `matchKey`. 

`Instance.matchKey` can be used in a setting where multiple libraries contribute bibliographic records to the same Inventory -- a shared index --  and thus need a way to identify identical instances across multiple catalogs in order to provide updates to bibliographic data or holdings in the shared index. One option is to use match key values based on metadata in the bibliographic record for example.  

Please note that the change includes a new index added to the DB schema. 

2) 
It's suggested to extend the enumeration of `Item.status` with a new option "Unknown".  With a shared index, circulation may be handled by the participating libraries individually, outside of FOLIO. When presenting the bibliographic data and their holdings from the shared index it is thus not necessarily possible to know the status of any given item. There's not currently a status option that indicates that the item's status isn't currently known. Having that option is convenient if FOLIO Stripes module `ui-inventory` is being used for displaying records of the shared index, since it would have to display an arbitrary (misleading) status in lieu of 'Unknown', 'NA' or similar.  

These two changes are the only ones we've identified as necessary/desirable during over six months of working with Inventory Storage in a shared index setting, giving us some confidence that this _is_ what will be required.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "8.5",
+      "version": "8.6",
       "handlers": [
         {
           "methods": ["GET"],
@@ -88,7 +88,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.4",
+      "version": "7.5",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -12,6 +12,10 @@
       "type": "string",
       "description": "The human readable ID, also called eye readable ID. A system-assigned sequential ID which maps to the Instance ID"
     },
+    "matchKey": {
+      "type": "string",
+      "description" : "An unique instance identifier matching a client-side bibliographic record identification. Could be an actual local identifier or a key generated from metadata in the local bibliographic record. Enables the client to determine if a client side bibliographic record already exists as an Instance in Inventory"
+    },
     "source": {
       "type": "string",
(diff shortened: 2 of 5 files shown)


nielserik:
This pull request suggest two changes to Inventory Storage in order to use Inventory for shared indexes, with the first one being the most important: 

1)
It's suggested to extend the Instance schema of Inventory storage with a unique property `matchKey`. 

`Instance.matchKey` can be used in a setting where multiple libraries contribute bibliographic records to the same Inventory -- a shared index --  and thus need a way to identify identical instances across multiple catalogs in order to provide updates to bibliographic data or holdings in the shared index. One option is to use match key values based on metadata in the bibliographic record for example.  

2) 
It's suggested to extend the enumeration of `Item.status` with a new option "Unknown".  With a shared index, circulation may be handled by the participating libraries individually, outside of FOLIO. When presenting the bibliographic data and their holdings from the shared index it is thus not necessarily possible to know the status of any given item. There's not currently a status option that indicates that the item's status isn't currently known. Having that option is convenient if FOLIO Stripes module `ui-inventory` is being used for displaying records of the shared index, since it would have to display an arbitrary (misleading) status in lieu of 'Unknown', 'NA' or similar.

julianladisch (changes requested):
Instance matchKey and item status Unknown should be in separated pull request. I've created https://github.com/folio-org/mod-inventory-storage/pull/513 for the new item status.

The approach should be discussed on Jira:
* [MODINVSTOR-587](https://issues.folio.org/browse/MODINVSTOR-587) Instance match key for union catalog identifier
* [MODINVSTOR-588](https://issues.folio.org/browse/MODINVSTOR-588) New item status: Unknown

PR #513, opened 2020-09-29
MODINVSTOR-588: New item status: Unknown

julianladisch:
It's suggested to extend the enumeration of Item.status with a new option
"Unknown". With a shared index, circulation may be handled by the
participating libraries individually, outside of FOLIO. When presenting
the bibliographic data and their holdings from the shared index it is thus
not necessarily possible to know the status of any given item. There's not
currently a status option that indicates that the item's status isn't currently
known and item status is a required field. Having that option is convenient
if FOLIO Stripes module ui-inventory is being used for displaying records of
the shared index, since it would have to display an arbitrary (misleading)
status in lieu of 'Unknown', 'NA' orsimilar.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -88,7 +88,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "7.5",
+      "version": "7.4",
       "handlers": [
         {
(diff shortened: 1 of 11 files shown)


adamdickmeiss (approved):
Looks ok. But not an expert in the workings..

PR #515, opened 2020-09-30
MODINVSTOR-587: Add property `Instance.matchKey` for building a shared index with Inventory

nielserik:
matchKey can be used for storing a key like the Gold Rush® match key
  that identifies "same" bibliographic records from separate
  library catalogs. It basically implements a compare of metadata
  in the bibliographic records.

  This way, one Inventory Instance can represent multiple underlying
  bibliographic records.

  matchKey could conceivably hold other matching schemes instead if
  desired.

### DIFF
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -12,6 +12,10 @@
       "type": "string",
       "description": "The human readable ID, also called eye readable ID. A system-assigned sequential ID which maps to the Instance ID"
     },
+    "matchKey": {
+      "type": "string",
+      "description" : "An unique instance identifier matching a client-side bibliographic record identification. Could be an actual local identifier or a key generated from metadata in the local bibliographic record. Enables the client to determine if a client side bibliographic record already exists as an Instance in Inventory"
+    },
     "source": {
       "type": "string",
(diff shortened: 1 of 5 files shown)


ramls/instance.json
```diff
@@ -12,6 +12,10 @@
       "type": "string",
       "description": "The human readable ID, also called eye readable ID. A system-assigned sequential ID which maps to the Instance ID"
     },
+    "matchKey": {
+      "type": "string",
+      "description" : "A unique instance identifier matching a client-side bibliographic record identification scheme, in particular for a scenario where multiple separate catalogs with no shared record identifiers contribute to the same Instance in Inventory. A match key is typically generated from select, normalized pieces of metadata in bibliographic records"
```
julianladisch, line 17:
```suggestion
      "description" : "A unique instance identifier matching a client-side bibliographic record identification scheme, in particular for a scenario where multiple separate catalogs with no shared record identifiers contribute to the same Instance in Inventory. A match key is typically generated from selected, normalized pieces of metadata in bibliographic records"
```

julianladisch (approved):

ramls/instance.json
nielserik: Thank you Julian.

PR #516, opened 2020-10-05
MODINVSTOR-589 Add location code to hierarchy api

andrii-novytskyi:
Purpose:
https://issues.folio.org/browse/MODINVSTOR-589
Approach:
For data-export module, we want to use https://s3.amazonaws.com/foliodocs/api/mod-inventory-storage/inventory-hierarchy.html API. We need location code to be present in the response, modified https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql to return location code. Updated schemas and unit test.

### DIFF
--- a/ramls/examples/inventory-items-and-holdings.json
+++ b/ramls/examples/inventory-items-and-holdings.json
@@ -20,13 +20,15 @@
           "name": "Annex",
           "campusName": "City Campus",
           "libraryName": "Datalogisk Institut",
-          "institutionName": "Københavns Universitet"
+          "institutionName": "Københavns Universitet",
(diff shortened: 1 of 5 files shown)


KVupp (approved):
The changes look good to me Andrii, but were you able to test it anywhere. On our rancher environment may be?

Igor-Gorchakov (approved):

oleksandr-yatsenko1 (approved):
Looks good for me.

andrii-novytskyi:
Verified on the local machine:
The location code returned
image

PR #517, opened 2020-10-08
MODOAIPMH-237: Timeout on the initial load

IlliaDaliek:
Fix timeout on initial oai-pmh repository load (both empty until & from).

### DIFF
--- a/src/main/java/org/folio/rest/impl/InventoryHierarchyAPI.java
+++ b/src/main/java/org/folio/rest/impl/InventoryHierarchyAPI.java
@@ -1,11 +1,14 @@
 package org.folio.rest.impl;
 
+import static java.lang.String.format;
+
 import java.util.Map;
 import java.util.UUID;
 
 import javax.validation.constraints.Pattern;
 import javax.ws.rs.core.Response;
 
+import org.apache.commons.lang.StringUtils;
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.InventoryInstanceIds;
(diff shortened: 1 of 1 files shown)


src/main/java/org/folio/rest/impl/InventoryHierarchyAPI.java
```diff
@@ -14,24 +15,51 @@
...
-    fetchRecordsByQuery(SQL_UPDATED_INSTANCES_IDS,
-      () -> createPostgresParams(startDate, endDate, deletedRecordSupport, skipSuppressedFromDiscoveryRecords, tuple -> {
-        tuple.addBoolean(onlyInstanceUpdateDate);
-      }),
-      routingContext, okapiHeaders, asyncResultHandler, vertxContext,
-      "Get updated instances completed successfully");
+    if(StringUtils.isEmpty(startDate) && StringUtils.isEmpty(endDate)) {
+      String sql = SQL_INITIAL_LOAD;
+      if(skipSuppressedFromDiscoveryRecords) {
+        sql+=SUPPRESSED_TRUE_FILTER;
```
MikhailFokanov, line 47: is it correct in terms of checkstyle not to put spaces after equal sign.

MikhailFokanov (approved):
Please correct checkstyle

src/main/java/org/folio/rest/impl/InventoryHierarchyAPI.java
IlliaDaliek: > is it correct in terms of checkstyle not to put spaces after equal sign.

I've already formatted the code at the last commit

PR #520, opened 2020-10-12
MODINVSTOR-595: Upgrade master to RMB 31.1.2 and Vert.x 3.9.3

julianladisch:
This requires some code changes because Vert.x RowStream, Vert.x Pipe and Vert.x HttpServerResponse
don't work together: The response is ended before the first record has been written.

This change needs to be backported to the Goldenrod branch b19.3 when upgrading
to RMB 30.2.8 and Vert.x 3.9.3.


adamdickmeiss (commented):
When running this branch I get these.. Which I don't like.

```
NFO: response from oai pmh view:
13 Oct 2020 12:29:26:604 WARN  DefaultPromise [] An exception was thrown by io.vertx.core.net.impl.ConnectionBase$$Lambda$1115/0x0000000801391440.operationComplete()
java.lang.IllegalStateException: Response head already sent
        at io.vertx.core.http.impl.HttpServerResponseImpl.checkHeadWritten(HttpServerResponseImpl.java:638) ~[vertx-core-3.9.3.jar:3.9.3]
        at io.vertx.core.http.impl.HttpServerResponseImpl.setChunked(HttpServerResponseImpl.java:156) ~[vertx-core-3.9.3.jar:3.9.3]
        at io.vertx.core.http.impl.HttpServerResponseImpl.setChunked(HttpServerResponseImpl.java:64) ~[vertx-core-3.9.3.jar:3.9.3]
        at org.folio.rest.RestVerticle.endRequestWithError(RestVerticle.java:835) ~[domain-models-runtime-31.1.2.jar:?]
        at org.folio.rest.RestVerticle.sendResponse(RestVerticle.java:757) ~[domain-models-runtime-31.1.2.jar:?]
```

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -385,6 +402,74 @@ public void testResponseStatus400WhenRequestingWithInvalidUntilDate()
...
+
+  private HttpServerResponse succeedingHttpServerResponse() {
+    HttpServerResponse httpServerResponse = mock(HttpServerResponse.class);
+    doAnswer(AdditionalAnswers.answerVoid(
+        (Handler<AsyncResult<Void>> handler) -> handler.handle(Future.succeededFuture())))
+        .when(httpServerResponse).end(any(Handler.class));
+    return httpServerResponse;
+  }
+
+  public void testFetch300(TestContext testContext) {
```
marcjohnson-kint, line 436: This method isn't marked as a test whereas the name suggests it should be. Should it be a test?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -385,6 +402,74 @@ public void testResponseStatus400WhenRequestingWithInvalidUntilDate()
...
+    RoutingContext routingContext = mock(RoutingContext.class);
+    HttpServerResponse httpServerResponse = succeedingHttpServerResponse();
+    when(routingContext.response()).thenReturn(httpServerResponse);
+    AbstractInstanceRecordsAPIImpl.fetchRecordsByQuery("SELECT generate_series(1, 300)",
+        routingContext, () -> Tuple.tuple(), testContext.asyncAssertSuccess(response -> {
+          assertThat(response, is(nullValue()));
+        }));
+  }
+
+  public void testWriteQueueFull(TestContext testContext) {
```
marcjohnson-kint, line 446: This method isn't marked as a test whereas the name suggests it should be. Should it be a test?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -385,6 +402,74 @@ public void testResponseStatus400WhenRequestingWithInvalidUntilDate()
     });
   }
 
+  @Test
+  public void test500(TestContext testContext) {
```
marcjohnson-kint, line 406: Under what circumstances should a 500 response be returned?

From the investigation I've done, this is triggered by a null pointer exception thrown because of the null parameter. Was that intentional?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -385,6 +402,74 @@ public void testResponseStatus400WhenRequestingWithInvalidUntilDate()
...
+    RoutingContext routingContext = mock(RoutingContext.class);
+    when(routingContext.response()).thenReturn(httpServerResponse);
+    AbstractInstanceRecordsAPIImpl.fetchRecordsByQuery("SELECT 1",
+        routingContext, null, testContext.asyncAssertSuccess(response -> {
+          assertThat(response, is(nullValue()));
+          verify(httpServerResponse).close();
+        }));
+  }
+
+  private HttpServerResponse succeedingHttpServerResponse() {
```
marcjohnson-kint, line 428: Under what circumstances should the response be closed?

From the investigation I've done, this is triggered by a null pointer exception thrown because of the null parameter. However the setup includes setting that the response headers had already been written.

If a non-null value is used for the parameter (using an empty tuple like in the other tests) the expected behaviour does not occur.

Which scenario was this test intending to describe?

src/test/java/org/folio/rest/impl/AbstractInstanceRecordsAPIImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.function.Supplier;
+import javax.ws.rs.core.Response;
+import org.folio.rest.RestVerticle;
+import org.folio.rest.api.StorageTestSuite;
+
+public class AbstractInstanceRecordsAPIImpl extends AbstractInstanceRecordsAPI {
+  private static Map<String,String> okapiHeaders = Collections.singletonMap(
+      RestVerticle.OKAPI_HEADER_TENANT, StorageTestSuite.TENANT_ID);
+
+  public static void fetchRecordsByQuery(String sql, RoutingContext routingContext,
```
marcjohnson-kint, line 19: As the `fetchRecordsByQuery` method is only used in one test class, it can be moved there (if the underlying method is made public) and this class removed.

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
marcjohnson-kint, line 406: These tests don't appear to be related to the OAI-PMH view and instead are integration tests for the abstract base class used to allow for streaming fetches. I think these tests could be moved to a separate class to reflect this.

marcjohnson-kint (commented):
A lot of the code in the AbstractInstanceRecordsAPI class appears to be fairly general purpose, is there an intention to move this code to RAML Module Builder, maybe as part of PgUtil or similar?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
marcjohnson-kint, line 446: What behaviour is this test intended to demonstrate? I'm not confident I can work it out from the code

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
marcjohnson-kint, line 436: What behaviour is this test intended to demonstrate? I'm not confident I can work it out from the code

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
julianladisch: 500 should be returned when AbstractInstanceRecordsAPI.respondWithError is called, this PR doesn't change this.

julianladisch: The null pointer exception is one way to trigger a respondWithError call and is intentional to unit test the catch clause.

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
julianladisch: Yes.

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
julianladisch: This method returns a mock. The next two unit tests use it.

marcjohnson-kint: Apologies, this comment should have been on the TCP close test

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
```diff
@@ -385,6 +402,74 @@ public void testResponseStatus400WhenRequestingWithInvalidUntilDate()
...
+    RoutingContext routingContext = mock(RoutingContext.class);
+    when(routingContext.response()).thenReturn(mock(HttpServerResponse.class));
+    AbstractInstanceRecordsAPIImpl.fetchRecordsByQuery("SELECT 1",
+        routingContext, null, testContext.asyncAssertSuccess(response -> {
+          assertThat(response.getStatus(), is(500));
+        }));
+  }
+
+  @Test
+  public void testTcpClose(TestContext testContext) {
```
marcjohnson-kint, line 416: Under what circumstances should the response be closed?

From the investigation I've done, this is triggered by a null pointer exception thrown because of the null parameter. However the setup includes setting that the response headers had already been written.

If a non-null value is used for the parameter (using an empty tuple like in the other tests) the expected behaviour does not occur.

Which scenario was this test intending to describe? Is it that the response is closed when an error occurs and the response has already been (partially) written?

src/test/java/org/folio/rest/api/OaiPmhViewTest.java
julianladisch: Can successfully process 300 records, more than RMB's cursor page size of 100.
Other APIs had regressions at 100:
https://issues.folio.org/browse/RMB-732
https://issues.folio.org/browse/MODINVSTOR-574
https://issues.folio.org/browse/MODINVSTOR-592

marcjohnson-kint: > Can successfully process 300 records, more than RMB's cursor page size of 100.

By successfully process, are you referring to writing the rows from the database to the HTTP response?

PR #522, opened 2020-10-14
MODINVSTOR-595: Code cleanup, fix 'Response head already sent'

julianladisch:
* Fix 'Response head already sent' when doing response.end()
* Move AbstractInstanceRecordsAPI tests from OaiPmhViewTest and
  AbstractInstanceRecordsAPIImpl into AbstractInstanceRecordsAPITest
* Add missing `@Test` annotations
* Test that HttpServerResponse.write has been called 300 times
* Rename succeedingHttpServerResponse to getHttpServerResponseMock
* Bump JUnit version, old version causes security warnings


marcjohnson-kint (commented):
Thanks @julianladisch this is similar to an experiment branch I put together yesterday - https://github.com/folio-org/mod-inventory-storage/tree/MODINVSTOR-595-marcs-experiment

I would prefer it if the test names described the situation they are trying to describe in a more specific way.

src/test/java/org/folio/rest/impl/AbstractInstanceRecordsAPITest.java
```diff
@@ -0,0 +1,123 @@
...
+  public static void fetchRecordsByQuery(String sql, RoutingContext routingContext,
+      Supplier<Tuple> paramsSupplier, Handler<AsyncResult<Response>> asyncResultHandler) {
+
+    new AbstractInstanceRecordsAPI(){}.fetchRecordsByQuery(
+        sql, paramsSupplier, routingContext, okapiHeaders,
+        asyncResultHandler, Vertx.vertx().getOrCreateContext(), null);
+  }
+
+  @Test
+  public void test500(TestContext testContext) {
```
marcjohnson-kint, line 53: I find it helpful to understand the scenario the test is trying to describe from the name.

I think this test demonstrates that a 500 response should be sent when an error occurs during fetching of records.

```suggestion
  public void shouldRespondWith500StatusWhenErrorsOccursWhilstFetchingRecords(TestContext testContext) {
```

julianladisch: The test method names have been renamed.

src/test/java/org/folio/rest/api/AbstractInstanceRecordsAPITest.java
```diff
@@ -0,0 +1,124 @@
...
+        }));
+  }
+
+  private HttpServerResponse getHttpServerResponseMock() {
+    HttpServerResponse httpServerResponse = mock(HttpServerResponse.class);
+    doAnswer(AdditionalAnswers.answerVoid(
+        (Handler<AsyncResult<Void>> handler) -> handler.handle(Future.succeededFuture())))
+        .when(httpServerResponse).end(any(Handler.class));
+    return httpServerResponse;
+  }
```
marcjohnson-kint, line 82: Personally, I would move this method to the bottom. In test classes I prefer that all of the tests are listed first, as I consider those to be the important parts of the class.

marcjohnson-kint (approved):

PR #535, opened 2020-10-28
MODINVSTOR-508 Item status date missing when set default value Available

K-Felk:
Addressed this by making the item post method check for available status,
and fill it in with the item creation date from the item metadata.The status
check may not be strictly necessary, from what I can see, new item records
are automatically given a status of Available, and you can't change that.
I'm assuming that the method I'm altering is only used for creating new item
records-it looks like the "put" method is used for updating item data.
I also had to modify several tests that expected the initial status of a
created item to be null.

refs: https://issues.folio.org/browse/MODINVSTOR-508

### DIFF
--- a/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
@@ -42,6 +42,8 @@ public class ItemBatchSyncAPI implements ItemStorageBatchSynchronous {
 
     for (Item item : items) {
       futures.add(setHrid(item, hridManager));
+      item.getStatus().setDate(item.getMetadata().getCreatedDate());
+
     }
 
     CompositeFuture.all(futures)
--- a/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
@@ -42,7 +42,6 @@ public class ItemBatchSyncAPI implements ItemStorageBatchSynchronous {
 
     for (Item item : items) {
       futures.add(setHrid(item, hridManager));
-      item.getStatus().setDate(item.getMetadata().getCreatedDate());
 
     }
 
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -56,14 +56,9 @@ public class ItemStorageAPI implements ItemStorage {
       RoutingContext routingContext, Map<String, String> okapiHeaders,
       Handler<AsyncResult<Response>> asyncResultHandler,
       Context vertxContext) {
-    //if the initial status is "Available", set the last updated date of that 
-    //status to be the creation date of the item
+  
+    entity.getStatus().setDate(entity.getMetadata().getCreatedDate());
     
-    if (entity.getStatus().getName().toString().equals("Available")) {
-      entity.getStatus().setDate(entity.getMetadata().getCreatedDate());
-    }
-    
-
     final Future<String> hridFuture;
     if (isBlank(entity.getHrid())) {
       final HridManager hridManager = new HridManager(vertxContext,
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -56,6 +56,13 @@ public class ItemStorageAPI implements ItemStorage {
       RoutingContext routingContext, Map<String, String> okapiHeaders,
       Handler<AsyncResult<Response>> asyncResultHandler,
       Context vertxContext) {
+    //if the initial status is "Available", set the last updated date of that 
+    //status to be the creation date of the item
+    
+    if (entity.getStatus().getName().toString().equals("Available")) {
+      entity.getStatus().setDate(entity.getMetadata().getCreatedDate());
+    }
+    
 
     final Future<String> hridFuture;
     if (isBlank(entity.getHrid())) {
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -1265,7 +1265,7 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
   }
 
   @Test
-  public void cannotSetStatusDate() throws Exception {
+  public void cannotChangeStatusDate() throws Exception {
     UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
     UUID id = UUID.randomUUID();
     JsonObject itemToCreate = smallAngryPlanet(id, holdingsRecordId);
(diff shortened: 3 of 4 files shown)


K-Felk:
I'm unsure of how concerned I need to be about the build failure-it looks like there was some kind of timeout error-I'm assuming it's unrelated to my changes.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -56,6 +56,13 @@ public void postItemStorageItems(
       RoutingContext routingContext, Map<String, String> okapiHeaders,
       Handler<AsyncResult<Response>> asyncResultHandler,
       Context vertxContext) {
+    //if the initial status is "Available", set the last updated date of that 
+    //status to be the creation date of the item
+    
+    if (entity.getStatus().getName().toString().equals("Available")) {
```
marcjohnson-kint, line 62: Why is this date only set when the status is available?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1323,16 +1326,19 @@ public void statusUpdatedDateRemainsAfterUpdate() throws Exception {
   }
 
   @Test
-  public void statusUpdatedDateIsNullOnSubsequentUpdates() throws Exception {
+  public void statusUpdatedDateIsUnchangedOnSubsequentUpdates() throws Exception {
```
marcjohnson-kint, line 1329: Thank you for changing the name of this test to reflect the changes

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1274,8 +1274,11 @@ public void cannotSetStatusDate() throws Exception {
     JsonObject createdItem = getById(id).getJson();
     JsonObject initialStatus = createdItem.getJsonObject("status");
 
+    String createdDate = createdItem.getJsonObject("metadata").getString("createdDate");
+
+
     assertThat(initialStatus.getString("name"), is("Available"));
-    assertThat(initialStatus.getString("date"), nullValue());
+    assertThat(initialStatus.getString("date"), is(createdDate));
```
marcjohnson-kint, line 1281: Please can the name of this test be changed to reflect these changes

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: Does this test check that status date does not change when the item is updated and status hasn't been changed?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
K-Felk: Because the bug is written in such a way that it implies this is only desirable when the initial status is available.  As I said above, I'm not entirely sure this check is necessary, from what I can see, newly-created items can *only* have a status of available.

src/test/java/org/folio/rest/api/ItemStorageTest.java
K-Felk: That's what it looks like.  I'm not super-familiar with the testing syntax, but it looks like what it's doing is copying the item data and then overwriting it to force an update, then checks to see if the status data has changed.  Hmm.  It might then be better if I only grab the items created date once, and use that for all the tests.  I'll make that change.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
marcjohnson-kint: Please can you check with the author of the JIRA issue if that check is needed.

I believe it is possible to create items in other statuses, I think available is the default.

src/test/java/org/folio/rest/api/ItemStorageTest.java
K-Felk: Yep.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: I wonder if it is worth making that aspect explicit in the name

```suggestion
  public void statusUpdatedDateIsUnchangedAfterAnUpdateThatDoNotChangeStatus() throws Exception {
```

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
bohdan-suprun: Yes, I agree that we should set the date for all the statuses.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -56,6 +56,13 @@ public void postItemStorageItems(
       RoutingContext routingContext, Map<String, String> okapiHeaders,
       Handler<AsyncResult<Response>> asyncResultHandler,
       Context vertxContext) {
+    //if the initial status is "Available", set the last updated date of that
```
bohdan-suprun, line 59: It is also possible to create an item using Item batch API, could you please apply the same logic to the `org.folio.rest.impl.ItemBatchSyncAPI` class?

K-Felk: Good catch, done.

K-Felk: Hmm.  Trying to set the updated date within the class results in what looks like every test that batch-creates items failing with a 400 error.  It looks like implementing that in ItemBatchSyncAPI is not going to be as simple as adding the same logic.  I'll probably ned to close this PR and go figure out how to implement it, and then re-issue.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
K-Felk: Charlotte confirms this.  I'll remove the check.

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
K-Felk: Also, from looking at the code, it looks like that class may actually be handing off the items one by one to the ItemStorageAPI class, in which case the status date should get set by the post method there.  I'll need to run a batch load to be sure.

src/main/java/org/folio/rest/impl/ItemBatchSyncAPI.java
```diff
@@ -42,6 +42,7 @@ public void postItemStorageBatchSynchronous(boolean upsert, ItemsPost entity, Ma
 
     for (Item item : items) {
       futures.add(setHrid(item, hridManager));
+
```
marcjohnson-kint, line 45: Please can this blank line be removed?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1274,8 +1274,11 @@ public void cannotSetStatusDate() throws Exception {
     JsonObject createdItem = getById(id).getJson();
     JsonObject initialStatus = createdItem.getJsonObject("status");
 
+    String createdDate = createdItem.getJsonObject("metadata").getString("createdDate");
+
+
```
marcjohnson-kint, line 1279: Why are there two blank lines between this and the assertions?

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
K-Felk: Okay, I've written a test, and assuming it's correct, it's indicating that batch-created items don't have status dates set with present changes.  So I'll have to go figure out how to make that happen.

PR #539, opened 2020-11-03
MODINVSTOR-509: Instance. New Instance. Instance status date is missing

K-Felk:
In line with MODINVSTOR-508, I changed operations here to use the current 
date/time as the initial status updated date.  I felt again that it was 
important that creation operations behave consistently.

Getting the current date and time and inserting it into the instance creation
process was the easy part.  What took more time was rewriting tests that 
either expected a null value, or were not checking the status update field
for any value at all.  It also took a while for me to realize there were two 
separate endpoints that batch-create instances.  For single instance and regular
batch-creation processes, I built checking for the status update date into existing 
tests where it seemed to make sense, usually into the initial check to see if such 
items could be correctly created.  For batch asynchronous creation, it was actually 
easier to create a separate test just for checking the status date.

refs: https://issues.folio.org/browse/MODINVSTOR-509

### DIFF
--- a/src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
+++ b/src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
@@ -22,6 +22,8 @@ import javax.ws.rs.core.Response;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.text.SimpleDateFormat;
+import java.util.Date;
 
 public class InstanceBatchSyncAPI implements InstanceStorageBatchSynchronous {
   @Validate
@@ -36,8 +38,12 @@ public class InstanceBatchSyncAPI implements InstanceStorageBatchSynchronous {
     final List<Future> futures = new ArrayList<>();
     final HridManager hridManager = new HridManager(Vertx.currentContext(), postgresClient);
 
+    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    final String currentDate = sdf.format(new Date());
+
     for (Instance instance : instances) {
       futures.add(setHrid(instance, hridManager));
+      instance.setStatusUpdatedDate(currentDate);
     }
 
     CompositeFuture.all(futures).setHandler(ar -> {
--- a/src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
+++ b/src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
@@ -1,6 +1,7 @@
 package org.folio.rest.impl;
 
 import static org.apache.commons.lang3.StringUtils.isBlank;
+import static org.folio.rest.support.StatusUpdatedDateGenerator.generateStatusUpdatedDate;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.CompositeFuture;
@@ -38,12 +39,10 @@ public class InstanceBatchSyncAPI implements InstanceStorageBatchSynchronous {
     final List<Future> futures = new ArrayList<>();
     final HridManager hridManager = new HridManager(Vertx.currentContext(), postgresClient);
 
-    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
-    final String currentDate = sdf.format(new Date());
-
+    final String statusUpdatedDate = generateStatusUpdatedDate();
     for (Instance instance : instances) {
       futures.add(setHrid(instance, hridManager));
-      instance.setStatusUpdatedDate(currentDate);
+      instance.setStatusUpdatedDate(statusUpdatedDate);
     }
 
     CompositeFuture.all(futures).setHandler(ar -> {
--- a/src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
+++ b/src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
@@ -1,0 +1,13 @@
+package org.folio.rest.support;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public final class StatusUpdatedDateGenerator {
+
+    public static String generateStatusUpdatedDate() {
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+      return sdf.format(new Date());
+    } 
+    
+}
--- a/src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
+++ b/src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
@@ -4,10 +4,8 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 
 public final class StatusUpdatedDateGenerator {
-
     public static String generateStatusUpdatedDate() {
       SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
       return sdf.format(new Date());
     } 
-    
 }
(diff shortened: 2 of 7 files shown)


src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
```diff
@@ -36,8 +38,12 @@ public void postInstanceStorageBatchSynchronous(boolean upsert, InstancesPost en
     final List<Future> futures = new ArrayList<>();
     final HridManager hridManager = new HridManager(Vertx.currentContext(), postgresClient);
 
+    SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+    final String currentDate = sdf.format(new Date());
```
marcjohnson-kint, line 42: Given this code appears in this class, `InstanceStorageAPI` and `InstanceStorageBatchAPI`, could it be worth extracting a shared method to do this? It might also help describe the intent of this code.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -208,8 +211,8 @@ public void canCreateAnInstanceWithoutProvidingID()
     JsonArray identifiers = instanceFromGet.getJsonArray("identifiers");
     assertThat(identifiers.size(), is(1));
     assertThat(identifiers, hasItem(identifierMatches(UUID_ISBN.toString(), "9781473619777")));
-    assertThat(
-      instanceFromGet.getString(STATUS_UPDATED_DATE_PROPERTY), nullValue());
+    assertThat(instanceFromGet.getString(STATUS_UPDATED_DATE_PROPERTY), notNullValue());
+    assertThat(instanceFromGet.getString(STATUS_UPDATED_DATE_PROPERTY), hasIsoFormat());
```
marcjohnson-kint, line 215: Does `hasIsoFormat` allow null as a valid value?

If it does not, could the `notNullValue` above be redundant?

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1881,13 +1892,38 @@ public void canPostSynchronousBatch() throws Exception {
...
 
+  @Test
+  public void newSynchronousBatchInstancesShouldHaveInitialStatusDate() throws Exception {
+    JsonArray instancesArray = new JsonArray();
+    
+    for (int i = 0; i < 3; i++) {
+      instancesArray.add(smallAngryPlanet(UUID.randomUUID()));
+    }
+
+    JsonObject instanceCollection = new JsonObject().put(INSTANCES_KEY, instancesArray);
```
marcjohnson-kint, line 1911: Could it be worth extracting a method here to convey the intent

```suggestion
    final var batchRequest = createRequestForThreeInstances()
```

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1881,13 +1892,38 @@ public void canPostSynchronousBatch() throws Exception {
...
+
+    CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+    client.post(instancesStorageSyncUrl(""), instanceCollection, TENANT_ID, ResponseHandler.empty(createCompleted));
+    assertThat(createCompleted.get(30, SECONDS), statusCodeIs(HttpStatus.HTTP_CREATED));
```
marcjohnson-kint, line 1915: Could it be worth extracting a method here as well
```suggestion
    final var createCompleted = createInstancesBatch(batchRequest);
    
    assertThat(createCompleted.get(30, SECONDS), statusCodeIs(HttpStatus.HTTP_CREATED));
```

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1881,13 +1892,38 @@ public void canPostSynchronousBatch() throws Exception {
...
+
     assertExists(uprooted);
     assertExists(smallAngryPlanet);
     assertExists(temeraire);
 
     assertNotSuppressedFromDiscovery(instancesArray);
   }
 
+  @Test
+  public void newSynchronousBatchInstancesShouldHaveInitialStatusDate() throws Exception {
```
marcjohnson-kint, line 1904:
```suggestion
  public void instancesCreatedInABatchShouldHaveStatusDate() throws Exception {
```

src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
K-Felk: Can you explain further what you mean by shared method?  Do you mean pulling the code out and putting it in a utility class they all use?  I could certainly do that.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
K-Felk: I assumed it allows null based on how the existing tests are written-when the original test-writer is testing for a value in that field, they always test for null first and then format-so I do as well.  But I can test and make sure.

K-Felk: hasIsoFormat will not accept a null value-and it will tell you if the value was null.  so yes, the null checks are redundant, I will remove.

src/main/java/org/folio/rest/impl/InstanceBatchSyncAPI.java
marcjohnson-kint: > Do you mean pulling the code out and putting it in a utility class they all use?

Yes

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1881,13 +1892,38 @@ public void canPostSynchronousBatch() throws Exception {
...
+    CompletableFuture<Response> createCompleted = new CompletableFuture<>();
+    client.post(instancesStorageSyncUrl(""), instanceCollection, TENANT_ID, ResponseHandler.empty(createCompleted));
+    assertThat(createCompleted.get(30, SECONDS), statusCodeIs(HttpStatus.HTTP_CREATED));
+
+    Response resp;
+    for (int i = 0; i < 3; i++) {
+      resp = getById(instancesArray.getJsonObject(i).getString("id"));
+      assertThat(resp.getJson().getString(STATUS_UPDATED_DATE_PROPERTY), notNullValue());
+      assertThat(resp.getJson().getString(STATUS_UPDATED_DATE_PROPERTY), hasIsoFormat());
+    }
```
marcjohnson-kint, line 1922: Similarly to my feedback on the other pull request, I think this can be changed to use the `JSONArrayHelper` and `forEach`.

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
```diff
@@ -0,0 +1,13 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public final class StatusUpdatedDateGenerator {
+
+    public static String generateStatusUpdatedDate() {
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
+      return sdf.format(new Date());
+    } 
+
```
marcjohnson-kint, line 12: Personally, I would remove the blank lines around this method, I don't think they add to the readability of the code.

```suggestion
    public static String generateStatusUpdatedDate() {
      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
      return sdf.format(new Date());
    } 
```

marcjohnson-kint (approved):

marcjohnson-kint:
@K-Felk Please fix the failing test in the [PR merge](https://jenkins-aws.indexdata.com/job/folio-org/job/mod-inventory-storage/view/change-requests/job/PR-539/7/testReport/org.folio.rest.api/ItemStorageTest/cannotChangeStatusDate/). 

If it is comparing the initial created date and the status date, they might not be exactly the same all of the time. I believe there is a date matcher that allows some variance: `withinSecondsBefore`. Alternatively, if variance in either direction is needed, please add a new matcher to check that to `DateTimeMatchers`.

K-Felk:
Odd. This test does not fail locally, nor did it fail when I pushed it to the 508 branch, or I would not have been allowed to merge.  In any case, it's fixed.

marcjohnson-kint:
> Odd. This test does not fail locally, nor did it fail when I pushed it to the 508 branch, or I would not have been allowed to merge. In any case, it's fixed.

It's probably intermittent and not caused by your changes. I think it's best to fix these kinds of tests when they are found. If you get a green build feel free to merge this and raise a separate issue to fix the existing tests.

PR #545, opened 2020-11-11
MODINVSTOR-610: Data import with 856 --> electronicAccess match performance issues.

VRohach:
1. New "GIN-index" added for the "electronicAccess" field.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -669,6 +669,14 @@
           "tOps": "ADD"
         }
       ],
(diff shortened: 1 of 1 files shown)


bohdan-suprun:
@VRohach, have you verified that the index is really used for expected query?

VRohach:
Discussed with @bohdan-suprun . It seems like this index will be valid here. @bohdan-suprun Thank you!

bohdan-suprun (approved):

KaterynaSenchenko (approved):

PR #553, opened 2020-12-02
MODINVSTOR-625 - Change effective location for items if parent holding location is changed

siarhei-charniak:
[MODINVSTOR-625](https://issues.folio.org/browse/MODINVSTOR-625) - Change effective location for items if parent holding location is changed
Added functionality to update item effective location upon holding or item update. 
Removed legacy database triggers.


src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
```diff
@@ -307,7 +308,7 @@ public void putHoldingsStorageHoldingsByHoldingsRecordId(
                   if (Objects.equals(entity.getHrid(), existingHoldings.getHrid())) {
                     try {
                       postgresClient.startTx(connection -> {
-                        updateItemEffectiveCallNumbersByHoldings(connection, postgresClient, entity).onComplete(updateResult -> {
+                        updateItemEffectiveCallNumbersAndLocationsByHoldings(connection, postgresClient, entity).onComplete(updateResult -> {
```
MikhailFokanov, line 311: I'd change it to updateRelatedItemsAttributes for brevity.

MikhailFokanov (approved):
@marc-johnson, should we wait for your approval?

marcjohnson-kint:
> @marc-johnson, should we wait for your approval?

There is no need to wait for my approval. The teams are responsible for their own work. If you are acting as a reviewer for that team, then no further reviews are necessary.

src/main/java/org/folio/rest/impl/HoldingsStorageAPI.java
marcjohnson-kint: I might propose `updateItemEffectiveProperties` or `deriveItemEffectiveProperties` instead.

PR #558, opened 2020-12-11
MODINVSTORE-620: Domain events for instance storage.

bohdan-suprun:
## Changes:

* Add vertx kafka dependency;
* Implement kafka producer;
* Create Kafka topics on tenant init.
* Send domain events when instance created/updated/removed/all instances removed.


MikhailFokanov (approved):
Looks good.

pfilippovEPAM (approved):

src/main/java/org/folio/rest/impl/InstanceStorageAPI.java
```diff
@@ -144,11 +152,23 @@ public void postInstanceStorageInstances(
...
-                              PostInstanceStorageInstancesResponse.headersFor201().withLocation(reply.result()))));
+                    domainEventService.instanceCreated(entity)
+                    .whenComplete((notUsed, error) -> {
+                      if (error == null) {
+                        asyncResultHandler.handle(
+                          io.vertx.core.Future.succeededFuture(
+                            PostInstanceStorageInstancesResponse
+                              .respond201WithApplicationJson(entity,
+                                PostInstanceStorageInstancesResponse.headersFor201()
+                                  .withLocation(reply.result()))));
```
marcjohnson-kint, line 163: Does this mean that it has been decided that if publishing to Kafka fails (and hence any consumers will become inconsistent) the API operation should still succeed?

bohdan-suprun: No, we agreed that if kafka message failed to be published then the failure should be sent back to the client.

marcjohnson-kint: @bohdan-suprun Thanks, I misread the code.

Does this mean that we will communicate to the client that the update failed, yet the database state has been changed?

bohdan-suprun: Unfortunately yes.

marcjohnson-kint: I think it might make make sense to document that as a design decision.

@MikhailFokanov what do you think?

src/main/resources/kafka-topics.json
```diff
@@ -0,0 +1,9 @@
+{
+  "topics": [
+    {
+      "name": "inventory.instance",
+      "numPartitions": 1,
+      "replicationFactor": 1
```
marcjohnson-kint, line 6: Does this mean that all messages will be published to the same topic irrespective of tenant?

bohdan-suprun: Yes.

bohdan-suprun:
Hi @marcjohnson-kint, do you want to review the changes or it is fine to merge them?

src/main/resources/kafka-topics.json
```diff
@@ -0,0 +1,9 @@
+{
+  "topics": [
+    {
+      "name": "inventory.instance",
+      "numPartitions": 1,
```
marcjohnson-kint, line 5: Does this mean that the topic is not going to partitioned at the moment? 

I thought there was an intent to partition these messages by instance ID, did I misunderstand?

src/main/resources/kafka-topics.json
```diff
@@ -0,0 +1,9 @@
+{
+  "topics": [
+    {
+      "name": "inventory.instance",
```
marcjohnson-kint, line 4: This name is specific to the domain, however it does not mention storage. If mod-inventory were to publish higher level, domain meaningful events (rather than the change messages here), would we expect them to be published to the same topic?

PR #560, opened 2020-12-21
Handle both 201 and 204 for tenant operation MODINVSTOR-648

adamdickmeiss:
See FOLIO-2908. Purge and other operations are now handled
almost same way so code refactored a bit.


bohdan-suprun (approved):
Thank you, Adam!

PR #563, opened 2020-12-28
MODINVSTOR-657 Add limit query param to /instance-bulk/ids endpoint

illia-borysenko:
Moved ticket to mod-inventory-storage project https://issues.folio.org/browse/MODINVSTOR-657

## Approach
Add offset&&limit query params to /instance-bulk/ids endpoint
Update RecordBulkAPI class and unit tests
## Important to know
If the request contains offset&&limit query params then the response will contain the incorrect value of "totalRecords" that doesn't correspond to the actual amount of records in JSON. It happens because of the wrong building count query without limit and offset in PostgresClient.buildQueryHelper method. At the same time, the same method builds select query with limit and offset.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -131,6 +131,11 @@
       <version>1.19.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>2.6.0</version>
+    </dependency>
   </dependencies>
 
   <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -131,11 +131,6 @@
       <version>1.19.0</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>org.apache.kafka</groupId>
-      <artifactId>kafka-clients</artifactId>
-      <version>2.6.0</version>
-    </dependency>
   </dependencies>
 
   <properties>
--- a/src/main/java/org/folio/rest/impl/RecordBulkAPI.java
+++ b/src/main/java/org/folio/rest/impl/RecordBulkAPI.java
@@ -1,9 +1,10 @@
 package org.folio.rest.impl;
 
-import java.util.Map;
-
-import javax.ws.rs.core.Response;
-
+import io.vertx.core.AsyncResult;
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import io.vertx.core.Handler;
+import io.vertx.ext.web.RoutingContext;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -131,6 +131,11 @@
       <version>1.19.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.kafka</groupId>
+      <artifactId>kafka-clients</artifactId>
+      <version>2.6.0</version>
+    </dependency>
```
marcjohnson-kint, line 138: Why does a pull request that states it is adding a limit to an API endpoint require a new Kafka dependency?

marcjohnson-kint (commented):
Pull requests should be associated with a JIRA issue in that module's project. Please could this pull request be associated with a pull request in the MODINVSTOR project

src/main/java/org/folio/rest/impl/RecordBulkAPI.java
```diff
@@ -28,33 +26,31 @@
...
+            "ids", routingContext, okapiHeaders, vertxContext);
+        } else {
+          CQLWrapper wrapper = getCQL(query, INSTANCE_TABLE, limit, offset);
+          PgUtil.streamGet(INSTANCE_TABLE, RecordID.class, wrapper, null,
+            "ids", routingContext, okapiHeaders, vertxContext);
+        }
+      } catch (Exception e) {
+        LOG.error(e.getMessage(), e);
+        asyncResultHandler.handle(Future.succeededFuture(GetRecordBulkIdsResponse
+          .respond500WithTextPlain(e.getMessage())));
```
marcjohnson-kint, line 53: Why has the formatting of this code been changed during this pull request?

If folks want to make significant changes to the formatting, it could be easier to follow if that was made in a separate pull request to the behavioural change.

pom.xml
illia-borysenko: I removed it

andrii-novytskyi (approved):

src/main/java/org/folio/rest/impl/RecordBulkAPI.java
illia-borysenko: Updated to the previous formatting

illia-borysenko:
> Pull requests should be associated with a JIRA issue in that module's project. Please could this pull request be associated with a pull request in the MODINVSTOR project

Hi @marcjohnson-kint I moved the ticket from mod-data-export to mod-inventory-storage project

marcjohnson-kint:
> I moved the ticket from mod-data-export to mod-inventory-storage project

Thank you

oleksandr-dekin (approved):

PR #564, opened 2020-12-28
MODINVSTOR-639: Send domain events for items.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-639 - Domain events sending in inventory-storage for item

### Changes:
* Send domain events on item create/update/delete;
* Send domain events when items created/updated through batch API;
* Send domain events when items updated as result of holding record change;
* Introduce `*Repository` classes for item and holding - they are responsible for DB create/update/get;
* Create kafka topics **before** tenant init, because they have to be created before sample data upload;
* Implement `ItemDomainEventService` class;
* Add new kafka topic - `inventory.item`;
* Move domain events assert logic to a separate class;
* `FakeKafkaConsumer` subscribed to the new `inventory.item` topic and stores the events to a separate storage.


README.MD
```diff
@@ -213,13 +213,20 @@ duplicate HRID, the module will reject the submission.
...
-  "new": {...}, // the instance after update or create
+  "old": {...}, // the instance/item before update or delete
+  "new": {...}, // the instance/item after update or create
   "type": "UPDATE|DELETE|CREATE", // type of the event
   "tenant": "diku" // tenant name
 }
 ```
+
+Kafka partition key for the both topics is instance id (for items it is retrieved from
+associated holding record).
```
marcjohnson-kint, line 232: Which instance ID is used as the partition key when the change to an item moves the item between two instances (via a move between holdings)? Is it the instance ID before the change or after?

src/main/java/org/folio/rest/support/ResponseUtil.java
```diff
@@ -10,4 +11,12 @@ private ResponseUtil() {}
   public static boolean isUpdateSuccessResponse(Response response) {
     return response != null && response.getStatus() == HTTP_NO_CONTENT.toInt();
   }
+
+  public static boolean isDeleteSuccessResponse(Response response) {
+    return response != null && response.getStatus() == HTTP_NO_CONTENT.toInt();
```
marcjohnson-kint, line 16: The methods seem to share a high degree of commonality, could a method be extracted to check that a response has the expected status code?

README.MD
bohdan-suprun: It is instance id after the change.

src/main/java/org/folio/rest/support/ResponseUtil.java
bohdan-suprun: Shure, extracted.

README.MD
marcjohnson-kint: If the elastic search import focuses on the instance ID, how will the item be removed from the projection of the information for the instance before the change?

bohdan-suprun: It is probably responsibility of the new `mod-search` service, it should figure it out. However, we haven't fully agreed final index structure for the instances and items yet, there was an idea to have a separate index for instance, and in this case we don't need this manipulation at all.

marcjohnson-kint: > It is probably responsibility of the new `mod-search` service, it should figure it out. 

Agreed. I only raise it here because the designs of both likely need to be complementary. Decisions here could introduce constraints for the way mod-search is implemented.

> However, we haven't fully agreed final index structure for the instances and items yet, there was an idea to have a separate index for instance, and in this case we don't need this manipulation at all.

In which case, why partition items by instance ID?

I don't think I follow this.

As I understand it, from my basic knowledge of Elastic Search, searches to be performant they typically need to be fulfilled by a single index. And searches in inventory, like find the copy of Harry Potter and the Goblet of Fire with barcode 12345 rely on a combination of instance and item information.

I would imagine these kinds of searches wouldn't be well fulfilled if that information is split across indexes.

src/main/java/org/folio/rest/support/ResponseUtil.java
```diff
@@ -8,6 +9,18 @@
...
+
+  public static boolean isDeleteSuccessResponse(Response response) {
+    return responseIsInStatus(response, HTTP_NO_CONTENT.toInt());
+  }
+
+  public static boolean isCreateSuccessResponse(Response response) {
+    return responseIsInStatus(response, HTTP_CREATED.toInt());
+  }
+
+  private static boolean responseIsInStatus(Response response, int expectedStatus) {
```
marcjohnson-kint, line 23:
```suggestion
  private static boolean responseHasStatus(Response response, int expectedStatus) {
```

marcjohnson-kint: I might also be tempted to change the method to take a HttpStatus, then the `.toInt()` can be done in the shared method

README.MD
bohdan-suprun: Yeah, you're right. But Mikhail mentioned that if we use a single index for instance with item nested, than a search query, for example: `instance.title=="..." and item.status=="..."` most likely will work as an OR (will return instances even though they don't have items with the status). But maybe I'm missing something or misunderstood it. That's why I said that we haven't agreed final structure of the index.

> In which case, why partition items by instance ID?

In any case we must keep this relation in order to allow this item and instance searches.

marcjohnson-kint: > But Mikhail mentioned that if we use a single index for instance with item nested, than a search query, for example: `instance.title=="..." and item.status=="..."` most likely will work as an OR (will return instances even though they don't have items with the status). But maybe I'm missing something or misunderstood it. That's why I said that we haven't agreed final structure of the index.

Hmm, ok. Based upon my understanding of what the users want, that would be incorrect. I may well have not understood.

> > In which case, why partition items by instance ID?
> 
> In any case we must keep this relation in order to allow this item and instance searches.

Which brings me back to the challenge of making sure that the instance the item used to belong to doesn't erroneously appear in searches involving items.

src/main/java/org/folio/services/batch/BatchOperationContext.java
```diff
@@ -0,0 +1,35 @@
...
+  public BatchOperationContext(Collection<T> recordsToBeCreated, Collection<T> recordsToBeUpdated,
+    Collection<T> existingRecords) {
+
+    this.recordsToBeCreated = new ArrayList<>(recordsToBeCreated);
+    this.recordsToBeUpdated = new ArrayList<>(recordsToBeUpdated);
+    this.existingRecordsBeforeUpdate = new ArrayList<>(existingRecords);
+  }
+
+  public List<T> getRecordsToBeCreated() {
+    return unmodifiableList(recordsToBeCreated);
```
marcjohnson-kint, line 25: Why is the conversion to an unmodifiable list done at the point when the list is accessed externally?

src/test/java/org/folio/rest/support/matchers/DomainEventAsserts.java
```diff
@@ -0,0 +1,130 @@
...
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+
+import io.vertx.core.json.JsonObject;
+
+public final class DomainEventAsserts {
```
marcjohnson-kint, line 23:
```suggestion
public final class DomainEventAssertions {
```

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
```diff
@@ -0,0 +1,163 @@
...
+
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+
+public class ItemDomainEventService {
+  private static final Logger log = getLogger(ItemDomainEventService.class);
+
+  private final ItemRepository itemRepository;
+  private final HoldingsRepository holdingsRepository;
+  private final BaseDomainEventService<Item> domainEventService;
```
marcjohnson-kint, line 37: The `Base` prefix usually suggests to me that it is intended to be used as a base class for an inheritance hierarchy.

This class appears to be a collaborator rather than a part of an inheritance hierarchy, could that collaboration be better described with a different name?

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
```diff
@@ -0,0 +1,163 @@
...
+
+  public ItemDomainEventService(Context context, Map<String, String> okapiHeaders) {
+    itemRepository = new ItemRepository(context, okapiHeaders);
+    holdingsRepository = new HoldingsRepository(context, okapiHeaders);
+    domainEventService = new BaseDomainEventService<>(context, okapiHeaders, INVENTORY_ITEM);
+  }
+
+  public Function<Response, Future<Response>> itemUpdated(String instanceId, Item oldItem) {
```
marcjohnson-kint, line 45: What does this method do? Does it publish events when an item has been updated?

src/main/resources/kafka-topics.json
```diff
@@ -4,6 +4,11 @@
       "name": "inventory.instance",
       "numPartitions": 1,
       "replicationFactor": 1
+    },
+    {
+      "name": "inventory.item",
+      "numPartitions": 50,
```
marcjohnson-kint, line 10: Why are items partitioned differently to instances?

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
bohdan-suprun: Yes, this is the main its responsibility.

src/main/resources/kafka-topics.json
bohdan-suprun: We will update the property for instances soon. This will be the same number.

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
bohdan-suprun: I'll add `Common` prefix, if that's fine.

src/test/java/org/folio/rest/support/matchers/DomainEventAsserts.java
bohdan-suprun: Done

src/main/java/org/folio/services/batch/BatchOperationContext.java
bohdan-suprun: Yeah, make sense, now I wrap the collection in constructor.

src/main/java/org/folio/rest/support/ResponseUtil.java
bohdan-suprun: Make sense, refactored.

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
marcjohnson-kint: Maybe the name of the method or class could reveal that intent more.

```suggestion
  public Function<Response, Future<Response>> publishDomainEventsOnItemUpdated(String instanceId, Item oldItem) {
```

```suggestion
  public Function<Response, Future<Response>> publishItemUpdatedDomainEvent(String instanceId, Item oldItem) {
```

Or the class could be renamed to `ItemDomainEventPublisher` and the method to `onItemUpdated` or `publishItemUpdated`

```suggestion
  public Function<Response, Future<Response>> onItemUpdated(String instanceId, Item oldItem) {
```

```suggestion
  public Function<Response, Future<Response>> publishItemUpdated(String instanceId, Item oldItem) {
```

README.MD
marcjohnson-kint: @bohdan-suprun @MikhailFokanov 

> Yeah, you're right. But Mikhail mentioned that if we use a single index for instance with item nested, than a search query, for example: `instance.title=="..." and item.status=="..."` most likely will work as an OR (will return instances even though they don't have items with the status). But maybe I'm missing something or misunderstood it. That's why I said that we haven't agreed final structure of the index.

According to [this comment](https://wiki.folio.org/pages/viewpage.action?pageId=52135163&focusedCommentId=54886420#comment-54886420) by @MikhailFokanov there is intended to be an index that has both instance and item properties.

How will the index entry for the prior instance be updated with these messages?

src/main/resources/kafka-topics.json
marcjohnson-kint: Ok

MikhailFokanov (approved):
Looks good.

README.MD
bohdan-suprun: We haven't started implementation of items, so I can not answer your question. But there definitely should be some linkage between instances and items, we will make sure that item is detached from the old instance and holding.

bohdan-suprun: Had a chat with Mikhail, we will also have to send instanceId in the new/old item. I will create a separate story for this - https://issues.folio.org/browse/MODINVSTOR-649

src/main/java/org/folio/services/domainevent/ItemDomainEventService.java
bohdan-suprun: I've renamed the classes to *Publisher and added `publish` prefix to the methods.

PR #566, opened 2021-01-08
MODINVSTOR-649: Set instance id for item domain event (old and new).

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-649 - Populate instanceId for new/old item when publishing domain event.

### Changes:
* Revert effective values logic so it fetches holding records only when needed.
* Refactor item domain event service so it fetches holdings for items and sets `instanceId` property.
* Introduce `OldOrNewItem` it represents old or new property in the domain event and contains all item's properties + `instanceId`;
* Refactor item domain event service, so code is a bit more readable and clearer.
* Adjust DomainEventAssertions class to retrieve and set `instanceId` for item during assertion.


MikhailFokanov (approved):
Looks good.

pfilippovEPAM (approved):

src/main/java/org/folio/services/domainevent/OldOrNewItem.java
```diff
@@ -0,0 +1,32 @@
+package org.folio.services.domainevent;
+
+import static io.vertx.core.json.Json.decodeValue;
+import static io.vertx.core.json.Json.encode;
+
+import org.folio.rest.jaxrs.model.HoldingsRecord;
+import org.folio.rest.jaxrs.model.Item;
+
+public class OldOrNewItem extends Item {
```
hjiebsco, line 9: The name `OldOrNewItem` seems a bit interesting. Not sure what a good alternative would be. `ItemWithInstanceId`, `DomainEventItem`, `ItemMessage`, is there any other better one? Also using inheritance is simpler here than composition, but do we have any concern about potential frequent `decodeValue` call in `fromItem` methods?

hjiebsco (approved):
Looks good. Left a minor question.

src/main/java/org/folio/services/domainevent/OldOrNewItem.java
bohdan-suprun: Renamed to `ItemWithInstanceId`. Also, I'm using `@JsonUnwrapped` annotation now, so that composition is used now and decodeValue is not needed. Thanks.

PR #568, opened 2021-01-14
MODINVSTOR-640: Send holdings records domain events.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-640 - Domain events sending in inventory-storage for holdings-record.

## Changes:
* Move common logic from Item domain event to an abstract class;
* Send domain events on holdings-record create/update/delete and batch operation.
* Introduce new topic for holdings `inventory.holdings-record`.


bohdan-suprun:
Depends on https://github.com/folio-org/mod-inventory-storage/pull/566

src/main/java/org/folio/persist/AbstractRepository.java
```diff
@@ -21,19 +21,21 @@
...
+public abstract class AbstractRepository<T> {
   protected final PostgresClientFuturized postgresClientFuturized;
   protected final PostgresClient postgresClient;
   protected final String tableName;
   protected final Class<T> recordType;
+  protected final String tenantId;
 
   protected AbstractRepository(PostgresClient postgresClient, String tableName,
-    Class<T> recordType) {
+    Class<T> recordType, String tenantId) {
```
hjiebsco, line 32: Why do need `tenantId`? Should not the `postgresClient` know about tenant already?

src/main/java/org/folio/persist/AbstractRepository.java
```diff
@@ -72,4 +74,11 @@ protected AbstractRepository(PostgresClient postgresClient, String tableName,
   public Future<RowSet<Row>> update(SQLConnection connection, String id, T record) {
     return update(succeededFuture(connection), id, record);
   }
+
+  public Future<RowSet<Row>> deleteAll() {
+    final String removeAllQuery = format("DELETE FROM %s_mod_inventory_storage.%s",
+      tenantId, tableName);
+
+    return postgresClientFuturized.execute(removeAllQuery);
+  }
```
hjiebsco, line 83: The implementation of this method seems a bit odd (and risky?). Can't `postgresClient` do that?

src/main/java/org/folio/persist/HoldingsRepository.java
```diff
@@ -2,20 +2,23 @@
...
   public HoldingsRepository(Context context, Map<String, String> okapiHeaders) {
-    this(postgresClient(context, okapiHeaders));
+    super(postgresClient(context, okapiHeaders), HOLDINGS_RECORD_TABLE, HoldingsRecord.class,
+      tenantId(okapiHeaders));
   }
 
-  public HoldingsRepository(PostgresClient postgresClient) {
-    super(postgresClient, HOLDINGS_RECORD_TABLE, HoldingsRecord.class);
+  public Future<List<HoldingsRecord>> getAll() {
+    return postgresClientFuturized.get(tableName, new HoldingsRecord());
```
hjiebsco, line 22: Can this actually return all holding records? How about performance for large tenant?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+import org.apache.commons.lang3.tuple.Triple;
+import org.apache.logging.log4j.Logger;
+import org.folio.persist.AbstractRepository;
+import org.folio.services.batch.BatchOperationContext;
+
+import io.vertx.core.Future;
+
+abstract class AbstractDomainEventPublisher<DomainType, EventType> {
```
hjiebsco, line 27: Maybe later `InstanceDomainEventPublisher` will be refactored to extend this for consistence?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -0,0 +1,136 @@
...
+    this.repository = repository;
+    this.domainEventService = domainEventService;
+  }
+
+  public Function<Response, Future<Response>> publishUpdated(DomainType oldRecord) {
+    return response -> {
+      if (!isUpdateSuccessResponse(response)) {
+        log.warn("Record update failed, skipping event publishing");
+        return succeededFuture(response);
+      }
```
hjiebsco, line 45: Just a `warn` in the log?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -0,0 +1,136 @@
...
+
+      return publishUpdated(singletonList(oldRecord)).map(response);
+    };
+  }
+
+  public Function<Response, Future<Response>> publishCreated(DomainType record) {
+    return response -> {
+      if (!isCreateSuccessResponse(response)) {
+        log.warn("Item create failed, skipping event publishing");
```
hjiebsco, line 54: Use `Record` instead of `Item` to be consistent.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  public Function<Response, Future<Response>> publishCreatedOrUpdated(
+    BatchOperationContext<DomainType> batchOperation) {
+
+    return response -> {
+      if (!isCreateSuccessResponse(response)) {
+        log.warn("Records create/update failed, skipping event publishing");
+        return succeededFuture(response);
+      }
```
hjiebsco, line 69: Same here. Is a `warn` in the log enough?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -0,0 +1,136 @@
...
+        .map(response);
+    };
+  }
+
+  public Function<Response, Future<Response>> publishRemoved(DomainType record) {
+    return response -> {
+      if (!isDeleteSuccessResponse(response)) {
+        log.warn("Record removal failed, no event will be sent");
+        return succeededFuture(response);
+      }
```
hjiebsco, line 86: Same here. Is a `warn` in the log enough?

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
```diff
@@ -25,20 +24,31 @@
 public final class FakeKafkaConsumer {
   private final static Map<String, List<KafkaMessage<JsonObject>>> itemEvents = new ConcurrentHashMap<>();
   private final static Map<String, List<KafkaMessage<JsonObject>>> instanceEvents = new ConcurrentHashMap<>();
+  private final static Map<String, List<KafkaMessage<JsonObject>>> holdingsEvents = new ConcurrentHashMap<>();
 
   public final FakeKafkaConsumer consume(Vertx vertx) {
     final KafkaConsumer<String, String> consumer = create(vertx, consumerProperties());
 
-    consumer.subscribe(Set.of("inventory.instance", "inventory.item"));
+    consumer.subscribe(Set.of("inventory.instance", "inventory.item", "inventory.holdings-record"));
```
hjiebsco, line 32: Consider to use the `enum KafkaTopic getTopicName()` instead of hard-coding.

src/test/java/org/folio/services/kafka/topic/KafkaAdminClientServiceTest.java
```diff
@@ -27,7 +27,8 @@
 
 @RunWith(VertxUnitRunner.class)
 public class KafkaAdminClientServiceTest {
-  private final Set<String> allTopics = Set.of("inventory.instance", "inventory.item");
+  private final Set<String> allTopics = Set.of("inventory.instance", "inventory.item",
+    "inventory.holdings-record");
```
hjiebsco, line 31: Consider to use the `enum KafkaTopic getTopicName()` instead of hard-coding.

hjiebsco (commented):
Left some inline questions and comments. Thx.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
bohdan-suprun: Sure, I've created https://issues.folio.org/browse/MODINVSTOR-658.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
bohdan-suprun: At this point the response should forward error to the client and RMB should log all exceptions, the only think that we can is to log a warn that event sending is skipped and return the response so it will be sent to the client. 

What do you think?

src/main/java/org/folio/persist/AbstractRepository.java
bohdan-suprun: It does now, however the method that returns tenant name has package only access. I've introduced an utility class that is in the same package as `PostgresClient`. Let me know if it is not fine.

src/main/java/org/folio/persist/AbstractRepository.java
bohdan-suprun: Changed to `postgresClient.delete(tableName, new Criterion())`

src/main/java/org/folio/persist/AbstractRepository.java
bohdan-suprun: This is now removed, because a delete method used from postgresClient.

src/main/java/org/folio/persist/HoldingsRepository.java
bohdan-suprun: That's a good question, actually this method is used only when a deleteAll endpoint is called for instances/hildings/items (in this case we need to issue domain events for all the records). 

The delete all endpoint is really dangerous itself, and I'm not sure whether it is used by anyone except unit tests, maybe @marcjohnson-kint has an answer. My personal preference would be to mark the endpoints deprecated and remove them  at all. But if it is not an option, should we set a limit (e.g. 1000 records for the method)?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
bohdan-suprun: Fixed, thanks.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
bohdan-suprun: Replaced with the enum constants.

src/main/java/org/folio/persist/HoldingsRepository.java
marcjohnson-kint: > The delete all endpoint is really dangerous itself

Why do you consider it dangerous?

> and I'm not sure whether it is used by anyone except unit tests, maybe @marcjohnson-kint has an answer. 

Alas, this came up [recently](https://folio-project.slack.com/archives/CFQU1MF61/p1610456907318300?thread_ts=1610455157.316900&cid=CFQU1MF61) and there are folks using them.

> My personal preference would be to mark the endpoints deprecated and remove them at all. But if it is not an option, should we set a limit (e.g. 1000 records for the method)?

Yeah, I would like to remove them. I regret defining them in the early days. 

However, I believe part of the scope of the forthcoming batch APIs is to include a delete by CQL endpoint, which would likely have similar challenges as it's possible to define CQL that matches all records.

bohdan-suprun: > Why do you consider it dangerous?
Because it is easy to drop the whole table using them... Someone just may forgot to specify an id and the delete all API will be executed.

bohdan-suprun: Hi @marcjohnson-kint, @hjiebsco, how do you think we should proceed with this? We can 1) Deprecate API; 2) Implement some sort of a batching removal. 3) an another idea?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
hjiebsco: By looking at the response, does the client know that saving to DB was a success but domain event publishing failed? Also, when this happens, db and elastic search will not be in-sync anymore. How do we handle this?

src/main/java/org/folio/persist/HoldingsRepository.java
hjiebsco: I was concerned about the `getAll` performance because it did not use streaming. I do not think that we have to get rid of `deleteAll` if it is needed. The risk can be controlled by permissions. As for passing all deleted records to Kafka/elastic search, we can use either streaming or maybe consider other approaches, for example, only pass necessary info (id) or a single delete all message to delete all related elastic search indexes.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
bohdan-suprun: `if (!isUpdateSuccessResponse(response)) {` means that save to DB is unsuccessful. 

If this happens that DB operation is success but event publish is not then the error will be propagated to the client, and probably clients have to handle it manually. 

We can handle it with transaction, but it will be a significant complication of the current code.

What do you think?

CC: @MikhailFokanov

src/main/java/org/folio/persist/HoldingsRepository.java
bohdan-suprun: I afraid that streaming won't help us here as well. We have to fetch all records **before** executing deleteAll and hold them in memory until receiving response from the deleteAll method (just for a case when we published events that records are removed but then DB operation is failed). Another concern, the streaming operation should lock records, as far as I know, so delete have to wait the lock to be released, there might be some issues with this.  

Probably we can fetch a portion of records (e.g. 1000) and then remove them by ids and then send events, repeat until any record exist.

The separate event for delete all action sounds really interesting, what do you think @MikhailFokanov?

marcjohnson-kint: > We can 1) Deprecate API;

I don't think deprecating the API will help as it will still be available.

> 2) Implement some sort of a batching removal. 

What do you mean by batching? Select a batch of records, delete them and then publish messages for those?

> 3) an another idea?

Publish a specific message for the deletion of all records instead of trying to publish a message for each record.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
hjiebsco: Got it. If db fails to save, there will be no event sent. The log statement is sufficient in this case. I agree it is a hard decision whether we should consider db save and elastic search update in the same transaction scope. If not, how do we make those in sync. Thanks.

marcjohnson-kint: >  I agree it is a hard decision whether we should consider db save and elastic search update in the same transaction scope.

They really can't be (even with the major pain of distributed transactions), there will likely always be a way for compensating actions could not happen.

src/main/java/org/folio/persist/HoldingsRepository.java
MikhailFokanov: > Publish a specific message for the deletion of all records instead of trying to publish a message for each record.

@marc-johnson, it will break the Domain Event pattern. If item is deleted, all consumers should know, that it is deleted. And also it would be redundant complexity of supporting additional logic in all modules, which consume to this topic to handle this edge case, instead of just sending the standard messages.

@marc-johnson, As I got from the conversation in slack, that you mentioned: this endpoint is used by "one or two data migration folks". So maybe it is not the place for spending efforts on complicated optimizations. Actually, I don't see any business reason, why the migration folks cannot wait a bit longer for this particular request to be executed.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
MikhailFokanov: My take on it was that if someone (or some service) receive an error response due to failing of saving to the database, he (it) will manage to do appropriate actions (it is how it works (should work) right now).
If the database saving fails, the message won't be sent. So ES and DB are in sync.
If saving to database is succeed, and for some misty reason (not the likely case, as Kafka is cluster and fault-tolerant) sending of message fails, user should get the same error. And manage to retry his request (as it is for the first case). So ES and DB should be also in sync. AFAIK, for the inventory the writing of the same data to the database don't break anything (during the retry).

src/main/java/org/folio/persist/HoldingsRepository.java
marcjohnson-kint: > it will break the Domain Event pattern. If item is deleted, all consumers should know, that it is deleted.

If a single item is deleted why should *all* consumers know about it? I thought the current design meant that only the consumer in the group that had the partition for the related instance received the message. Have I misunderstood?

Or by consumers are you referring to different modules rather than different consumers in a consumer group?

> @marc-johnson, As I got from the conversation in slack, that you mentioned: this endpoint is used by "one or two data migration folks". So maybe it is not the place for spending efforts on complicated optimizations. Actually, I don't see any business reason, why the migration folks cannot wait a bit longer for this particular request to be executed.

I'm not sure it's appropriate for me to comment on this. I think it is the Product Owners who decide whether there is a business reason for something not us. Whether they can tolerate a longer latency is a question to ask them.

hjiebsco: For record deletion, the current standard message is to publish the full old record content, do we have use case where the consumer actually needs the full content, or maybe the id would be enough?
As for the "delete all" message, is not it a regular message in a message driven system where the publisher emits, and consumer can decide what type of messages to consume?

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
hjiebsco: The issue is that the caller can put system out of sync. It should be the microservice's responsibility to maintain its own consistent state. The retry (if caller is responsible) might work for update, but not sure about create/delete.

marcjohnson-kint: > If saving to database is succeed, and for some misty reason (not the likely case, as Kafka is cluster and fault-tolerant) sending of message fails, user should get the same error. And manage to retry his request (as it is for the first case).

This suggests to me that we are relying on the client to retry in order for the system to become consistent. 

Am I interpreting this correctly?

bohdan-suprun: Hi guys, I think it is a general FOLIO challenge rather then an issue for this particular case. There are at least few more APIs that can easily break DB in inconsistent state: for example circulation check-in/out, requests, data import (I imagine) etc. 

So we probably need a community agreement for such a problem. If it you still think that it is critical problem, than we should investigate if we are able to use DB transactions for this (actually we can, but RMB has some limited toolset for this: for example there is no getById method with connection - it is needed to read records within the same transaction before commit. And this limited toolset requires introducing some custom methods in mod-inventory - which should be very well tested).

What do you think?

src/main/java/org/folio/persist/HoldingsRepository.java
bohdan-suprun: An Id should be enough, but for Items we have to also fetch related holdings-records to get instance id because it is partition key for the topic. 

But mod-search is not the only consumer, mod-remote-storage is an another consumer, probably they need locationId of the item as well, am I correct @khandramai?

hjiebsco: https://issues.folio.org/browse/MODINVSTOR-664

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
hjiebsco: https://issues.folio.org/browse/MODINVSTOR-663

hjiebsco (approved):
Open questions are postponed to new JIRAs
MODINVSTOR-663 - Log failed domain event sending to an internal DB table
MODINVSTOR-664 - Publish single deleteAll message for items/holdings/instances

PR #570, opened 2021-01-21
MODINVSTOR-656 enable detection only OL for a few tables

hjiebsco:
Please see https://issues.folio.org/browse/MODINVSTOR-656

- Added _version to json schema
- Added logOnConflict to db schema
- Fixed unit test failures
- TBD about version bump up
- TBD if the _version should be marked as read-only


src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
```diff
@@ -39,7 +39,7 @@ private static void assertCreateEvent(KafkaMessage<JsonObject> createEvent, Json
     assertThat(createEvent.getPayload().getString("type"), is("CREATE"));
     assertThat(createEvent.getPayload().getString("tenant"), is(TENANT_ID));
     assertThat(createEvent.getPayload().getJsonObject("old"), nullValue());
-    assertThat(createEvent.getPayload().getJsonObject("new"), is(newRecord));
+    assertThat(createEvent.getPayload().getJsonObject("new").putNull("_version"), is(newRecord.putNull("_version")));
```
julianladisch, line 42: assertThat(createEvent.getPayload().getJsonObject("new").remove("_version"), is(newRecord.remove("_version")));

ramls/holdingsrecord.json
```diff
@@ -8,6 +8,10 @@
       "description": "the unique ID of the holdings record; UUID",
       "$ref": "uuid.json"
     },
+    "_version": {
+      "type": "integer",
+      "description": "Record version for optimistic locking"
+    },
```
marcjohnson-kint, line 14: Is this property intended to be read only (from the client's perspective)?

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: Why is there a need to change the JSON being compared during the assertions?

ramls/holdingsrecord.json
hjiebsco: There is no agreement on that. In common scenarios, if client reads a record from db and that record already has _version field, client should pass that back to avoid OL conflict.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
hjiebsco: > ```
> assertThat(createEvent.getPayload().getJsonObject("new").remove("_version"), is(newRecord.remove("_version")));
> ```

That won't work because "remove" returns the object that is removed or null.

hjiebsco: > Why is there a need to change the JSON being compared during the assertions?

Because mod-inventory-storage does not read after save.

marcjohnson-kint: > Because mod-inventory-storage does not read after save.

Is that because we've chosen to use triggers to do the version incrementing and so the code doesn't know about it unless it reads the record from the database after the write?

ramls/holdingsrecord.json
marcjohnson-kint: > There is no agreement on that. 

I don't think I follow that. Does the protocol not rely on an agreement on that?

> In common scenarios, if client reads a record from db and that record already has _version field, client should pass that back to avoid OL conflict.

Maybe we are talking about {{read only}} as different ideas. I'm referring to a client not being allowed to change it. Are you referring to whether a client needs to provide it to the server?

hjiebsco: I am taking both ideas into consideration for this discussion. Looks like we are on the same page that clients will pass back _version. As for whether client can change it or not (or initialize _version with value 1 during bootstrap), I do not know if it was decided or not.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
hjiebsco: Correct

ramls/holdingsrecord.json
marcjohnson-kint: >  As for whether client can change it or not (or initialize _version with value 1 during bootstrap), I do not know if it was decided or not.

I don't understand how this work can be done without making that decision.

As I understand it, in order to log optimistic concurrency violations, the client needs to be involved and there needs to be agreement on when the version will change. Won't a client changing the version impede the detection process?

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: Have we checked with the folks using these messages if version is important (and needs to be provided in them)?

For example, if the instance representation provided by mod-search (one of the clients) will eventually be used for updating instances in ui-inventory then if it does not have the version, the front end won't have it to provide to the server during a change.

hjiebsco: > Have we checked with the folks using these messages if version is important to them (and hence needs to be provided in them)?
> 
> For example, if the instance representation provided by mod-search (one of the clients) will eventually be used for updating instances in ui-inventory then if it does not have the version, the front end won't have it to provide to the server during a change.

Good point. Hopefully relevant backend dev will comment.

ramls/holdingsrecord.json
hjiebsco: > > As for whether client can change it or not (or initialize _version with value 1 during bootstrap), I do not know if it was decided or not.
> 
> I don't understand how this work can be done without making that decision.
> 
> As I understand it, in order to log optimistic concurrency violations, the client needs to be involved and there needs to be agreement on when the version will change. Won't a client changing the version impede the detection process?

Are you suggesting that we should mark this field as read-only? We did that to metadata and that prevented us to put the new version property to metadata. @skoczko and @julianladisch, what do you think?

marcjohnson-kint: bq. Are you suggesting that we should mark this field as read-only? We did that to metadata and that prevented us to put the new version property to metadata.

I think we need to a way to communicate to the client what they are expected to do with this property. I was wondering if that might fit read only.

As I understand it, the concerns around putting this in the `metadata` property were that
* many of the clients in FOLIO throw this away and do not provide it back to the server
* RAML Module Builder throws away the metadata property before module implementation code gets access to it

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: @bohdan-suprun @MikhailFokanov Do you have thoughts on this?

bohdan-suprun: I have created https://issues.folio.org/browse/MODINVSTOR-662

marcjohnson-kint: @bohdan-suprun Which means you and @MikhailFokanov consider this behaviour to be acceptable for the moment?

bohdan-suprun: Yes, correct.

ramls/holdingsrecord.json
julianladisch: "_version" is not read-only because the client must not remove this property when sending the record to the PUT API.
The semantics of the "_version" property is that the client should never change it.
If optimistic locking reports an update conflict the client needs to read the complete record again with the more recent "_version" property and the new data that some other process has put in. The client merges its own updates into the new record and tries the update again.
The client should not guess the "_version" property value that makes the update pass the optimistic locking check. The "_version" property might have been incremented multiple times: There might be multiple other processes that have updated the record, or one process that uses multiple SQL statements to update a single record.

marcjohnson-kint: Thanks @julianladisch for sharing your perspective.

> "_version" is not read-only because the client must not remove this property when sending the record to the PUT API.
> The semantics of the "_version" property is that the client should never change it.

I'm basing my understanding of a read only property on the following (from the [latest JSON schema draft](), which admittedly FOLIO does not support):
> If "readOnly" has a value of boolean true, it indicates that the value of the instance is managed exclusively by the owning authority, and attempts by an application to modify the value of this property are expected to be ignored or rejected by that owning authority.

To me, that is consistent with your statement about the client never changing it.

Are you suggesting that the need for the client to provide the property in requests makes it not read only?

> If optimistic locking reports an update conflict the client needs to read the complete record again with the more recent "_version" property and the new data that some other process has put in. The client merges its own updates into the new record and tries the update again.

How is FOLIO intending to convey that expectation to clients / authors of clients?

> The client should not guess the "_version" property value that makes the update pass the optimistic locking check. 

The server cannot stop the client from guessing the version.

> The "_version" property might have been incremented multiple times: There might be multiple other processes that have updated the record, or one process that uses multiple SQL statements to update a single record.

I would hope we would try to avoid that, as it leaks internal behaviour outside of the server.

hjiebsco: Maybe we can update the description to make it clear about its usage. Anyway, this thread is getting long. Let's use JIRA https://issues.folio.org/browse/MODINVSTOR-656 to continue the discussion.

ramls/holdingsrecord.json
```diff
@@ -8,6 +8,10 @@
       "description": "the unique ID of the holdings record; UUID",
       "$ref": "uuid.json"
     },
+    "_version": {
```
bohdan-suprun, line 11: Hi @hjiebsco, Since the schemas are changed, could you please bump up API versions (`holdings-records-storage`, `item-storage`, `instance-storage`) and also version in the related RAML file?

bohdan-suprun (approved):

ramls/holdingsrecord.json
marcjohnson-kint: @hjiebsco Please don't merge this pull request (even though it has already been approved) until this has been done.

skoczko: @hjiebsco I agree that this should be a minor API change.

hjiebsco: Thanks guys. I bumped up the minor version for above three interfaces. Note, `instance-storage.raml` has incorrect version `7.4`, so I changed it to new correct one `7.6`. Just a FYI.

marcjohnson-kint: Thanks @hjiebsco (alas, that happens often, folks forget to change the interface definition itself)

PR #571, opened 2021-02-01
Use RAML with 409 as possible response MODINVSTOR-665

adamdickmeiss:
This updates both raml-util to include the item-collection with
409 as response. It also adds 409 as possible response for 3 sync
POST (item-sync, instance-sync, holdings-sync).

The raml-util submodule points to a commit that's not ideal.. So this PR should have one more commit where raml-util is updated to  HEAD of raml1.0 branch.


ramls/holdings-sync.raml
```diff
@@ -28,6 +28,11 @@ types:
     responses:
       201:
         description: "All holdings have been successfully created or updated"
+      409:
+        description: "Optimistic locking failed"
```
julianladisch, line 32: The locking successfully rejected the update.
```suggestion
        description: "Optimistic locking version conflict"
```

julianladisch (approved):
Looks good, thanks!

adamdickmeiss:
Thx for  looking. I'll make one more submodule commit when the raml1.0 is merged too.. It just looks nicer when the module uses a point on that branch

PR #572, opened 2021-02-02
MODINVSTOR-661: Allow generation of HRID values with and without leading zeroes

Aliaksandr-Fedasiuk:
## Purpose
Allow generation of HRID values with and without leading zeroes

## Approach
Include boolean flag "retainLeadingZeroes" that would whether leading zeros should be applied or not

## Learning
[MODINVSTOR-661](https://issues.folio.org/browse/MODINVSTOR-661)


OleksiiKuzminov (approved):

RuslanLavrov (approved):

VRohach (approved):

NEWS.md
```diff
@@ -2,6 +2,7 @@
 
 * Status date set on newly created instance records (MODINVSTOR-509)
 * Item status date is set for newly created item records (MODINVSTOR-508)
+* Allow generation of HRID values with and without leading zeroes (MODINVSTOR-661)
```
marcjohnson-kint, line 5: Thank you for adding a news entry that is meaningful from the user's perspective

ramls/hridsetting.json
```diff
@@ -14,6 +14,11 @@
       "type": "integer",
       "minimum": 1,
       "maximum": 99999999999
+    },
+    "retainLeadingZeroes" : {
+      "type": "boolean",
+      "description": "If true, leading zeros should be applied for HRIDs",
+      "default": true
```
marcjohnson-kint, line 21: A new property has been introduced, this should be reflected in a the minor version of the interface (both in the module descriptor and RAML)

NEWS.md
Aliaksandr-Fedasiuk: Thank you for review.

ramls/hridsetting.json
Aliaksandr-Fedasiuk: Completed. Thank you.

KaterynaSenchenko (approved):

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
```diff
@@ -478,6 +608,29 @@ public void canGetNextHridWhenStartNumberIsLong(TestContext testContext) {
...
+      .compose(v -> hridManager.getNextInstanceHrid())
+      .compose(hrid -> validateHrid(hrid, "9999999997", testContext))
+      .compose(v -> hridManager.getNextHoldingsHrid())
+      .compose(hrid -> validateHrid(hrid, "9999999998", testContext))
+      .compose(v -> hridManager.getNextItemHrid())
+      .compose(hrid -> validateHrid(hrid, "9999999999", testContext))
+      .onComplete(testContext.asyncAssertSuccess(
+        v1 -> log.info("Finished canGetNextHridWhenStartNumberIsLongWithoutLeadingZeroes()")));
+  }
+
```
marcjohnson-kint, line 633: Why is this test needed? Why is it significantly different to `canGetNextHridWhenStartNumberIsLong` above it?

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
```diff
@@ -423,38 +509,82 @@ public void canRollbackFailedTransaction(TestContext testContext) {
...
+            .onComplete(promise);
+        });
+        return promise.future();
+      })
+      .onComplete(testContext.asyncAssertSuccess(
+        v1 -> log.info("Finished canRollbackFailedTransaction()")));
+  }
+
+  @Test
+  public void canRollbackFailedTransactionWithoutLeadingZeroes(TestContext testContext) {
```
marcjohnson-kint, line 547: Why is this test needed? Why does not having leading zeros make it significantly different to `canRollbackFailedTransaction` above it?

marcjohnson-kint (commented):
If possible, please could significant reformatting of code be done in a separate pull request, it makes it easier for a reviewer to understand the meaningful changes to behaviour if they are separated.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
Aliaksandr-Fedasiuk: Hmm... I wrote these tests during investigation how HridManager works. These tests hasn't big sense.

mariia-aloshyna (approved):

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
Aliaksandr-Fedasiuk: I wrote these test also during investigation how HridManager works. These test also hasn't big sense.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
marcjohnson-kint: > These tests hasn't big sense.

Are you referring to the new tests?

ramls/hridsettings.json
```diff
@@ -24,6 +24,11 @@
       "description": "The HRID settings for items",
       "type": "object",
       "$ref": "hridsetting.json"
+    },
+    "commonRetainLeadingZeroes" : {
+      "type": "boolean",
+      "description": "If true, leading zeros should be applied for HRIDs",
+      "default": true
     }
```
marcjohnson-kint, line 32: Does this setting apply to all HRIDs irrespective of the record type?

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -152,6 +154,12 @@ public HridManager(Context context, PostgresClient postgresClient) {
     return promise.future().map(v -> null);
   }
 
+  private void updateRetainLeadingZeroes(HridSettings hridSettings) {
+    hridSettings.getInstances().setRetainLeadingZeroes(hridSettings.getCommonRetainLeadingZeroes());
+    hridSettings.getHoldings().setRetainLeadingZeroes(hridSettings.getCommonRetainLeadingZeroes());
+    hridSettings.getItems().setRetainLeadingZeroes(hridSettings.getCommonRetainLeadingZeroes());
+  }
```
marcjohnson-kint, line 161: This seems to overwrite the settings for each record with a single general setting. Have I understood this correctly?

Aliaksandr-Fedasiuk: Yes. Because I added one common parameter for UI and I added individual parameters for Instances/Holdings and Items. Now we will manage "Retain Leading Zeroes" by the one checkbox for all, but in the future we can to add individual checkbox for every group. 
In addition this way contains minimal number of changes in the module HridManager, by another way the number of changes that would be required would be much more.

ramls/hridsettings.json
Aliaksandr-Fedasiuk: Yes.  One parameter will be used for Instances/Holdings and Items (one checkbox on UI).

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
Aliaksandr-Fedasiuk: I mean that I have to delete this two tests because these tests not needed to testing my changes. New changes do not affect on Long Type overflow and on transaction rollback. I wrote them for investigating how HridManager works.

ramls/hridsettings.json
marcjohnson-kint: Then please remove the per record-type parameters that were added.

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: Thanks for clarifying. I think this logic makes those individual parameters redundant and introduces unnecessary complexity into the code.

Given that they have been superseded by a single parameter, please remove them and update the code / upgrade scripts accordingly.

Aliaksandr-Fedasiuk: I would prefer to leave the current implementation because all HridManager methods are designed to use separate HridSetting classes. Changing the code from using HridSetting to HridSetting**s** will require a lot of changes.
For an example, please look at the getNextHrid(final HridSetting hridSetting) method...) 
HridSetting is a main unit that provides us with the HridManager. Removing individual parameters will require serious changes in the HridManager module and tests. 
Maybe for this method I will add detailed comment for code clarifying.

marcjohnson-kint: > I would prefer to leave the current implementation because all HridManager methods are designed to use separate HridSetting classes. Changing the code from using HridSetting to HridSettings will require a lot of changes.

I understand that it would be more convenient to have these properties on those classes. Unfortunately, due to the way the code generation works, that is only possible by extending the public API.

I think it is confusing and misleading to change the public API for this to include properties whose values are not respected and not how the API is intended to be used.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
marcjohnson-kint: Thanks for explaining that. Does that mean those new tests will be removed?

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: @Aliaksandr-Fedasiuk Why was this comment marked as resolved? 

Is the resolution that it has been decided by the Folijet team to keep the changes to the public API?

Aliaksandr-Fedasiuk: This is solution is ok for our team because in the future we have plans to add such a property to every type separately. The refactoring of HridManager is redundant at this time.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
Aliaksandr-Fedasiuk: Yes. I will remove them.

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: > in the future we have plans to add such a property to every type separately.

If that is the case, why not only introduce the setting for each type in the back end API and have the front end set all of them from the same element in the form presented to the user?

By introducing the single setting into the back end API a change to settings per record type would require compatibility breaking changes in the future.

Aliaksandr-Fedasiuk: This solution is the fastest and requires minimal changes, and when switching to the settings for each record, we will only need to delete one field from the database (because the system does not support the Transient functionality) and this method.

src/test/java/org/folio/rest/api/HridSettingsStorageTest.java
Aliaksandr-Fedasiuk: Fixed.

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: > This solution is the fastest and requires minimal changes

By this solution, are you referring to introducing both record type specific settings *and* a general setting in the API?

> and when switching to the settings for each record, we will only need to delete one field from the database (because the system does not support the Transient functionality) and this method.

The removal of the shared setting property from the public API is a compatibility breaking change and would require synchronised changes which have caused challenges in the past. We try to avoid these where possible.

What do you mean by `transient functionality`?

Aliaksandr-Fedasiuk: About Transient. I meant that would be more comfortable to have some property in raml-json like "@Transient" in JPA to ignore a field from persisting.

marcjohnson-kint: > About Transient. I meant that would be more comfortable to have some property in raml-json like "@transient" in JPA to ignore a field from persisting.

Ok, thank you for explaining that more.

OleksiiKuzminov:
@marcjohnson-kint Hello! After long read I don’t understand what is the main concern? This PR cover needed functionality with possibility to extend for setting per entity with minimum code changes and no gaps from functionality side. Is it still blocked to merge and what kind of possible problems you see here?

marcjohnson-kint:
@OleksiiKuzminov 

> @marcjohnson-kint Hello! After long read I don’t understand what is the main concern? 

My primary concern is that this pull request introduces both a general and a record type specific configuration setting for leading zeros. 

The implementation makes the record type specific settings redundant as they are overwritten by the general configuration (for ease of implementation). I think this presents a misleading and confusing public API for clients. How would they know that the record level settings are ignored?

If the general settings is intended to be the only setting, I think the public API should reflect that. As should the database upgrade scripts which only update the record specific settings.

I don't understand why introduce both when one makes the other redundant.

Does that make sense?

Is the intention to have a single configuration entry or one per record type?

> Is it still blocked to merge and what kind of possible problems you see here?

I'm not blocking it from being merged, only providing feedback that I think this change introduces a confusing API and an inconsistent implementation. This work is owned by the Folijet team, it can be merged whenever the team is happy with it.

src/main/java/org/folio/rest/support/HridManager.java
bohdan-suprun: An equivalent for a transient should be readonly=true, I think. See `effectiveCallNumberComponents` in `item.json` as an example.

src/main/resources/templates/db_scripts/populateRetainLeadingZeroesSetting.sql
```diff
@@ -0,0 +1,23 @@
+START TRANSACTION;
+
+UPDATE ${myuniversity}_${mymodule}.hrid_settings
```
bohdan-suprun, line 3: Will it be possible to have a test for the script? You can use `org.folio.rest.api.ItemCopyNumberMigrationScriptTest` as an example

bohdan-suprun (approved):
Please create a test for the migration script or verify it manually.

src/main/java/org/folio/rest/support/HridManager.java
Aliaksandr-Fedasiuk: @bohdan-suprun, thank you for explanation.

src/main/resources/templates/db_scripts/populateRetainLeadingZeroesSetting.sql
Aliaksandr-Fedasiuk: Completed in RetainLeadingZeroesMigrationScriptTest.

ramls/hridsettings.json
Aliaksandr-Fedasiuk: Ready. Per record-type parameters were deleted.

src/main/java/org/folio/rest/support/HridManager.java
Aliaksandr-Fedasiuk: Completed. Per record-type parameters were deleted.

ramls/hridsettings.json
marcjohnson-kint: @Aliaksandr-Fedasiuk Thank you

ramls/hridsettings.json
```diff
@@ -24,6 +24,11 @@
       "description": "The HRID settings for items",
       "type": "object",
       "$ref": "hridsetting.json"
+    },
+    "commonRetainLeadingZeroes" : {
```
marcjohnson-kint, line 28: I think the location of this property already makes this property common across all record types and that makes the `common` prefix unnecessary.

Please feel free to ignore this if you don't agree with the feedback or want the explicit `common` part in the name.

```suggestion
    "retainLeadingZeroes" : {
```

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -235,4 +237,29 @@ private Void rollback(AsyncResult<SQLConnection> conn, Promise<Void> promise,
...
+  private interface Inventory<H> {
+    String getPrefix(HridSettings hridSettings);
+  }
+
+  private enum InventoryType implements Inventory<HridSettings> {
+    Holdings {
+      @Override
+      public String getPrefix(HridSettings hridSettings) {
+        return hridSettings.getHoldings().getPrefix();
+      }
```
marcjohnson-kint, line 250: This is a neat way of handling the variation between record types

PR #573, opened 2021-02-03
MODINVSTOR-627: add allTitles index

hjiebsco:
Please see https://issues.folio.org/browse/MODINVSTOR-627

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -467,6 +467,10 @@
           "multiFieldNames": "title,contributors[*].name,identifiers[*].value"
         },
         {
+          "fieldName": "allTitles",
+          "multiFieldNames": "title,indexTitle,alternativeTitles[*].alternativeTitle,series"
+        },
+        {
           "fieldName": "title"
         },
         {


adamdickmeiss (approved):
LGTM

marcjohnson-kint (commented):
Please add some tests that demonstrate the behaviour that this new index introduces

hjiebsco:
> Please add some tests that demonstrate the behaviour that this new index introduces

Why? This is a functionality provided by RMB.

marcjohnson-kint:
> > Please add some tests that demonstrate the behaviour that this new index introduces
> 
> Why? This is a functionality provided by RMB.

In order to demonstrate the behaviour. Otherwise, how do folks know that it works and what it does?

adamdickmeiss:
> > > Please add some tests that demonstrate the behaviour that this new index introduces
> > 
> > 
> > Why? This is a functionality provided by RMB.
> 
> In order to demonstrate the behaviour. Otherwise, how do folks know that it works and what it does?
> 
> How would a client know how to use it?

It's nice with tests. But do you expect the consumer, in this case Mike Taylor who's a UI guy to go through unit tests?

Two things that I think I've raised a number of times.. But I'll repeat anyway.

1. We should have a way for a module to "explain" what indices are supported and how they are used. Query support is a corner in the RAML spec but it's very important. In this case a "user" would know what is supported and how it's used.. Perhaps even an "example" tag in schema.json that could be reported.. We're talking a simple transform of schema.json through the web service. Not rocket science..
2. Reject queries not explicitly defined in schema.json. This has been rejected because the perception is that it will limit the usage of the product.. ON the contrary .. slow queries slows down the system as a whole and are better rejected up front fro a UI person when he/she are querying a database with 100 records and thinks all is good. Perhaps we could have a way to enable this.. Like strictQuery=query parameter.. Or stricit=true parameter .. Or even a CQL relation modifier.. Eg `title =/strict hensen`

marcjohnson-kint:
> It's nice with tests. But do you expect the consumer, in this case Mike Taylor who's a UI guy to go through unit tests?

I think the tests are mostly for the maintainers / contributors to the module. I use them to try to understand how the system works (or at least is intended to work).

I think they are available to folks if they want to use them. I might like it if folks did. I accept that folks on FOLIO mostly do not.

I think you raise an excellent question. How do folks, like Mike, and others (who haven't actively asked for this behaviour) know about it? How do they know how it is intended to behave?

> Two things that I think I've raised a number of times.. But I'll repeat anyway.
> 
> 1. We should have a way for a module to "explain" what indices are supported and how they are used. Query support is a corner in the interface spec as RAML can not capture it; but it's very important. In this case a "user" would know what is supported and how it's used.. Perhaps even an "example" tag in schema.json that could be reported.. We're talking a simple transform of schema.json through the web service. Not rocket science..

I totally agree with the need for documentation for CQL querying. FOLIO has not yet invested in this.

> 2. Reject queries not explicitly defined in schema.json. This has been rejected because the perception is that it will limit the usage of the product.. ON the contrary .. slow queries slows down the system as a whole and are better rejected up front fro a UI person when he/she are querying a database with 100 records and thinks all is good. Perhaps we could have a way to enable this.. Like strictQuery=query parameter.. Or stricit=true parameter .. Or even a CQL relation modifier.. Eg `title =/strict hensen`

I agree with making active choices about the queries the system supports. I don't believe the conversation on the issue for this was ever concluded, nor has FOLIO got any guidance for this at present AFAIK.

Given that we don't have either of those. How do we want to communicate the new behaviour that this change introduces?

adamdickmeiss:
> Given that we don't have either of those. How do we want to communicate the new behaviour that this change introduces?

We have JIRA issues to communicate about work and how it's implemented. The fix-version will tell when somebody can use the search. The NEWS file will either point to the JIRA ticket or include the title. I think that suffices.

marcjohnson-kint:
> > Given that we don't have either of those. How do we want to communicate the new behaviour that this change introduces?
> 
> We have JIRA issues to communicate about work and how it's implemented. The fix-version will tell when somebody can use the search. The NEWS file will either point to the JIRA ticket or include the title. I think that suffices.

We expect folks to trawl through JIRA issues to understand how the system behaves?

If the NEWS is only updated at release time, how does that help the developers using this new functionality prior to release?

marcjohnson-kint:
> > Given that we don't have either of those. How do we want to communicate the new behaviour that this change introduces?
> 
> We have JIRA issues to communicate about work and how it's implemented. The fix-version will tell when somebody can use the search. The NEWS file will either point to the JIRA ticket or include the title. I think that suffices.

I shall leave that decision to the Core Platform team who owns this work. I was only trying to share some feedback.

hjiebsco:
Per discussion of core platform, we do not need to add unit test for this new index. BTW the `instance` table also has another similar index `keyword` that was defined a while ago and is used by UI. Last but not the least, I did do a local test to verify that the new `allTitles` works. Some queries (decoded for convenience) I tried are listed below. They all return the modified `The Girl on the Train` record.
```
http://localhost:9130/inventory/instances?limit=100&query=(allTitles all "girl") sortby title
http://localhost:9130/inventory/instances?limit=100&query=(allTitles all "alternative") sortby title
http://localhost:9130/inventory/instances?limit=100&query=(allTitles all "indexTitleContent") sortby title
http://localhost:9130/inventory/instances?limit=100&query=(allTitles all "seriesA") sortby title

marcjohnson-kint:
> Per discussion of core platform, we do not need to add unit test for this new index. 

Ok :-/ Thank you for letting me know the decision.

> BTW the `instance` table also has another similar index `keyword` that was defined a while ago and is used by UI

It was, and I would've liked tests written for that too but hey ho :-/

julianladisch:
The `allIndexes` index is documented in the schema.json file: https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/schema.json
This file is linked from https://dev.folio.org/source-code/map/#mod-inventory-storage and that page is linked from https://dev.folio.org/reference/api/#mod-inventory-storage as "Documentation for mod-inventory-storage".

PR #574, opened 2021-02-09
MODINVSTOR-664: Publish single delete all event when all items/holdings/instances are removed.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-664 - Publish single deleteAll message for items/holdings/instances.

## Changes:
* Publish DELETE_ALL event when all instances/items/holdings are being removed.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -224,7 +224,7 @@ The event payload has following structure:
 {
   "old": {...}, // the instance/item before update or delete
   "new": {...}, // the instance/item after update or create
-  "type": "UPDATE|DELETE|CREATE", // type of the event
+  "type": "UPDATE|DELETE|CREATE|DELETE_ALL", // type of the event
   "tenant": "diku" // tenant name
 }
 ```
@@ -234,5 +234,28 @@ The event payload has following structure:
 Kafka partition key for all the events is instance id (for items it is retrieved from
 associated holding record).
 
-*Note for items:* the `new` and `old` records also includes `instanceId` property,
-on the same level with other item properties, which defined in the schema.
+## Domain events for items
+
+The `new` and `old` records also includes `instanceId` property,
+on the same level with other item properties, which defined in the schema:
+```javascript
+{
+  "instanceId": "<the instance id>",
+  // all other properties that defined in the schema
+}
+```
+
+## Domain events for delete all APIs
+
+There are delete all APIs for items instances and holding records. For such
+APIs we're issuing a special domain event:
+* Partition key: `00000000-0000-0000-0000-000000000000`
+* Event payload:
+```javascript
+{
+  "type": "DELETE_ALL",
+  "tenant": "<the tenant name>"
+  "old": null,
+  "new": null
+}
+```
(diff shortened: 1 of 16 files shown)


README.MD
```diff
@@ -234,5 +234,28 @@ The event payload has following structure:
...
+  "instanceId": "<the instance id>",
+  // all other properties that defined in the schema
+}
+```
+
+## Domain events for delete all APIs
+
+There are delete all APIs for items instances and holding records. For such
+APIs we're issuing a special domain event:
+* Partition key: `00000000-0000-0000-0000-000000000000`
```
marcjohnson-kint, line 252: What does this partition key do? 

I imagine Kafka still allocates it to a partition and only a single consumer in the group will process it. Is that a sensible understanding?

bohdan-suprun: > What does this partition key do?
This is a nil UUID, just to prevent NPEs for consumers. 

This event should be generated only once per tenant life, so it is fine to process it by a single consumer.

marcjohnson-kint: > This event should be generated only once per tenant life

How do we know this is likely to be the case?

marcjohnson-kint: Thanks. I think that means that my understanding that this message will be partitioned like other messages and processed by a single consumer is sensible.

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
```diff
@@ -129,6 +125,11 @@ public AbstractDomainEventPublisher(AbstractRepository<DomainType> repository,
   protected abstract Future<List<Pair<String, EventType>>> toInstanceIdEventTypePairs(
     Collection<DomainType> records);
 
+  private Future<Pair<String, EventType>> toInstanceIdEventTypePair(DomainType records) {
+    return toInstanceIdEventTypePairs(List.of(records))
+      .map(list -> list.get(0));
+  }
```
hjiebsco, line 131: Change `records` to `record`

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
```diff
@@ -24,19 +24,17 @@
...
 import java.util.stream.Collectors;
-
 import org.folio.services.kafka.KafkaMessage;
 
-import io.vertx.core.MultiMap;
-import io.vertx.core.json.JsonArray;
-import io.vertx.core.json.JsonObject;
-
 public final class DomainEventAssertions {
+  private static final String NULL_INSTANCE_ID = "00000000-0000-0000-0000-000000000000";
```
hjiebsco, line 37: This is also defined in `CommonDomainEventPublisher`. Consider to define it just once.

hjiebsco (approved):

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
bohdan-suprun: Done

src/main/java/org/folio/services/domainevent/AbstractDomainEventPublisher.java
bohdan-suprun: renamed

MikhailFokanov (approved):
Looks good.

PR #575, opened 2021-02-11
MODINVSTOR-381

AnShapovalov:
Store Shelving Order in item

### DIFF
--- a/ramls/item.json
+++ b/ramls/item.json
@@ -39,6 +39,10 @@
       "type": "string",
       "description": "Unique inventory control number for physical resources, used largely for circulation purposes"
     },
+    "shelvingOrder": {
+      "type": "string",
+      "description": "Shelving order automatically generated form Call Number (R/O for manually editing)"
+    },
     "itemLevelCallNumber": {
       "type": "string",
       "description": "Call Number is an identifier assigned to an item, usually printed on a label attached to the item. The call number is used to determine the items physical position in a shelving sequence, e.g. K1 .M44. The Item level call number, is the call number on item level."
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -4,7 +4,7 @@ import static io.vertx.core.Future.succeededFuture;
 import static org.folio.rest.support.EndpointFailureHandler.handleFailure;
 
 import java.util.Map;
-import java.util.Optional;
+
 
 import javax.ws.rs.core.Response;
 
@@ -12,7 +12,6 @@ import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.rest.jaxrs.resource.ItemStorage;
 import org.folio.rest.persist.PgUtil;
-import org.folio.services.CallNumberUtils;
 import org.folio.services.item.ItemService;
 
 import io.vertx.core.AsyncResult;
@@ -45,9 +44,6 @@ public class ItemStorageAPI implements ItemStorage {
     RoutingContext routingContext, Map<String, String> okapiHeaders,
     Handler<AsyncResult<Response>> asyncResultHandler,
     Context vertxContext) {
-
-    Optional<String> shelvingOrder = CallNumberUtils.getShelfKeyFromCallNumber(entity.getItemLevelCallNumber());
-    if (shelvingOrder.isPresent()) entity.setShelvingOrder(shelvingOrder.get());
 
     new ItemService(vertxContext, okapiHeaders).createItem(entity)
       .onSuccess(response -> asyncResultHandler.handle(succeededFuture(response)))
@@ -83,9 +79,6 @@ public class ItemStorageAPI implements ItemStorage {
     String itemId, String lang, Item entity, java.util.Map<String, String> okapiHeaders,
     io.vertx.core.Handler<io.vertx.core.AsyncResult<Response>> asyncResultHandler,
     Context vertxContext) {
-
-//    Optional<String> shelvingOrder = CallNumberUtils.getShelveNumber(entity.getEffectiveCallNumberComponents().getCallNumber());
-//    if (shelvingOrder.isPresent()) entity.setShelvingOrder(shelvingOrder.get());
 
     new ItemService(vertxContext, okapiHeaders).updateItem(itemId, entity)
       .onSuccess(response -> asyncResultHandler.handle(succeededFuture(response)))
--- a/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
+++ b/src/main/java/org/folio/rest/impl/ItemStorageAPI.java
@@ -4,6 +4,7 @@ import static io.vertx.core.Future.succeededFuture;
 import static org.folio.rest.support.EndpointFailureHandler.handleFailure;
 
 import java.util.Map;
+import java.util.Optional;
 
 import javax.ws.rs.core.Response;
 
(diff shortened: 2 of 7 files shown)


ramls/item.json
```diff
@@ -39,6 +39,10 @@
       "type": "string",
       "description": "Unique inventory control number for physical resources, used largely for circulation purposes"
     },
+    "shelvingOrder": {
```
marcjohnson-kint, line 42: As this is a derived property, I think we should follow the convention and use the `effective` prefix.

```suggestion
    "effectiveShelvingOrder": {
```

ramls/item.json
```diff
@@ -39,6 +39,10 @@
...
+      "type": "string",
```
marcjohnson-kint, line 43: As this is a derived property, I think it should be marked read only.

```suggestion
      "type": "string",
      "readonly": true,
```

src/main/java/org/folio/rest/impl/ItemStorageAPI.java
```diff
@@ -39,10 +40,10 @@ public void getItemStorageItems(
...
   @Override
   public void postItemStorageItems(
-      String lang, Item entity,
-      RoutingContext routingContext, Map<String, String> okapiHeaders,
-      Handler<AsyncResult<Response>> asyncResultHandler,
-      Context vertxContext) {
+    String lang, Item entity,
+    RoutingContext routingContext, Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
```
marcjohnson-kint, line 46: If possible, in future, please could reformatting be done in a separate pull request. It helps make it easier to distinguish the changes relating to the feature from other changes.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -6,12 +6,25 @@
...
+import org.folio.services.CallNumberUtils;
+
+import java.util.Optional;
 
 public final class EffectiveCallNumberComponentsUtil {
   private EffectiveCallNumberComponentsUtil() {}
 
   public static void setCallNumberComponents(Item item, HoldingsRecord hr) {
     item.setEffectiveCallNumberComponents(buildComponents(item, hr));
+    setShelvesNumber(item);
```
marcjohnson-kint, line 18: I think whether this belongs here depends upon whether the `shelving order` is of the effective call number components or separate.

If it is separate, I think this should be moved to it's own class.

ramls/item.json
```diff
@@ -39,6 +39,10 @@
...
+      "description": "Shelving order automatically generated form Call Number (R/O for manually editing)"
```
marcjohnson-kint, line 44: I suggest we use the description provided by Charlotte from [UIIN-816](https://issues.folio.org/browse/UIIN-816)

```suggestion
      "description": "A system generated normalization of the call number that allows for call number sorting in reports and search results"
```

src/main/java/org/folio/services/CallNumberUtils.java
```diff
@@ -0,0 +1,18 @@
...
+
+  private CallNumberUtils() {}
+
+  public static Optional<String> getShelfKeyFromCallNumber(String callNumberParam) {
+    if (callNumberParam == null) return Optional.empty();
+    CallNumber callNumber = new LCCallNumber();
+    callNumber.parse(callNumberParam);
+    return  callNumber.isValid() ? Optional.of(callNumber.getShelfKey()) : Optional.empty();
+  }
+}
```
marcjohnson-kint, line 18: Why is this separate to the effective call number components utility class?  Is the intention to have a class specific to generation to the shelving order?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -172,6 +179,7 @@ public void canCreateAnItemViaCollectionResource()
         is(inTransitServicePointId));
     assertThat(itemFromPost.getString("hrid"), is("it00000000001"));
     assertThat(itemFromPost.getString("copyNumber"), is("copy1"));
+    assertThat(itemFromPost.getString("shelvingOrder"), is("PS 43623 R534 P37 42005"));
```
marcjohnson-kint, line 182: Thank you for adding this check to the existing tests

src/main/java/org/folio/services/CallNumberUtils.java
AnShapovalov: At this time marc4j library implemented Library Of Congress Call Number Type. If you see at CallNumberType attribute of Item you could found that it has a bunch of predefined values. It means that we can meet with unpredictable development implementation of handlers for the other types of Call Number. As for me good idea to define separate class where any implementation could be integrated without affect any classes on more high level of logic.

ramls/item.json
AnShapovalov: Does it mean that we need new object "shelvingOrder" with single field "effectiveShelvingOrder" or field "effectiveShelvingOrder" should be placed in effective attributes block? In this case may be "itemLevelShelvingOrder" should be added too?

marcjohnson-kint: > Does it mean that we need new object "shelvingOrder" with single field "effectiveShelvingOrder" or field "effectiveShelvingOrder" should be placed in effective attributes block?

I only mean renaming the property that you have defined.

I don't think there needs to be a separate object. I believe the [domain folks have expressed](https://issues.folio.org/browse/MODINVSTOR-381?focusedCommentId=97850&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-97850) that the shelving order is not part of the call number components, so the property should not belong there.

Does that make sense?

> In this case may be "itemLevelShelvingOrder" should be added too?

I don't think there is a need for this prefix, as I don't believe there is the idea of a shelving order at a different level.

(I don't really know why that prefix was used in the previous work)

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
marcjohnson-kint: I believe the domain folks have now [answered](https://issues.folio.org/browse/MODINVSTOR-381?focusedCommentId=97850&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-97850) that question, the working out of shelving order should not be part of working out the call number components.

src/main/java/org/folio/services/CallNumberUtils.java
marcjohnson-kint: I think what you are describing is that this class is intended to hide the use of the external marc4j library from the rest of the system. Have I understood that correctly?

ramls/item.json
marcjohnson-kint: Have you already made this change?

src/main/java/org/folio/services/CallNumberUtils.java
AnShapovalov: Sure. Any other extention or library can be added without any changes in source code placed higer than library wrapper.

ramls/item.json
AnShapovalov: It looks like this :
 "effectiveShelvingOrder": {
      "type": "string",
      "description": "A system generated normalization of the call number that allows for call number sorting in reports and search results",
      "readonly": true
    },
    "itemLevelCallNumber": {
      "type": "string",
      "description": "Call Number is an identifier assigned to an item, usually printed on a label attached to the item. The call number is used to determine the items physical position in a shelving sequence, e.g. K1 .M44. The Item level call number, is the call number on item level."
    },

PR #576, opened 2021-02-12
MODINVSTOR-662: Publish instance/item/holdings from DB

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-662 - When instance/item/holding created send record from DB as domain event


hjiebsco (approved):

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2316,6 +2316,7 @@ private static JsonObject createItemRequest(
     itemToCreate.put("materialTypeId", materialType);
     itemToCreate.put("permanentLoanTypeId", canCirculateLoanTypeID);
     itemToCreate.put("temporaryLocationId", annexLibraryLocationId.toString());
+    itemToCreate.put("_version", 1);
```
marcjohnson-kint, line 2319: Why is the client specifying the version? I thought the server generated a version when none was provided by the client.

pfilippovEPAM (approved):

PR #577, opened 2021-02-16
Add identifier type to reference data (MODINVSTOR-636)

nielserik:
"Cancelled system control number"

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 ## 19.5.0 (in-progress)
 
+* Adds identifier type `Cancelled system control number` (MODINVSTOR-636)
 * Status date set on newly created instance records (MODINVSTOR-509)
 * Item status date is set for newly created item records (MODINVSTOR-508)
 * Allow generation of HRID values with and without leading zeroes (MODINVSTOR-661)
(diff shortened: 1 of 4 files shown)


marcjohnson-kint (commented):
As I understand it, systems already in production also need this new record. I believe this means the change also needs an upgrade script to create the record as usually reference records are not reloaded during upgrades.

adamdickmeiss:
I think it's cleaner to do it in loadData and do it regardless of parameters. Why?  It's just cleaner to keep things in files with the rest. With this PR it becomes even more obscure

nielserik:
> I think it's cleaner to do it in loadData and do it regardless of parameters. Why? It's just cleaner to keep things in files with the rest. With the second part of this PR it becomes even more obscure.

Right, I tend to agree.

It has to be done with care of course

- Only new records, that were added to `reference-data` since the currently installed release, should be inserted (since people may have deliberately decided to get rid of certain reference records)
- Only records that were not manually updated on the installed release since they first occurred in `reference-data`  should be updated (since people may have translated labels for example)  

I also wonder if all reference data are made equal and can be put under the same release update logic. Maybe they are and can, just wondering. 

For now, the SQL will serve its purpose of giving the POs the reference record in question, as I've been told they needed for their imminent work, and this should work whether or not said reference data update logic already is -- or will be -- in place for the time when the upgrade is rolled out for them.  

(the SQL does not itself honor the insert criteria to be sure; if people removed that record, it would keep getting added with each release until it's removed from schema.json)

marcjohnson-kint:
> I think it's cleaner to do it in loadData and do it regardless of parameters. Why? It's just cleaner to keep things in files with the rest. With the second part of this PR it becomes even more obscure.

I have no objection to this being done in java code. Upgrade scripts is merely how the Core Functional team have done it in the past.

PR #581, opened 2021-02-19
MODINVSTOR-673: Implement instance reindex API.

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-673 - Implement reindex endpoint

## Changes:
* Implement instance reindex API: POST - submit a job, DELETE - cancel a job, GET - get job status.
* Store the job statistic in DB table `reindex_job`;
* The reindexing logic is in [ReindexJobRunner](https://github.com/folio-org/mod-inventory-storage/pull/581/files#diff-00297561b0c586869185176b0c166b930ec1837b83b74a53e9f87c089969d11e).


hjiebsco (commented):
Nice work @bohdan-suprun. Since the job only sends instance ids to Kafka, what's the time difference between job complete status and the actual mod-search reindexing complete? I assume it will take some time for mod-search to query instance (and possible holdings/items) id by id and reindexing. Also, I noticed mod-search PR starts to add feature to query mod-inventory-storage, would that make it possible to implement reindex on mod-search side directly?

bohdan-suprun:
Hi @hjiebsco,

> Since the job only sends instance ids to Kafka, what's the time difference between job complete status and the actual mod-search reindexing complete? I assume it will take some time for mod-search to query instance (and possible holdings/items) id by id and reindexing. 

Yes, there might be some significant delays between when this job is finished and mod-search completed reindexing.  

> Also, I noticed mod-search PR starts to add feature to query mod-inventory-storage, would that make it possible to implement reindex on mod-search side directly?

Unfortunately no (at least for now). We have to send (stream) all instance IDs from inventory to search, if we use REST for this, it will be inefficient (or even impossible).

hjiebsco:
> Hi @hjiebsco,
> 
> > Since the job only sends instance ids to Kafka, what's the time difference between job complete status and the actual mod-search reindexing complete? I assume it will take some time for mod-search to query instance (and possible holdings/items) id by id and reindexing.
> 
> Yes, there might be some significant delays between when this job is finished and mod-search completed reindexing.
> 
> > Also, I noticed mod-search PR starts to add feature to query mod-inventory-storage, would that make it possible to implement reindex on mod-search side directly?
> 
> Unfortunately no (at least for now). We have to send (stream) all instance IDs from inventory to search, if we use REST for this, it will be inefficient (or even impossible).

1) It would be nice to do an experiment to see how much is the delay.
2) wouldn't mod-search pull the full record for each instance through REST? Also, if id first is a must (but why), I remember there is some API to get bulk instance ids.

bohdan-suprun:
Hi @hjiebsco,

> 1. It would be nice to do an experiment to see how much is the delay.

We have a story for e2e testing on the perf environment https://issues.folio.org/browse/MSEARCH-59.

> 2. wouldn't mod-search pull the full record for each instance through REST? 

It will, we're implementing a new API, that will return instance with it's holdings and items in one request. 

> Also, if id first is a must (but why), I remember there is some API to get bulk instance ids.

We do have such API. But the problem here is that network call is very unstable - what if the API call is broken in the middle of the streaming, then we don't have any ability to resume the streaming, so have to start the process from the scratch.

ramls/instance-reindex.raml
```diff
@@ -0,0 +1,66 @@
...
+          description: Reindex job with id not found
+          body:
+            text/plain:
+              example: Not found
+        500:
+          description: "Internal server error"
+          body:
+            text/plain:
+              example: "Internal server error"
+    delete:
```
pfilippovEPAM, line 53: description is missing for this method

bohdan-suprun: Thanks, fixed.

README.MD
```diff
@@ -255,7 +255,24 @@ APIs we're issuing a special domain event:
...
   "type": "DELETE_ALL",
   "tenant": "<the tenant name>"
-  "old": null,
-  "new": null
+}
+```
+
+## Reindex of instances
+
+Some consumers need to pull all instances from an existing database. There is
```
marcjohnson-kint, line 263: Which consumers need to do this? Re-index seems like a term specific to the search context. Is it being used here to mean something more general?

Under what circumstances does this need to happen, given that the Kafka topics already contain a complete history of changes?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1116,6 +1116,27 @@
...
+        {
+          "methods": ["GET"],
+          "pathPattern": "/instance-storage/reindex/{id}",
+          "permissionsRequired": ["inventory-storage.instance.reindex.item.get"]
+        },
+        {
+          "methods": ["DELETE"],
+          "pathPattern": "/instance-storage/reindex/{id}",
+          "permissionsRequired": ["inventory-storage.instance.reindex.item.delete"]
+        }
```
marcjohnson-kint, line 1138: Are these endpoints for management of a specific reindexing job?

hjiebsco:
> Hi @hjiebsco,
> 
> > 1. It would be nice to do an experiment to see how much is the delay.
> 
> We have a story for e2e testing on the perf environment https://issues.folio.org/browse/MSEARCH-59.
> 
> > 1. wouldn't mod-search pull the full record for each instance through REST?
> 
> It will, we're implementing a new API, that will return instance with it's holdings and items in one request.
> 
> > Also, if id first is a must (but why), I remember there is some API to get bulk instance ids.
> 
> We do have such API. But the problem here is that network call is very unstable - what if the API call is broken in the middle of the streaming, then we don't have any ability to resume the streaming, so have to start the process from the scratch.

1) If it takes mod-inventory-storage minutes to read and send millions instance id to Kafka, but it takes hours for mod-search to index those records, how do we cancel the job in the middle if it is desired?
2) if we add features in mod-search to query inventory, mod-search can get all instance id directly as well. So what's the benefit of sending millions of id messages through Kafka for reindexing all?

bohdan-suprun:
@hjiebsco,

> If it takes mod-inventory-storage minutes to read and send millions instance id to Kafka, but it takes hours for mod-search to index those records, how do we cancel the job in the middle if it is desired?

Hmm, good point. We probably need to propagate reindex job id somewhere in message headers and implement the same cancel endpoint but for mod-search. Then we can tell mod-search that a job with given id is cancelled and please ignore messages initiated by this reindex job. (CC: @pfilippovEPAM, @MikhailFokanov).

> if we add features in mod-search to query inventory, mod-search can get all instance id directly as well. So what's the benefit of sending millions of id messages through Kafka for reindexing all?

The main benefit is scalability and reliability. If we want to reindex inventory, then we may decide to scale up mod-search and deploy e.g. 10 instances, but with the REST approach these 10 instances won't be used effectively, because an instance should be elected and download all instance ids and start reindex (we can publish all ids to kafka ourselves - what actually the same approach that we have now but from another side of the dependency).

Another disadvantage of the downloading IDs, again it is unstable environment. The connection can be broken in the middle of download, and we have to retry the call from beginning.

hjiebsco:
> @hjiebsco,
> 
> > If it takes mod-inventory-storage minutes to read and send millions instance id to Kafka, but it takes hours for mod-search to index those records, how do we cancel the job in the middle if it is desired?
> 
> Hmm, good point. We probably need to propagate reindex job id somewhere in message headers and implement the same cancel endpoint but for mod-search. Then we can tell mod-search that a job with given id is cancelled and please ignore messages initiated by this reindex job. (CC: @pfilippovEPAM, @MikhailFokanov).
> 
> > if we add features in mod-search to query inventory, mod-search can get all instance id directly as well. So what's the benefit of sending millions of id messages through Kafka for reindexing all?
> 
> The main benefit is scalability and reliability. If we want to reindex inventory, then we may decide to scale up mod-search and deploy e.g. 10 instances, but with the REST approach these 10 instances won't be used effectively, because only one instance will process reindex - it should download all instance ids and start reindex (we can publish all ids to kafka ourselves - what actually the same approach that we have now but from another side of the dependency).
> 
> Another disadvantage of the downloading IDs, again it is unstable environment. The connection can be broken in the middle of download, and we have to retry the call from beginning.

for 2) mod-search also has a db backend, correct? Will all mod-search instances share the same db? If yes, the only difference here would be whether all mod-search instances read/store/process instance ids from Kafka (messages provided by other modules) or database (provided by mod-search itself).

bohdan-suprun:
> for 2) mod-search also has a db backend, correct? Will all mod-search instances share the same db? 

Yes, correct.

> If yes, the only difference here would be whether all mod-search instances read/store/process instance ids from Kafka (messages provided by other modules) or database (provided by mod-search itself).

For DB it will be much complicated - there should be some mechanism that distribute work among all available nodes. 

Let me know if I need to set up a meeting to discuss it?

hjiebsco:
> > for 2) mod-search also has a db backend, correct? Will all mod-search instances share the same db?
> 
> Yes, correct.
> 
> > If yes, the only difference here would be whether all mod-search instances read/store/process instance ids from Kafka (messages provided by other modules) or database (provided by mod-search itself).
> 
> For DB it will be much complicated - there should be some mechanism that distribute work among all available nodes.
> 
> Let me know if I need to set up a meeting to discuss it?

@bohdan-suprun, in the db case, I would imagine multiple mod-search instances (new or old) will simply just poll the same db table to see if there are more instance ids to reindex. BTW, I am just brainstorming ideas to help addressing the reindex status delay and cannot stop in the middle issues with current design/implementation. Please feel free to explore other options and/or maybe postpone to a different JIRA ticket.

bohdan-suprun:
Closing the PR for now, we're going to implement the reindex job on mod-search side.

bohdan-suprun:
Reopening as per discussion in slack.

descriptors/ModuleDescriptor-template.json
bohdan-suprun: yes

README.MD
bohdan-suprun: > Under what circumstances does this need to happen, given that the Kafka topics already contain a complete history of changes?

When DB already contains some instances and we want to add them to mod-search, or when a breaking change is applied to elasticsearch index structure

marcjohnson-kint: > When DB already contains some instances and we want to add them to mod-search

Is this going to be needed after mod-search is first enabled for it to have it's initial state?

> or when a breaking change is applied to elasticsearch index structure

In this case, why couldn't mod-search consume the original messages again?

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: What happens if more than one reindexing job runs at the same time? 

As I understand it, the re-index job will publish messages to the general `inventory.instances` topic, do all of the clients know how to ignore these if they aren't interested or discriminate messages for a particular job if they process them?

src/main/java/org/folio/services/domainevent/InstanceDomainEventPublisher.java
```diff
@@ -28,6 +28,7 @@
 
 public class InstanceDomainEventPublisher {
   private static final Logger log = getLogger(InstanceDomainEventPublisher.class);
+  private static final String JOB_ID_HEADER = "reindex-job-id";
```
hjiebsco, line 31: Consider renaming `JOB_ID_HEADER` to be more specific, for example `REINDEX_JOB_ID_HEADER`.

src/test/java/org/folio/rest/api/ReindexJobRunnerTest.java
```diff
@@ -75,6 +76,9 @@ public void canReindexInstances() {
     var lastInstanceEvent = getLastInstanceEvent(instanceId);
     assertThat(lastInstanceEvent.getPayload().getString("type"), is("REINDEX"));
     assertThat(lastInstanceEvent.getPayload().getString("tenant"), is(TENANT_ID));
+
+    assertThat(lastInstanceEvent.getHeaders(), hasEntry(
+      "reindex-job-id", job.getId()));
```
hjiebsco, line 81: Consider to reuse a constant rather than hardcoding `reindex-job-id`

hjiebsco (commented):
Since this PR was reopened, looks like the only change is to add the reindexing job id to the Kafka message. Can you elaborate the plan (maybe on mod-search side?) to address previously raised issues?
- how to handle the potential significant status delay between mod-inventory-storage id read/publish and mod-search reindexing?
- how to cancel the potential long running reindexing job quickly when needed?

pfilippovEPAM (approved):

src/test/java/org/folio/rest/api/ReindexJobRunnerTest.java
bohdan-suprun: Done, thanks.

src/main/java/org/folio/services/domainevent/InstanceDomainEventPublisher.java
bohdan-suprun: Renamed to `REINDEX_JOB_ID_HEADER`

bohdan-suprun:
@hjiebsco,

> Since this PR was reopened, looks like the only change is to add the reindexing job id to the Kafka message. 

This is now done, see 4664121.

> how to handle the potential significant status delay between mod-inventory-storage id read/publish and mod-search reindexing?

Why do you think it is an issue? Because the job status on inventory side becomes `Completed` however the entire reindex is still in progress? Or because of lack of observability for the entire process? (Please note, that we decided to postpone the job observability till R2).

> how to cancel the potential long running reindexing job quickly when needed?

This should be done on mod-search side. There should be API that should allow to specify a JOB ID (the one generated by mod-inv-storage) and should be persisted, and kafka workers should check the table periodically and ignore those events that have the `reindex-job-id` header equal to the one in DB table. Please note, when the instance IDs are published, we won't be able to entirely cancel the kafka message receiving, we have to just ignore such messages (it still useful, since the kafka consumer will call mod-inventory's API, if the job cancelled than we won't execute the call).

hjiebsco:
> Why do you think it is an issue? Because the job status on inventory side becomes `Completed` however the entire reindex is still in progress? Or because of lack of observability for the entire process? (Please note, that we decided to postpone the job observability till R2).

The first. Isn't it still an issue? If yes, can we mention that limitation in this PR (readme for example), so users won't be confused? Sure, it can be done later, but we should have a plan/design for the whole issue, so we know this work will fit into that plan and is not just a temporary throw-away work.

> This should be done on mod-search side. There should be API that should allow to specify a JOB ID (the one generated by mod-inv-storage) and should be persisted, and kafka workers should check the table periodically and ignore those events that have the `reindex-job-id` header equal to the one in DB table. Please note, when the instance IDs are published, we won't be able to entirely cancel the kafka message receiving, we have to just ignore such messages (it still useful, since the kafka consumer will call mod-inventory's API, if the job cancelled than we won't execute the call).

Is this (design) documented somewhere, and approved by @MikhailFokanov?

bohdan-suprun:
> If yes, can we mention that limitation in this PR (readme for example), so users won't be confused? 

We can rename job status, to something like `Ids published`, what do you think? 

> Sure, it can be done later, but we should have a plan/design for the whole issue, so we know this work will fit into that plan and is not just a temporary throw-away work.

We don't have a design document for now for the observability feature. The idea was to implement the required minimum.

hjiebsco:
> We can rename job status, to something like `Ids published`, what do you think?

That seems reasonable, but would that be temporary?

hjiebsco (approved):
Approve due to the time constraint. This need to be for R1

bohdan-suprun:
> That seems reasonable, but would that be temporary?

I think yes, we should review it when we will be implementing observability.

PR #583, opened 2021-02-24
MODINVSTOR-674: Implement inventory view instances API

bohdan-suprun:
https://issues.folio.org/browse/MODINVSTOR-674 - Implement bulk instance endpoint

## Changes:
* Implement `inventory-view` API;
* Register new `instance_holdings_item_view` view.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1116,6 +1116,17 @@
           "permissionsRequired": ["inventory-storage.inventory-hierarchy.items-and-holdings.collection.post"]
         }
       ]
+    },
+    {
+      "id": "inventory-view",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/inventory-view/instances",
+          "permissionsRequired": ["inventory-storage.inventory-view.instances.collection.get"]
+        }
+      ]
     }
   ],
   "permissionSets": [
@@ -2117,6 +2128,11 @@
       "description": "delete individual holdings-source from storage"
     },
     {
+      "permissionName": "inventory-storage.inventory-view.instances.collection.get",
+      "displayName": "inventory view - get instances by query with their holdings and items",
+      "description": "get instances by query with their holdings and items"
+    },
+    {
       "permissionName": "inventory-storage.all",
       "displayName": "inventory storage module - all permissions",
       "description": "Entire set of permissions needed to use the inventory storage module",
@@ -2318,7 +2334,8 @@
         "inventory-storage.holdings-sources.put",
         "inventory-storage.holdings-sources.delete",
         "inventory-storage.inventory-hierarchy.updated-instances-ids.collection.get",
-        "inventory-storage.inventory-hierarchy.items-and-holdings.collection.post"
+        "inventory-storage.inventory-hierarchy.items-and-holdings.collection.post",
+        "inventory-storage.inventory-view.instances.collection.get"
       ]
     }
   ],
--- a/pom.xml
+++ b/pom.xml
@@ -145,7 +145,7 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <ramlfiles_path>${basedir}/ramls/</ramlfiles_path>
     <raml-module-builder-version>32.0.2</raml-module-builder-version>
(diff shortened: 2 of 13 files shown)


marcjohnson-kint (commented):
How is this different to the hierarchical view API?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1116,6 +1116,17 @@
...
+    {
+      "id": "inventory-view",
+      "version": "1.0",
+      "handlers": [
+        {
+          "methods": ["GET"],
+          "pathPattern": "/inventory-view/instances",
+          "permissionsRequired": ["inventory-storage.inventory-view.instances.collection.get"]
+        }
+      ]
```
marcjohnson-kint, line 1129: How is this different to the existing inventory hierarchy API?

hjiebsco:
> How is this different to the hierarchical view API?

@marc-johnson, looks like this one returns all fields from `instance` rather than just four fields (`instanceId, source, modeOfIssuance, natureOfContent`) defined by the [hierarchical view](https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql#L169-L179).

hjiebsco (approved):
LGTM. It would be nice if `InventoryViewTest` demonstrates returning more than one `instance` records. That is more likely to be case used by mod-search.

pfilippovEPAM (approved):
Looks good to me.

bohdan-suprun:
> LGTM. It would be nice if `InventoryViewTest` demonstrates returning more than one `instance` records. That is more likely to be case used by mod-search.

Hi @hjiebsco, I have created https://issues.folio.org/browse/MODINVSTOR-684 for this, so that we can close this story till end of the sprint.

marcjohnson-kint:
> > How is this different to the hierarchical view API?
> 
> @marc-johnson, looks like this one returns all fields from `instance` rather than just four fields (`instanceId, source, modeOfIssuance, natureOfContent`) defined by the [hierarchical view](https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql#L169-L179).

Thanks @hjiebsco 

I really should've been more explicit with my question.

Why was a new API introduced instead of extending an existing one that seems to serve the same purpose?

bohdan-suprun:
> > > How is this different to the hierarchical view API?
> > 
> > 
> > @marc-johnson, looks like this one returns all fields from `instance` rather than just four fields (`instanceId, source, modeOfIssuance, natureOfContent`) defined by the [hierarchical view](https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql#L169-L179).
> 
> Thanks @hjiebsco
> 
> I really should've been more explicit with my question.
> 
> Why was a new API introduced instead of extending an existing one that seems to serve the same purpose?

@marcjohnson-kint, the existing view SQL is too complex to update it especially before module releases. Also, I'm not sure about it's performance, because it retrieves some additional records that are not needed e.g. for search. 

We might want to merge this two APIs at some future. But I'd recommend to proceed with this new API for now, if that's fine.

marcjohnson-kint:
> > > > How is this different to the hierarchical view API?
> > > 
> > > 
> > > @marc-johnson, looks like this one returns all fields from `instance` rather than just four fields (`instanceId, source, modeOfIssuance, natureOfContent`) defined by the [hierarchical view](https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/resources/templates/db_scripts/inventory-hierarchy/createRecordsViewFunction.sql#L169-L179).
> > 
> > 
> > Thanks @hjiebsco
> > I really should've been more explicit with my question.
> > Why was a new API introduced instead of extending an existing one that seems to serve the same purpose?
> 
> the existing view SQL is too complex to update it especially before module releases. 

I understand the need for an urgent resolution given the timing.

That is a little concerning with respect to the long term maintenance of the module. Out of curiosity, why do you consider the existing view too complex to change?

> We might want to merge this two APIs at some future. But I'd recommend to proceed with this new API for now, if that's fine.

That isn't my decision. FOLIO devolved technical design decisions to individual teams, which makes this Falcon's decision.

bohdan-suprun:
@marcjohnson-kint, I'm going to merge the changes, because they are blockers for release. Feel free to leave comments here on in slack channel `#﻿elastic-search-backend-integration`

bohdan-suprun:
> Out of curiosity, why do you consider the existing view too complex to change?

The SQL file has already 400+ lines, which most likely is not covered enough with tests.

PR #588, opened 2021-03-09
MODINVSTOR-693: Return instances without holdings or (and) items.

bohdan-suprun:
Fix inventory-view API to return instances even if they don't have holdings and (or) items. Change the API response so that there is no instance property.

### DIFF
--- a/ramls/inventory-view-instance.json
+++ b/ramls/inventory-view-instance.json
@@ -1,26 +1,17 @@
 {
   "$schema": "http://json-schema.org/draft-04/schema#",
-  "description": "Instance with holdings and items (there will be all instance properties, they just omitted, see instance.json)",
+  "description": "Instance with holdings and items",
   "type": "object",
   "properties": {
-    "id": {
-      "type": "string",
-      "description": "Instance id"
-    },
-    "source": {
-      "type": "string",
-      "description": "The metadata source and its format of the underlying record to the instance record. (e.g. FOLIO if it's a record created in Inventory;  MARC if it's a MARC record created in MARCcat or EPKB if it's a record coming from eHoldings)"
-    },
-    "title": {
-      "type": "string",
-      "description": "The primary title (or label) associated with the resource"
-    },
-    "instanceTypeId": {
-      "type": "string",
-      "description": "UUID of the unique term for the resource type whether it's from the RDA content term list of locally defined",
+    "instanceId": {
+      "description": "Instance id",
       "$ref": "uuid.json"
     },
-    "holdings": {
+    "instance": {
+      "description": "Instance record",
+      "$ref": "instance.json"
+    },
+    "holdingsRecords": {
       "type": "array",
       "description": "Holdings records for the instance",
       "items": {
@@ -35,11 +26,6 @@
       }
     }
   },
-  "additionalProperties": true,
-  "required": [
-    "source",
-    "title",
-    "instanceTypeId"
-  ]
+  "additionalProperties": false,
+  "required": ["instanceId", "instance"]
 }
-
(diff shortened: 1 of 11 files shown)


ramls/inventory-view-instance.json
```diff
@@ -26,7 +22,6 @@
       }
     }
   },
-  "additionalProperties": false,
-  "required": ["instanceId", "instance"]
+  "additionalProperties": true
```
hjiebsco, line 25: Just a FYI that `instance.json` has some required fields.

src/test/java/org/folio/rest/api/InventoryViewTest.java
```diff
@@ -6,54 +6,138 @@
...
+    assertThat(getItemIds(secondInstance), matchesInAnyOrder(itemsForTwo));
+  }
+
+  @Test
+  public void shouldReturnInstanceEvenIfNoItems() {
+    var instanceOne = instancesClient.create(instance(randomUUID()));
+    var holdingForOne = createHolding(instanceOne.getId(), mainLibraryLocationId, null);
 
     var instanceTwo = instancesClient.create(instance(randomUUID()));
     var holdings = List.of(
```
hjiebsco, line 68: `holdingsForTwo` to be consistent?

hjiebsco (approved):

ramls/inventory-view-instance.json
bohdan-suprun: Thanks, I've added them to the schema and to the `required` section.

src/test/java/org/folio/rest/api/InventoryViewTest.java
bohdan-suprun: Yes, renamed.

ramls/inventory-view-instance.json
```diff
@@ -1,17 +1,26 @@
...
-      "description": "Instance record",
-      "$ref": "instance.json"
+    "title": {
+      "type": "string",
+      "description": "The primary title (or label) associated with the resource"
+    },
+    "instanceTypeId": {
+      "type": "string",
+      "description": "UUID of the unique term for the resource type whether it's from the RDA content term list of locally defined",
+      "$ref": "uuid.json"
```
marcjohnson-kint, line 21: These appear to be breaking compatibility changes, though the interface version has only been given a minor version change.

Have I misinterpreted these changes?

bohdan-suprun: Yes, it is a breaking change, however the API has been created few days ago and it should have the `0.1` version (but it has 1.0 because of a typo).

Should I change version to 0.2 instead?

marcjohnson-kint: @bohdan-suprun Thank you for checking my understanding.

The accidental declaration of a public API is unfortunate and does make this a more challenging decision.

I shall leave that decision to the Falcon team. 

The original work was feature work merged after the API freeze for 2021 R1, as I imagine this work is intended to be too.

bohdan-suprun: I'd better keep the previous schema, then, thanks!

marcjohnson-kint: @bohdan-suprun 

> I'd better keep the previous schema, then, thanks!

As I say, I suggest this be discussed by the Falcon team

hjiebsco: > I'd better keep the previous schema, then, thanks!

You mean the previous composite schema with `instanceid, instace, holdings, items`? I like that one too. Feels cleaner.

MikhailFokanov (approved):
Looks good. Thanks.

PR #591, opened 2021-03-11
MODINVSTOR-521: Determine shelving order for existing items during upgrade

oleksandr-yatsenko1:
## Purpose
Allow for items (that have LC call numbers) to be sorted by shelving order (based upon call number and related fields)

## Approach

1. During an upgrade determine and store the shelving order
2. This should be done in the same way as when holdings or item records are changed (please refer to [MODINVSTOR-381](https://issues.folio.org/browse/MODINVSTOR-381) for more information)
3. This should only happen the first time that the module is upgraded to this version or later - this should be done by checking that the from version provided to the tenant API is before version 19.5.0.
4. 
This process will likely take a long time for large quantities of items, a note about this should be added to the release notes for the distribution this change is included in (intended to be 2021 R1).

## Pre-Merge Checklist:
Before merging this PR, please go through the following list and take appropriate actions.

- Does this PR meet or exceed the expected quality standards?
  - [ ] Code coverage on new code is 80% or greater
  - [ ] Duplications on new code is 3% or less
  - [ ] There are no major code smells or security issues
- Does this introduce breaking changes?
  - [ ] Were any API paths or methods changed, added or removed?
  - [x ] Were there any schema changes?
  - [ ] Did any of the interface versions change?
  - [ ] Were permissions changed, added, or removed?
  - [ ] Are there new interface dependencies?
  - [ x] There are no breaking changes in this PR.
  
If there are breaking changes, please **STOP** and consider the following:

- What other modules will these changes impact?
- Do JIRAs exist to update the impacted modules?
  - [ ] If not, please create them
  - [ ] Do they contain the appropriate level of detail?  Which endpoints/schemas changed, etc.
  - [ ] Do they have all they appropriate links to blocked/related issues?
- Are the JIRAs under active development?  
  - [ ] If not, contact the project's PO and make sure they're aware of the urgency.
- Do PRs exist for these changes?
  - [ ] If so, have they been approved?

Ideally all of the PRs involved in breaking changes would be merged in the same day to avoid breaking the folio-testing environment.  Communication is paramount if that is to be achieved, especially as the number of intermodule and inter-team dependencies increase.  

While it's helpful for reviewers to help identify potential problems, ensuring that it's safe to merge is ultimately the responsibility of the PR assignee.

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -122,6 +122,6 @@ public class TenantRefAPI extends TenantAPI {
           }
           return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
         }).compose(loaded -> ShelvingOrderUpdate
-          .getInstance().updateItems(attributes, headers, vertxContext));
+          .getInstance().startUpdatingOfItems(attributes, headers, vertxContext));
   }
 }
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -9,6 +9,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.TenantAttributes;
+import org.folio.rest.support.ShelvingOrderUpdate;
 import org.folio.rest.tools.utils.TenantLoading;
 import org.folio.services.kafka.topic.KafkaAdminClientService;
 
@@ -120,6 +121,7 @@ public class TenantRefAPI extends TenantAPI {
               .add("users", "service-points-users");
           }
           return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
-        });
+        }).compose(loaded -> ShelvingOrderUpdate
+          .getInstance().updateItems(attributes, headers, vertxContext));
   }
 }
--- a/src/main/java/org/folio/rest/support/CompletableFutureUtil.java
+++ b/src/main/java/org/folio/rest/support/CompletableFutureUtil.java
@@ -1,12 +1,7 @@
 package org.folio.rest.support;
 
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
-
 import io.vertx.core.AsyncResult;
-import io.vertx.core.Future;
 import io.vertx.core.Handler;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -27,21 +22,6 @@ public final class CompletableFutureUtil {
         future.completeExceptionally(result.cause());
       }
     };
-  }
-
-  public static <T> T getFutureResult(Future<T> future) {
-    log.info("Started \"getFutureResult\", future: {} ...", future);
-    return getFutureResult(future.toCompletionStage().toCompletableFuture());
-  }
-
-  public static <T> T getFutureResult(CompletableFuture<T> completableFuture) {
-    log.info("Started \"getFutureResult\", completableFuture: {} ...", completableFuture);
-    try {
-      return completableFuture.get(TIMEOUT, TimeUnit.SECONDS);
-    } catch (InterruptedException | ExecutionException | TimeoutException e) {
-      log.error("Error occurs in getFutureResult: {}", e.getMessage());
-      throw new RuntimeException(e);
-    }
   }
 
 }
--- a/src/main/java/org/folio/rest/support/CompletableFutureUtil.java
+++ b/src/main/java/org/folio/rest/support/CompletableFutureUtil.java
@@ -1,11 +1,22 @@
 package org.folio.rest.support;
 
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import io.vertx.core.AsyncResult;
+import io.vertx.core.Future;
 import io.vertx.core.Handler;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public final class CompletableFutureUtil {
+
+  private static final Logger log = LogManager.getLogger();
+
+  public static final long TIMEOUT = 180;
+
   private CompletableFutureUtil() {}
 
   public static <T> Handler<AsyncResult<T>> mapFutureResultToJavaFuture(CompletableFuture<T> future) {
@@ -17,4 +28,20 @@ public final class CompletableFutureUtil {
       }
     };
   }
+
+  public static <T> T getFutureResult(Future<T> future) {
+    log.info("Started \"getFutureResult\", future: {} ...", future);
+    return getFutureResult(future.toCompletionStage().toCompletableFuture());
+  }
+
+  public static <T> T getFutureResult(CompletableFuture<T> completableFuture) {
+    log.info("Started \"getFutureResult\", completableFuture: {} ...", completableFuture);
+    try {
+      return completableFuture.get(TIMEOUT, TimeUnit.SECONDS);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      log.error("Error occurs in getFutureResult: {}", e.getMessage());
+      throw new RuntimeException(e);
+    }
+  }
+
 }
--- a/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
+++ b/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
@@ -2,16 +2,15 @@ package org.folio.rest.support;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.services.CallNumberUtils;
-
-import java.util.Objects;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 public final class EffectiveCallNumberComponentsUtil {
 
@@ -19,6 +18,12 @@ public final class EffectiveCallNumberComponentsUtil {
 
   public static void setCallNumberComponents(Item item, HoldingsRecord hr) {
     item.setEffectiveCallNumberComponents(buildComponents(item, hr));
+  }
+
+  public static Item getCalculateAndSetEffectiveShelvingOrder(Item item) {
+    // Calculate item's shelving order
+    calculateAndSetEffectiveShelvingOrder(item);
+    return item;
   }
 
   public static void calculateAndSetEffectiveShelvingOrder(Item item) {
@@ -35,12 +40,11 @@ public final class EffectiveCallNumberComponentsUtil {
           .map(StringUtils::trim)
           .collect(Collectors.joining(" "))
       );
-//      String suffixValue = Objects.toString(item.getEffectiveCallNumberComponents().getSuffix(), "").trim();
-      String suffixValue =
-        Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
-          .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
-          .trim();
 
+      String suffixValue =
+          Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
+              .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
+              .trim();
       String nonNullableSuffixValue = suffixValue.isEmpty() ? "" : " " + suffixValue;
 
       item.setEffectiveShelvingOrder(
--- a/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
+++ b/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
@@ -2,15 +2,16 @@ package org.folio.rest.support;
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-import java.util.Objects;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
 import org.apache.commons.lang3.StringUtils;
 import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.services.CallNumberUtils;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 public final class EffectiveCallNumberComponentsUtil {
 
@@ -42,9 +43,10 @@ public final class EffectiveCallNumberComponentsUtil {
       );
 
       String suffixValue =
-          Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
-              .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
-              .trim();
+        Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
+          .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
+          .trim();
+
       String nonNullableSuffixValue = suffixValue.isEmpty() ? "" : " " + suffixValue;
 
       item.setEffectiveShelvingOrder(
--- a/src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
+++ b/src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
@@ -1,17 +1,24 @@
 package org.folio.rest.support;
 
-import io.vertx.core.AsyncResult;
-import io.vertx.core.Handler;
+import static org.folio.rest.support.CompletableFutureUtil.getFutureResult;
+
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.TimeZone;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.SerializationFeature;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
@@ -34,6 +41,21 @@ import org.folio.rest.persist.PostgresClient;
 
 public class ShelvingOrderUpdate {
   private static final Logger log = LogManager.getLogger();
+
+//  public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+  public static final String DATE_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSS+0000";
+
+  public static final ObjectMapper jsonMapper =
+      new ObjectMapper()
+          .enable(SerializationFeature.INDENT_OUTPUT)
+          .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
+          .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+
+  static {
+    SimpleDateFormat df = new SimpleDateFormat(DATE_FORMAT);
+    df.setTimeZone(TimeZone.getTimeZone("UTC"));
+    jsonMapper.setDateFormat(df);
+  }
 
   private static final String SQL_SELECT_ITEMS =
       new StringBuilder()
@@ -126,31 +148,29 @@ public class ShelvingOrderUpdate {
       PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, headers);
 
       final long startTime = System.currentTimeMillis();
-      fetchAndUpdatesItems(vertx, postgresClient)
-          .onSuccess(rowsAffected -> {
-           totalUpdatedRows.addAndGet(rowsAffected);
-            // If previous updates returns affected rows, try it again
-            if (rowsAffected > 0) {
-              fetchAndUpdatesItems(vertx, postgresClient);
-            } else {
-              // No rows affected, just set accumulated amount of rows to result
-              log.info("Items update with shelving order property completed in {} seconds",
-                  (System.currentTimeMillis() - startTime) / 1000);
-              updateItemsPromise.complete(totalUpdatedRows.get());
-//              completionHandlerPromise.complete((totalUpdatedRows.get() > 0));
-            }
-          })
-          .onFailure(h -> {
-            updateItemsPromise.fail(h.getCause());
-//            completionHandlerPromise.fail(h.getCause());
-            log.error("Error updating items: {}", h.getCause().getMessage());
-          });
+
+      int rowsAffected = 0;
+      do {
+        Future<Integer> affectedRowsFuture = fetchAndUpdatesItems(vertx, postgresClient);
+        rowsAffected = getFutureResult(affectedRowsFuture);
+
+        if (affectedRowsFuture.failed()) {
+          log.error("Error updating portion of items: {}", affectedRowsFuture.cause().getMessage());
+          updateItemsPromise.fail(affectedRowsFuture.cause());
+        } else {
+          log.info("Portion of items has been updated: affected rows amount: {}", rowsAffected);
+          totalUpdatedRows.addAndGet(rowsAffected);
+        }
+      } while (rowsAffected > 0);
+
+      log.info("Items updates with shelving order property completed in {} seconds",
+          (System.currentTimeMillis() - startTime) / 1000);
+      updateItemsPromise.complete(totalUpdatedRows.get());
 
     } else {
       log.info("Module isn't eligible for upgrade, just exit.");
       // Not allowed to perform items updates, just returns
       updateItemsPromise.complete(0);
-//      completionHandlerPromise.complete(Boolean.FALSE);
     }
 
     return updateItemsPromise.future();
@@ -196,6 +216,10 @@ public class ShelvingOrderUpdate {
 
                   // Updates of items
                   .compose(listOfTuple -> updateItems(listOfTuple, connection))
+                    .onFailure(h -> {
+                      log.error("Failed updateItems execution: {}, transaction will be rolled back/", h.getMessage());
+                      tx.rollback();
+                      })
 
                   // Finish transaction
                   .compose(updatedRowsCount -> completeTransaction(tx, updatedRowsCount, fetchAndUpdatePromise))
@@ -234,6 +258,27 @@ public class ShelvingOrderUpdate {
     return Future.succeededFuture();
   }
 
+  public <T> String toJsonString(T object) {
+    try {
+      return jsonMapper.writeValueAsString(object);
+    } catch (JsonProcessingException jpe) {
+      log.error("Error occurs when converting {} to JSON string", object);
+    }
+    return StringUtils.EMPTY;
+  }
+
+  public <T> T fromJsonString(String jsonString, Class<T> clazz) {
+    try {
+      T object = jsonMapper.readValue(jsonString, clazz);
+      return object;
+    } catch (JsonMappingException jme) {
+      log.error("Mapping error: {} occurs when converting {} from JSON string to {}", jme, jsonString, clazz);
+    } catch (JsonProcessingException jpe) {
+      log.error("Processing error: {} occurs when converting {} from JSON string to {}",jpe, jsonString, clazz);
+    }
+    return null;
+  }
+
   /**
    * Read items and fill them into data structures for further processing
    * 
@@ -248,7 +293,10 @@ public class ShelvingOrderUpdate {
       List<Item> targetList = new ArrayList<>();
       rowSet.forEach(row -> {
         JsonObject itemJsonObject = row.getJsonObject("itemJson");
-        Item item = itemJsonObject.mapTo(Item.class);
+        log.info("Read JSON from database as JsonObject : {}", itemJsonObject);
+        log.info("JonObject toString: {}", itemJsonObject.toString());
+        Item item = fromJsonString(itemJsonObject.toString(), Item.class);
+        log.info("Deserialized item which was converted to JSON string : {}", toJsonString(item));
         targetList.add(item);
       });
       promise.complete(targetList);
@@ -269,9 +317,15 @@ public class ShelvingOrderUpdate {
     log.info("Invoking of calculateShelvingOrder, itemList size: {}",
         Optional.ofNullable(itemList).orElse(new ArrayList<>()).size());
 
+    final ObjectMapper objectMapper = new ObjectMapper();
     Map<UUID, JsonObject> updatedItemsMap = itemList.stream()
         .map(EffectiveCallNumberComponentsUtil::getCalculateAndSetEffectiveShelvingOrder)
-        .collect(Collectors.toMap(item -> UUID.fromString(item.getId()), JsonObject::mapFrom));
+        .collect(Collectors.toMap(
+            item -> UUID.fromString(item.getId()),
+            item -> {
+              log.info("map's value, an item: {}", item);
+              return new JsonObject(toJsonString(item));
+            }));
 
     log.info("Finishing of calculateShelvingOrder, updatedItemsMap: {}", updatedItemsMap);
     return Future.succeededFuture(updatedItemsMap);
@@ -305,7 +359,7 @@ public class ShelvingOrderUpdate {
   public Future<Integer> updateItems(List<Tuple> itemsParams, PgConnection connection) {
     log.info("Invoking of updateItems, itemsParams size: {}", itemsParams.size());
     log.info("Items parameters for update SQL: {}",
-        StringUtils.defaultIfBlank(itemsParams.stream().map(Tuple::toString).collect(Collectors.joining(",")),
+        StringUtils.defaultIfBlank(itemsParams.stream().map(Tuple::deepToString).collect(Collectors.joining(",")),
             "none"));
 
     if (itemsParams.isEmpty()) {
--- a/src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
+++ b/src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
@@ -2,23 +2,17 @@ package org.folio.rest.support;
 
 import static org.folio.rest.support.CompletableFutureUtil.getFutureResult;
 
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
-import java.util.TimeZone;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonMappingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.SerializationFeature;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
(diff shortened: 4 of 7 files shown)


marcjohnson-kint:
@AnShapovalov and @K-Felk Please review this pull request

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -2,17 +2,16 @@
...
 import org.folio.rest.jaxrs.model.EffectiveCallNumberComponents;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.Item;
 import org.folio.services.CallNumberUtils;
 
-import java.util.Objects;
-import java.util.Optional;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-
```
marcjohnson-kint, line 15: This module typically uses [Eclipse import order|https://issues.folio.org/browse/FOLIO-1667]. Please configure your IDE to use that import ordering to minimise the amount of unnecessary changes to import statements during pull requests.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -35,12 +40,11 @@ public static void calculateAndSetEffectiveShelvingOrder(Item item) {
...
-//      String suffixValue = Objects.toString(item.getEffectiveCallNumberComponents().getSuffix(), "").trim();
-      String suffixValue =
-        Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
-          .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
-          .trim();
 
+      String suffixValue =
+          Objects.toString(Optional.ofNullable(item.getEffectiveCallNumberComponents())
+              .orElse(new EffectiveCallNumberComponents()).getSuffix(), "")
+              .trim();
```
marcjohnson-kint, line 47: Why has the indentation of these lines changed?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -126,6 +120,17 @@ String servicePointUserFilter(String s) {
...
+          long startTime = System.currentTimeMillis();
+          new ShelvingOrderUpdate(5000)
+              .updateItems(attributes, headers, vertxContext)
+              .onSuccess(h -> log.info("Shelving order updates completed in {} seconds",
+                  (System.currentTimeMillis() - startTime) / 1000))
+              .onFailure(
+                  h -> log.error("Error occurs during shelving order update: {}", h.getMessage()));
+      });
   }
+
```
marcjohnson-kint, line 135: Why has this blank line been introduced?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,186 @@
...
+      .append("LIMIT $1").toString();
+
+
+  private static final String SQL_UPDATE_ITEMS = new StringBuilder()
+      .append("UPDATE ITEM ")
+      .append("SET jsonb = $1 ")
+      .append("WHERE id = $2").toString();
+
+  // Defines version threshold where this property was introduces from
+  private static final String FROM_VERSION_SHELVING_ORDER = "19.5.0";
```
marcjohnson-kint, line 49: This now needs to be 20.1.0, as the initial release for 2021 R1 (20.0.0) has now been done

marcjohnson-kint (commented):
There needs to be some automated tests to demonstrate this behaviour.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,186 @@
...
+                  .onSuccess(h -> log.info("Items update transaction succeeded"))
+                  .onFailure(h -> log.error("Items update transaction filed: {}", h.getMessage())));
+
+        } // Connection acquired successfully
+
+      }); // Eof getConnection scope
+
+     }); // Eof vert.x timer scope
+
+    return promise.future();
```
marcjohnson-kint, line 183: When does this promise complete? I couldn't find any code that completes this promise.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,186 @@
...
+      }
+    });
+
+    return promise.future();
+  }
+
+  private Future<Integer> fetchAndUpdatesItems(Vertx vertx, PostgresClient postgresClient) {
+    Promise<Integer> promise = Promise.promise();
+
+    vertx.setTimer(500L, timer -> {
```
marcjohnson-kint, line 113: Why is a timer being used for these updates?

I think a timer introduces some additional challenges for this process. It means that multiple batches might be processed at the same time, which I think is counter-productive to the intent of using batches to maintain a consistent load. Alternatively, if each batch takes less than 500ms, it might mean the process is slower than it needs to be.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
oleksandr-yatsenko1: Yes, it looks I have eclipse formatter plugin turned off. I followed by link and only found import configs there attached ("Eclipse style imports.xml" and "folio.importorder " files). Could you please point where the "eclipse formatter config" could be taken?

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
oleksandr-yatsenko1: Seems it related disabled eclipse formatting plugin. I'll compare it with already merged code to make correction until formatting be fixed.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.commons.lang3.StringUtils;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.dbschema.Versioned;
+import org.folio.rest.jaxrs.model.Item;
+import org.folio.rest.jaxrs.model.TenantAttributes;
+import org.folio.rest.persist.PgUtil;
+import org.folio.rest.persist.PostgresClient;
+
+public class ShelvingOrderUpdate extends Versioned {
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: I thought that mentioned once functionality in the code [https://github.com/folio-org/raml-module-builder/blob/v32.2.0/dbschema/src/main/java/org/folio/dbschema/Versioned.java#L30] could be leveraged here. I tried to reuse it by considering this upgrade fictionality as a kind of "patching code" which needs to be applied at some of version. Possibly that's not the best approach and it's need just to leverage mentioned above method? But it has non-static nature, that why it looks it needs to be a part of entity with upgrade functionality what I tried to implement. Should it be simplified, how do you think?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Good point, thank you for recent module release, I'll update the version here.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
marcjohnson-kint: @oleksandr-yatsenko1 

> Could you please point where the "eclipse formatter config" could be taken?

I don't know what Eclipse formatter is. If it is an intelliJ plugin, I don't use it and cannot provide configuration. I believe some of the EPAM developers do use that plugin, maybe they can help you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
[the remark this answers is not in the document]
marcjohnson-kint: > Possibly that's not the best approach and it's need just to leverage mentioned above method? But it has non-static nature, that why it looks it needs to be a part of entity with upgrade functionality what I tried to implement. Should it be simplified, how do you think?

I did suggest the use of this class.  

Personally, I prefer composition over inheritance. If the method(s) from this class can be used without inheritance, I would prefer that. If they are public and non-static, then could an instance be created? If the methods are protected then we have to use inheritance.

Does that make sense?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Good point, seems missed setting it up  when current transaction finished.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Using of timer here was attempt to not load main verticle with log-time update operation. Do you mean that running of blocking code approach will be better here?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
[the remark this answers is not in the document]
oleksandr-yatsenko1: I'll change this to composition, thanks you for noticing this.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
oleksandr-yatsenko1: Why I asked about this plugin, because it looks like the picture attached in the mentioned ticket is the screenshot from Eclipse Code Formatter plugin itself - https://issues.folio.org/secure/attachment/15917/ecf.png. And there are both configuration files specified one for importing settings and another one for formatting.

oleksandr-yatsenko1: I have found this https://wiki.folio.org/display/FOLIJET/Java+code+formatter+for+IDEs, it looks like it contains formatting config used by FOLIO's developers.

marcjohnson-kint: I think that some of the EPAM developers use that plugin. This is not an agreed FOLIO method. 

I do not use it. I use the configuration settings within IntelliJ itself.

I suggest asking other EPAM folks who use IntelliJ for guidance on the plugin.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
marcjohnson-kint: > Using of timer here was attempt to not load main verticle with log-time update operation. Do you mean that running of blocking code approach will be better here?

I am not advocating for blocking code.

If the intent is to repeatedly run the code for a batch until there are no longer any items left to update, I suggest triggering the next batch when the last batch finishes. 

One approach is to trigger the batches recursively, e.g. https://github.com/folio-org/mod-circulation/blob/494e73fab76f637bc6c35603d3669e6214206cd4/src/main/java/org/folio/circulation/support/fetching/PageableFetcher.java#L37

oleksandr-yatsenko1: >I suggest triggering the next batch when the last batch finishes

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+    itemsMap = new LinkedHashMap<>();
+  }
+
+  public Future<String> updateItems(TenantAttributes attributes, Map<String, String> headers, Context vertxContext) {
+    Promise<String> promise = Promise.promise();
+
+    String fromModuleVersion = attributes.getModuleFrom().trim();
+    if (StringUtils.isBlank(fromModuleVersion) || versionChecker.isNewForThisInstall(fromModuleVersion)) {
+      // Update items only for versions before the expected
+      promise.fail(OPERATION_STATUS_VERSION_MISMATCH);
```
bohdan-suprun, line 76: Is it fail entire tenant init operation?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+
+  private static final String SQL_EXISTS_ITEMS_TO_UPDATE = new StringBuilder()
+      .append("SELECT EXISTS (SELECT null FROM item WHERE not (jsonb ? 'effectiveShelvingOrder'))")
+      .toString();
+
+  private static final String SQL_SELECT_ITEMS = new StringBuilder()
+      .append("SELECT id AS \"id\", ")
+      .append("jsonb AS \"itemJson\" ")
+      .append("FROM item WHERE not (jsonb ? 'effectiveShelvingOrder') ")
+      .append(" FOR UPDATE ")
```
bohdan-suprun, line 39: As far as I know `for update` means that the rows will be locked, but I don't see where it is used. Do we really need it?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+
+  private static final Logger log = LogManager.getLogger();
+
+  private static final String SQL_EXISTS_ITEMS_TO_UPDATE = new StringBuilder()
+      .append("SELECT EXISTS (SELECT null FROM item WHERE not (jsonb ? 'effectiveShelvingOrder'))")
+      .toString();
+
+  private static final String SQL_SELECT_ITEMS = new StringBuilder()
+      .append("SELECT id AS \"id\", ")
+      .append("jsonb AS \"itemJson\" ")
```
bohdan-suprun, line 37: Do we need entire item json, or just few properties from it?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+import org.folio.rest.persist.PgUtil;
+import org.folio.rest.persist.PostgresClient;
+
+public class ShelvingOrderUpdate {
+
+  private static final Logger log = LogManager.getLogger();
+
+  private static final String SQL_EXISTS_ITEMS_TO_UPDATE = new StringBuilder()
+      .append("SELECT EXISTS (SELECT null FROM item WHERE not (jsonb ? 'effectiveShelvingOrder'))")
+      .toString();
```
bohdan-suprun, line 33: Do we need this query at all? If there will be no matches than the `SQL_SELECT_ITEMS` will return empty result, I think.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+    });
+
+    return promise.future();
+  }
+
+  private Future<Integer> fetchAndUpdatesItems(Vertx vertx, PostgresClient postgresClient) {
+    Promise<Integer> promise = Promise.promise();
+
+    vertx.setTimer(500L, timer -> {
+      itemsShelvingOrderMap.clear();
```
bohdan-suprun, line 116: This might be unsafe. If the code executed concurrently then the data will be mixed up. Can we make the map local vs holding the state in class?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+              })
+              .eventually(v -> conn.close())
+                  .onSuccess(h -> log.info("Items update transaction succeeded"))
+                  .onFailure(h -> log.error("Items update transaction filed: {}", h.getMessage())));
+
+        } // Connection acquired successfully
+
+      }); // Eof getConnection scope
+
+     }); // Eof vert.x timer scope
```
bohdan-suprun, line 188: We can omit these comments if we move code to separate methods.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+      promise.fail(OPERATION_STATUS_VERSION_MISMATCH);
+    } else {
+      PostgresClient postgresClient =  PgUtil.postgresClient(vertxContext, headers);
+
+      hasItemsToUpdate(vertxContext.owner(), postgresClient)
+          .onSuccess(itemsExists -> {
+            log.warn("Items update started");
+          }).onFailure(h -> {
+            log.info("No items for update");
+      });
```
adamdickmeiss, line 85: Error ignored here. I guess that was intented.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
adamdickmeiss: You can run for half an hour if you want. The async tenant API can handle it. Avoid cleverness like this.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,197 @@
...
+    itemsShelvingOrderMap = new LinkedHashMap<>();
+    itemsMap = new LinkedHashMap<>();
+  }
+
+  public Future<String> updateItems(TenantAttributes attributes, Map<String, String> headers, Context vertxContext) {
+    Promise<String> promise = Promise.promise();
+
+    String fromModuleVersion = attributes.getModuleFrom().trim();
+    if (StringUtils.isBlank(fromModuleVersion) || versionChecker.isNewForThisInstall(fromModuleVersion)) {
+      // Update items only for versions before the expected
```
adamdickmeiss, line 75: It's very modern with these isBlank thingy.. Okapi will NEVER send a fromModuleVersion with a blank. It will be omitted and for YOU == null.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
adamdickmeiss: Very long function that I can't even comprehend.

src/test/java/org/folio/services/CallNumberUtilsTest.java
```diff
@@ -48,20 +48,67 @@ public void inputForShelvingNumber(
...
 
-    HoldingsRecord holdingsRecord = new HoldingsRecord();
-    EffectiveCallNumberComponentsUtil.setCallNumberComponents(item,holdingsRecord);
-    EffectiveCallNumberComponentsUtil.calculateAndSetEffectiveShelvingOrder(item);
-
-    assertThat(item.getEffectiveShelvingOrder(),is (desiredShelvingOrder));
+    return item;
   }
+  
 }
```
adamdickmeiss, line 114: More tests needed. Add a test where you call tenant init or loadData and check that if your code fails, then in turn loadData fails or succeeds as you'd expect.

adamdickmeiss (changes requested):
More tests and use compose in loadData so that your code can actually signal success/failure.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Do you mean that it's need to avoid of using "setTimer" and just execute even huge amount  items update via vertx postgres client directly, even without leveraging of "execute blocking"?

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
adamdickmeiss: I'm using checkstyle Google in code that I maintain. Then these discussions become mood. Not because that style is better than any other as such, but because this will be fixed at people desktop or in Jenkins.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
marcjohnson-kint: @oleksandr-yatsenko1 

> I'll take a look the code by link, looks like that's background fetching of data for some kind of UI consumer like a navigable table.

The purpose of that code is different. It only fetches records. The principle is the same I believe, at the end of each page / batch of records, the next is fetched. The primary difference I can think of is that this process needs to also update records.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
marcjohnson-kint: @adamdickmeiss 

> I'm using checkstyle Google in code that I maintain. Then these discussions become mood.

I agree that adopting a code style would be very helpful with these kinds of topics. I do regret not adopting them much earlier in the lifetime of these modules. 

At the moment, it is a matter of finding the time to implement the checks and make the existing code compliant. The Core Functional team has no capacity for such work.

As far as I am aware, FOLIO has no agreed standard for this.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Yes, that's attempts to block selected rows (i.e. so names row blocking of table) in separate transaction where then update operation for these rows will be performed. Then at completion, the transaction going to be committed and closed. I thought that blocking of rows is a kind of tradeoff for the cases where items update are going on but some of clients tries to work with data and possibly updates them. In case of blocking rows by a relatively small sizes, these clients operations of course will be in wait state until rows blocking will be released, but from other hand that will allow to keep data in consistency and possibly probability of such case isn't so high, I mean attempt to access exactly the same data belong to blocking amount of rows.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Hum, that's a good question from point of view reducing of acquired amount of data as maximum as possible, but as it was implemented in calculation of shelving order there is whole Item object required as parameter. This is quite reasonable because where is not only string values involved in calculation like volume, enumeration etc. but also composite data object like "effective call number components", so anyway we need to restore the whole Item from database and pass it for calculation or just retrieve separate values as string with part of Items JSON data, i.e. "effective call number component", for further deserialization and passing into already implemented shelving calculation routine. By the way, this routine should be also extended with overloaded methods with another signature too. So the main question here is - will be current retrieval of items JSON with further deserialization be less effective that retrieval of separate part of items which in turns needs to include part of items as composite JSON object too.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: That a kind of tradeoff I think, because in this way we can check together with version checking if it's need to start update at all. The query itself supposed to be not so heavyweight, but possibly it can be removed indeed. Thank you for noticing this.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Here I think this extra data structure isn't required at all, it will be removed. All the data required for update will be collected internally in handler as simple list of object with further inline transformation and passing to update routine.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: The purpose if it to select items data with small portion, then process items with shelving order calculation and then update items in the same transaction with changed jsonb values. I'll try do split it.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -126,6 +120,19 @@ String servicePointUserFilter(String s) {
               .add("users", "service-points-users");
           }
           return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
-        });
+        }).onSuccess(handler -> {
+
+          if (ShelvingOrderUpdate.isModuleUpgradeRequest(attributes)) {
+            // Updates items with shelving order property. Only for version before the 19.5.0.
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Thank you for pointing this, it will be better to avoid mention about version number at all. Will fix it.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -126,6 +120,19 @@ String servicePointUserFilter(String s) {
               .add("users", "service-points-users");
           }
           return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
-        });
+        }).onSuccess(handler -> {
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Good point, not sure what's benefit of this because there is no further processing supposed (this part of code possibly will be removed via extra ticket in jira). I'll fix this, thank you Adam.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -126,6 +120,19 @@ String servicePointUserFilter(String s) {
...
+            new ShelvingOrderUpdate(5000)
+                .updateItems(attributes, headers, vertxContext)
+                .onSuccess(h -> log.info("Shelving order updates completed in {} seconds",
+                    (System.currentTimeMillis() - startTime) / 1000))
+                .onFailure(
+                    h -> log.error("Error occurs during shelving order update: {}", h.getMessage()));
+          }
+      });
   }
+
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: Do you mean that using compose with future will provide possibility to return failure status of load data process by set future's result in case if items update somehow fails?

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
```diff
@@ -35,7 +41,7 @@ public static void calculateAndSetEffectiveShelvingOrder(Item item) {
           .map(StringUtils::trim)
           .collect(Collectors.joining(" "))
       );
-//      String suffixValue = Objects.toString(item.getEffectiveCallNumberComponents().getSuffix(), "").trim();
+
```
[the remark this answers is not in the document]
oleksandr-yatsenko1: This was already removed have no ideas how it have appears here, will check it, thank you.

src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java
oleksandr-yatsenko1: I have configured Eclipse Code Formatter plugin (because as I remember it was recommended: https://wiki.folio.org/display/FOLIJET/Java+code+formatter+for+IDEs) with google style configuration got from here https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml. And for imports arranging I use folio.importorder from the ticket https://issues.folio.org/browse/FOLIO-1667.

image

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: If I understand correctly the better approach will be to use some kind of working pool to run several update processes simultaneously. Because Postgres Client possibly do the same under the hood, i.e. use typical executeBlocking like vertx JDBC does. That why possibly we don't need to care about blocking of main event-loop by database update operation. The only difference I think could be in multi-processing, it will be faster of course but require extra tracking of updating rows (possibly by remember borders of rowids or items ids), to avoid lock waits until parallel update completed (if any of rows range will cross, without such range control). Currently  there is only single update process supposed, possibly it will be enough for this, especially takin into account that this should be performed only once. But in general,  I think we lack of such functionality of data updates because unify approach, possibly with multi-processing capabilities should be introduced to keep data in consistency.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: It was supposed to fail only part of operation related to module module. Will check it, thanks.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Here just only messages logged, but it looks in case of failure we need to fail upgrade part of tenant loading. Not sure if it possibly to separate it from rest of loadData. Therefore there is only logging remain. How do you think, does it make sense?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Replacing of isBlank from the "Apache's common lang" with just checking for null (possibly by no less popular Objects.isNull) will avoid of using extra dependencies here. Because isBlank works well with all the undesired strings too. But if we are pretty sure about some of handling from Okapi side, it make sense to perform only bare minimum checking to not be confused with varieties of expected values of course.

oleksandr-yatsenko1:
> More tests and use compose in loadData so that your code can actually signal success/failure.

In case of failure of items update, sho

src/main/java/org/folio/rest/impl/TenantRefAPI.java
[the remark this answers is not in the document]
adamdickmeiss: yes.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
adamdickmeiss: Well.. That's the key "undesired" string. Are they desired? No. That's why systems should not allow it - EXCEPT when it's coming from a user and even there is should be thrown away (eg passwords). But not a serious issue.

oleksandr-yatsenko1: I see, thank you for clarifieng.

src/test/java/org/folio/services/CallNumberUtilsTest.java
oleksandr-yatsenko1: It looks it's need to try to test directly methods from added shelving order update class, because relying on the flow leveraging tenant API looks quite unstable and time consuming from point of view test execution.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+   *
+   * @param vertx
+   * @param postgresClient
+   * @return
+   */
+  private Future<Integer> fetchAndUpdateBunch(Vertx vertx, PostgresClient postgresClient) {
+    log.info("The routine of \"fetchAndUpdatesItems\" has started");
+    Promise<Integer> fetchAndUpdatePromise = Promise.promise();
+
+    postgresClient.getConnection(ar -> {
```
julianladisch, line 146: Consider using `postgresClient.withTransaction(tx -> {`

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+   * Performs fetching items by fixed bunches of rows with further property calculations and
+   * updating in database
+   *
+   * @param vertx
+   * @param postgresClient
+   * @return
+   */
+  private Future<Integer> fetchAndUpdateBunch(Vertx vertx, PostgresClient postgresClient) {
+    log.info("The routine of \"fetchAndUpdatesItems\" has started");
+    Promise<Integer> fetchAndUpdatePromise = Promise.promise();
```
julianladisch, line 144: Consider avoiding this Promise.
`return postgresClient.getConnection().compose(connection -> {`
or
`return postgresClient.withTransaction(tx -> {`
You may need to add `.map(x -> someReturnValue)` at the end.

julianladisch: Or `.mapEmpty()`.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+    return Future.succeededFuture();
+  }
+
+  /**
+   * Read items and fill them into data structures for further processing
+   *
+   * @param rowSet
+   * @return
+   */
+  public Future<List<Item>> aggregateRow2List(RowSet<Row> rowSet) {
```
julianladisch, line 222: No async operation is involved, this method should return `List<Item>`, no `Future` wrapping is needed.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+    log.info("Finishing of aggregateRow2List, promise: {}", promise);
+    return promise.future();
+  }
+
+  /**
+   * Calculate items shelving order
+   *
+   * @return
+   */
+  public Future<Map<UUID, String>> calculateShelvingOrder(List<Item> itemList) {
```
julianladisch, line 250: No async operation is involved, this method should return `Map<UUID, String>`, no `Future` wrapping is needed.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+    return Future.succeededFuture(updatedItemsShelvingOrderMap);
+  }
+
+  /**
+   * Prepare items update parameters
+   *
+   * @param itemIdShelvingOrderMap
+   * @return
+   */
+  public Future<List<Tuple>> prepareItemsUpdate(Map<UUID, String> itemIdShelvingOrderMap) {
```
julianladisch, line 274: No async operation is involved, this method should return `List<Tuple>`, no `Future` wrapping is needed.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+            "none"), SQL_UPDATE_ITEMS);
+
+    if (itemsParams.isEmpty()) {
+      log.info("Items parameters for update SQL are empty");
+      return Future.succeededFuture(0);
+    }
+
+    Promise<Integer> promise = Promise.promise();
+    connection.preparedQuery(SQL_UPDATE_ITEMS).executeBatch(itemsParams, res -> {
+      if (res.succeeded()) {
```
julianladisch, line 306: The Promise is not needed.

```
return connection.preparedQuery(SQL_UPDATE_ITEMS).executeBatch(itemsParams)
    .map(result -> {
      ...
      return updatedItemsCount;
    });
```

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
```diff
@@ -0,0 +1,366 @@
...
+
+  public static final String UPDATED_ATTR_NAME = "effectiveShelvingOrder";
+
+  private static final String SQL_SELECT_ITEMS =
+      new StringBuilder()
+          .append("SELECT id AS \"id\", ")
+          .append("jsonb AS \"itemJson\" ")
+          .append(String.format("FROM item WHERE not (jsonb ? '%s') ", UPDATED_ATTR_NAME))
+          .append(" FOR UPDATE ")
+          .append("LIMIT $1").toString();
```
julianladisch, line 44:
```suggestion
  private static final String SQL_SELECT_ITEMS =
          "SELECT id AS \"id\", " +
          "jsonb AS \"itemJson\" " +
          String.format("FROM item WHERE not (jsonb ? '%s') ", UPDATED_ATTR_NAME) +
          " FOR UPDATE " +
          "LIMIT $1";
```
This is more readable and has the same performance. The compiler automatically uses a StringBuilder internally: https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java
The same for the SQL_UPDATE_ITEMS1 and SQL_UPDATE_ITEMS.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: The point here is to place all the steps in single transaction related execution instead of spawning a new transaction with further needing to control it. Correct me if I'm wrong, please.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Is the difference in usage of withTtransaction in a way of control transaction itself?

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: I see, it will be better to avoid java code in such declarations. As for performance I never had any doubts that usage of plus operation uses builder under the hood. I just used build for sake of readability. I'll fix this, thank you for noticing.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
oleksandr-yatsenko1: Done, thank you Julian.

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
julianladisch: Yes, it is similar to PgPool.withTransaction and automatically runs COMMIT or ROLLBACK: https://vertx.io/docs/vertx-pg-client/java/#_simplified_transaction_api

src/main/java/org/folio/rest/support/ShelvingOrderUpdate.java
julianladisch: Thanks, the new code is good!

src/main/java/org/folio/rest/support/CompletableFutureUtil.java
```diff
@@ -1,11 +1,17 @@
...
 import io.vertx.core.Handler;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 
 public final class CompletableFutureUtil {
+
+  private static final Logger log = LogManager.getLogger();
+
+  public static final long TIMEOUT = 180;
+
```
julianladisch, line 14: This is not used and can be removed.

bohdan-suprun:
Hi all, closing this PR, a new https://github.com/folio-org/mod-inventory-storage/pull/609 is created, thanks all for thoughts.

PR #592, opened 2021-03-11
Fix failing tests on master.

bohdan-suprun:


marcjohnson-kint (commented):
Merged with my permission to expedite fix

PR #594, opened 2021-03-11
src/license/THIRD-PARTY.properties not in Git MODINVSTOR-695

adamdickmeiss:


julianladisch:
Five licenses were manually added to THIRD-PARTY.properties  in lines 43 ff.
Why should they be removed?

marcjohnson-kint:
> Five licenses were manually added to THIRD-PARTY.properties in lines 43 ff.
> Why should they be removed?

If we manually maintain this list of licenses then I don't think we can consider it generated.

Why were those licenses added?

julianladisch:
license-maven-plugin can extract license information from the pom files of most dependencies.
Some dependencies have a pom file that doesn't have a license entry. For those dependencies we need to manually add the correct license to src/license/THIRD-PARTY.properties.

adamdickmeiss:
> license-maven-plugin can extract license information from the pom files of most dependencies.
> Some dependencies have a pom file that doesn't have a license entry. For those dependencies we need to manually add the correct license to src/license/THIRD-PARTY.properties.

If this is a file that use edits, fine. But it isn't. This is modified by a Maven run!

julianladisch:
license-maven-plugin is designed to change the file if the dependency changes.
If you don't like it you can change pom.xml so that maven copies src/license/THIRD-PARTY.properties to a place below target and reconfigure <missingFile>.

adamdickmeiss:
> license-maven-plugin is designed to change the file if the dependency changes.
> If you don't like it you can change pom.xml so that maven copies src/license/THIRD-PARTY.properties to a place below target and reconfigure .

OK. So this is like a "regression" file.. Something you update when you feel like.

PR #596, opened 2021-03-16
MODINVSTOR-454:Set default value to false for previouslyHeld property

K-Felk:
I added the default value to the schema file, and looked for interfaces that
referenced it, incrementing their minor version numbers (I only found one).
Then added the SQL script that will update the database, changing all values 
of null to false, and updated the database creation schema.

This change didn't seem to require it's own test, I just added a line to 
one of the existing tests that creates an instance record, checking to see if the
default value is present.

refs: https://issues.folio.org/browse/MODINVSTOR-454


src/main/resources/templates/db_scripts/setPreviouslyHeldDefault.sql
```diff
@@ -0,0 +1,11 @@
...
+
+ALTER TABLE ${myuniversity}_${mymodule}.instance DISABLE TRIGGER USER;
+
+UPDATE ${myuniversity}_${mymodule}.instance
+  SET jsonb = JSONB_SET(instance.jsonb, '{previouslyHeld}', TO_JSONB(false))
+WHERE jsonb->>'previouslyHeld' IS NULL;
+
+ALTER TABLE ${myuniversity}_${mymodule}.instance ENABLE TRIGGER USER;
+
+END TRANSACTION;
```
marcjohnson-kint, line 11: I think it is a good idea to have a test for the upgrade scripts as well. For example, `ItemEffectiveCallNumberDataUpgradeTest` does something similar for the setting of effective call number components for existing records.

NEWS.md
```diff
@@ -1,5 +1,8 @@
 ## 20.1.0 In-progress
 
+* Set a default value of false for instance.previouslyHeld (https://issues.folio.org/browse/MODINVSTOR-454)
```
marcjohnson-kint, line 3: If we include links to JIRA issues, I think it is better to provide a label for them that shortens the length of the line in the formatted markdown.

```suggestion
* Defaults instance.previouslyHeld to false ([MODINVSTOR-454](https://issues.folio.org/browse/MODINVSTOR-454))
```

src/main/resources/templates/db_scripts/setPreviouslyHeldDefault.sql
K-Felk: I tried to write this, but I can't figure out the syntax for the database command to set the previouslyHeld field back to null in perparation for running the upgrade script.  I either get incomprehensible error messages about polymorphism or syntax errors.  I've posted what I have, the test is in PreviouslyHeldDataUpgradeTest.java

marcjohnson-kint: Gotcha, As the API has effectively already done the defaulting of any new records, they would need to be reset.

I hadn't thought of that :-/

src/test/java/org/folio/rest/api/PreviouslyHeldDataUpgradeTest.java
```diff
@@ -0,0 +1,66 @@
...
+        .asString("templates/db_scripts/setPreviouslyHeldDefault.sql")
+        .replace("${myuniversity}_${mymodule}", "test_tenant_mod_inventory_storage");
+
+    @Test
+    public void canMigrateToDefaultPreviouslyHeldValue() {
+        instancesClient.create(instance(instanceId));
+
+        String query = "UPDATE test_tenant_mod_inventory_storage.instance " +
+            "SET jsonb = JSONB_SET(instance.jsonb, '{previouslyHeld}', TO_JSONB(null)) " +
+            "WHERE jsonb->>'previouslyHeld' IS NOT NULL;";
```
marcjohnson-kint, line 36: Could this approach work instead - https://til.hashrocket.com/posts/78333d571f-remove-keyvalue-from-jsonb-data-structure ?

K-Felk: I spent most of the morning trying this-I either get syntax errors or it seems to execute, but do nothing.  Closest I've gotten to success is `UPDATE test_tenant_mod_inventory_storage.instance SET jsonb = jsonb - 'previouslyHeld';`
This executes without an sql error, but when I check the previouslyHeld key afterwards, it's still present and still set to false.

marcjohnson-kint: Thanks for the update. I don't think it is worth trying to do test this more thoroughly given these challenges.

Have / Could you test this manually as a one off?

K-Felk: What if I merge this without the test and add it later, once I've figured out how it's supposed to work?  I can ask Bohdan or some of the other back-end devs for help.  This is going to come up again for https://issues.folio.org/browse/MODINVSTOR-669, as that issue will also need a migration script, so I'd like to know what I'm doing wrong.  

I have not tested it manually, and I might need some guidance on how to do so-is there a script I can invoke from the command-line that will set up the database?  If I can do that, I could presumably run the sql script from the database command line.  I'm assuming I'll need to install postgresql locally?

marcjohnson-kint: > What if I merge this without the test and add it later, once I've figured out how it's supposed to work?

Sure, I'm ok with that. @oleksandr-yatsenko1 has been doing some work on testing upgrades for his pull request, there might be some overlap.

> I might need some guidance on how to do so-is there a script I can invoke from the command-line that will set up the database? 

There are some scripts in the repository. I might be the only person who uses them.

[This](https://wiki.folio.org/display/FOLIJET/How+to+test+data+migration+scripts+locally) is the best documentation we have at the moment.

> If I can do that, I could presumably run the sql script from the database command line. I'm assuming I'll need to install postgresql locally?

Yes, or use a docker image.

I've been working on a Docker compose file for setting up FOLIO infrastructure and running a module.

K-Felk: I asked @bohdan-suprun about this, he says what's happening is when I call the api, the JSON data is deserialized into a java object, and because I've set a default value for this attribute in the RAML, it's automatically populated at that point if it's null.  So I am actually erasing the value, I'm just automatically re-setting it when I call the API to retrieve it.  He recommends pulling the data directly from the database to make sure it's been set properly rather than calling the API, so that's what I'm going to try and do.

marcjohnson-kint: > what's happening is when I call the api, the JSON data is deserialized into a java object, and because I've set a default value for this attribute in the RAML, it's automatically populated at that point if it's null

That seems totally reasonable. It's one of the reasons I'm not keen on this code generation approach.

I think that means we cannot easily test the upgrade process.

marcjohnson-kint: > He recommends pulling the data directly from the database to make sure it's been set properly rather than calling the API, so that's what I'm going to try and do.

That is worth trying

K-Felk: Okay, I think I finally have this working.

NEWS.md
```diff
@@ -4,6 +4,9 @@
 * Introduces `REPLICATION_FACTOR` environment variable for configuring Kafka topic replication (MODINVSTOR-694)
 * Provides `instance-storage 7.7`
 
+* Defaults instance.previouslyHeld to false (MODINVSTOR-454)
+* Provides `instance-storage 7.7`
```
marcjohnson-kint, line 8: Please move this to a separate section of the news for the current release rather than together with the last release

marcjohnson-kint (approved):

PR #597, opened 2021-03-16
Add contributorsNames index

hjiebsco:
Please see https://issues.folio.org/browse/RMB-797

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -471,6 +471,10 @@
           "multiFieldNames": "title,indexTitle,alternativeTitles[*].alternativeTitle,series"
         },
         {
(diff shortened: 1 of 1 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -470,6 +470,10 @@
           "fieldName": "allTitles",
           "multiFieldNames": "title,indexTitle,alternativeTitles[*].alternativeTitle,series"
         },
+        {
+          "fieldName": "contributorsName",
```
julianladisch, line 474:
```suggestion
          "fieldName": "contributorsNames",
```
It should be plural because it concatenates the names of multiple contributors, similar to allTitles.

julianladisch (approved):
Looks good, thanks!

PR #599, opened 2021-03-19
KafkaContainer; allow combo of tests to work MODINVSTOR-706

adamdickmeiss:


src/main/java/org/folio/services/kafka/KafkaProperties.java
```diff
@@ -9,18 +9,20 @@
...
   private static int port = parseInt(getenv().getOrDefault("KAFKA_PORT", "9092"));
+  private static String host = getenv().getOrDefault("KAFKA_HOST", "localhost");
 
-  public static int changePort(int port) {
+  public static void setHostPort(String host, int port) {
+    KafkaProducerServiceFactory.clear();
     KafkaProperties.port = port;
-    return port;
+    KafkaProperties.host = host;
   }
```
marcjohnson-kint, line 18: Personally, I'd prefer these were set separately. However I understand that changing either of these is a signal to reset the service factory, and it's easier to do that in one place only.

marcjohnson-kint (approved):
Thank you for resolving the issues that stopped a subset of tests being run successfully.

Given that we seem to have defacto settled on the use of Test Containers in FOLIO (AFAIK no explicit decision has been made), I'm ok with moving to that instead of other tooling.

I've asked @bohdan-suprun to review this as well, as I believe he did the original work and may know why we chose different tooling for Kafka.

src/main/java/org/folio/services/kafka/KafkaProperties.java
adamdickmeiss: ok

bohdan-suprun (approved):
Looks good, Thanks!

PR #602, opened 2021-03-26
MODINVSTOR-710: Suspend reindex stream when producer write queue is full.

bohdan-suprun:
## Changes:

* Add `folio-kafka-wrapper` dependency.
* Remove unnecessary code that is now used from the `folio-kafka-wrapper`.
* Use separate shared producer per topic and separate producer when stream published.
* Pause reindex stream when producer write queue is full.
* Simplify publishing approach.


pfilippovEPAM (approved):

src/main/java/org/folio/services/kafka/KafkaProperties.java
```diff
@@ -1,44 +1,29 @@
...
-import static java.lang.Integer.parseInt;
 import static java.lang.Short.parseShort;
 import static java.lang.System.getenv;
 
-import java.util.Map;
-import org.apache.kafka.common.serialization.StringSerializer;
-
 public final class KafkaProperties {
-  private static int port = parseInt(getenv().getOrDefault("KAFKA_PORT", "9092"));
+  private static String port = getenv().getOrDefault("KAFKA_PORT", "9092");
```
hjiebsco, line 7: Any advantage to use `string` over `int` for `port`, or just try to match with `KafkaConfig` in `folio-kafka-wrapper`?

src/main/java/org/folio/services/reindex/ReindexJobRunner.java
```diff
@@ -1,53 +1,59 @@
...
+      new CommonDomainEventPublisher<>(vertxContext, okapiHeaders, INVENTORY_INSTANCE),
+      tenantId(okapiHeaders));
   }
 
-  public ReindexJobRunner(InstanceDomainEventPublisher publisher,
-    PostgresClientFuturized postgresClient, ReindexJobRepository repository,
-    Context vertxContext) {
+  public ReindexJobRunner(PostgresClientFuturized postgresClient, ReindexJobRepository repository,
+    Context vertxContext, CommonDomainEventPublisher<Instance> domainEventPublisher,
+    String tenantId) {
```
hjiebsco, line 51: Is `public` just for test convenience?

hjiebsco (approved):

src/main/java/org/folio/services/kafka/KafkaProperties.java
bohdan-suprun: I tried to match the `folio-kafka-wrapper`

src/main/java/org/folio/services/reindex/ReindexJobRunner.java
bohdan-suprun: Yes, the constructor is for tests.

PR #603, opened 2021-03-26
MODINVSTOR-669: Add effective holdings location attribute to holdings record

K-Felk:
The actual code that adds the field is pretty straightforward, I think.  
For the code that determines the actual value, I built a function that 
determines what the effective location should be for any given holdings
record, then wired it into all update/create code in the storage service.
I notice that for similar calculations for other data structures (like item 
records), functions like this are built into a service, and I considered building 
this that way as well. But in those records, there are many complex determinations 
that need to be made-holdings records only seem to have this value to calculate, 
and it's a relatively simple one, so forming it as an external service seemed like 
a lot of unnecessary code overhead. If it's desireable for reasons I'm not aware of,
I'll rewrite it.

For tests, I tried to integrate checks for the attribute on new records into existing tests
that created new records.  Tests for behavior that's supposed to happen when existing 
values change required their own tests, which I created.  I also created a migration 
script for existing values, wired that into the database creation schema, and created 
a migration test for it.  Work on other issues has taught me that it's inadvisable 
to use the API to retrieve changed values, as the API will often set default values for 
fields when you don't want them to be set (for example, when you're trying to verify 
you've successfully removed a value). So the migration script test only uses the API 
for creation of new data-all subsequent checks retrieve data directly from the database 
using postgresql queries.

Refs: https://issues.folio.org/browse/MODINVSTOR-669


NEWS.md
```diff
@@ -4,6 +4,9 @@
 * Introduces `REPLICATION_FACTOR` environment variable for configuring Kafka topic replication (MODINVSTOR-694)
 * Provides `instance-storage 7.7`
 
+* Provides `holdings-storage 4.6` (MODINVSTOR-669)
+* Add effective holdings location attribute to holdings record (MODINVSTOR-669)
```
marcjohnson-kint, line 8: I tend to follow the convention that the provides statements are less important than the functional changes.

Please re-order these statements and include them in their own section separate to the previous release.

```suggestion
* Adds `effective location` attribute to holdings record (MODINVSTOR-669)
* Provides `holdings-storage 4.6` (MODINVSTOR-669)
```

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -652,6 +652,11 @@
           "targetTable": "location",
           "tOps": "ADD"
         },
+        {
+          "fieldName": "effectiveLocationId",
+          "targetTable": "location",
+          "tOps": "ADD"
+        },
```
marcjohnson-kint, line 659: Why has a database index been introduced for this property? Is this a pre-emptive step because this is going to be searchable in future?

marcjohnson-kint (approved):

bohdan-suprun:
Hi @marcjohnson-kint, @K-Felk, do we really need an extra property for this case? The property uses only two properties that both on holding record, I assume if we need to search by 'effectiveLocationId' we can use following CQL - `holdings.permanentLocationId==<the-location-id> OR holdings.temporaryLocationId==<the-location-id>`. 

But I probably missing something, feel free to ignore my comment.

marcjohnson-kint:
@bohdan-suprun 

> do we really need an extra property for this case?

It's a reasonable question.

> I assume if we need to search by 'effectiveLocationId' we can use following CQL - `holdings.permanentLocationId==<the-location-id> OR holdings.temporaryLocationId==<the-location-id>`.

I don't think that would match the intended behaviour.

Let me use an example to explain my reasoning.

Given an holdings record with permanent location `history` and a temporary location `collection point A`
When a client searches for holdings with effective location `history`
Then this holdings record _should not_ be present in the results (as the effective location is `collection point A`)

My understanding is that the OR based query would include that record and so would not match the intended behaviour.

bohdan-suprun:
@marcjohnson-kint, ah yes, you're right.

src/main/resources/templates/db_scripts/schema.json
K-Felk: Since both permanent and temporary location IDs are searchable and have their own indexes, I thought this one might need one as well.  If that's not the case, I'll remove.

marcjohnson-kint: I think it is reasonable to include it. I only wanted to understand the reasoning for doing so.

marcjohnson-kint (commented):
Is there a batch API for holdings records?

src/main/resources/templates/db_scripts/schema.json
bohdan-suprun: I'd better add a foreign key instead. it will also create a b-tree index.

K-Felk: I think that's what I've done here (isn't it?). This entry is in the foreign keys array.

bohdan-suprun: Yes, you did. Sorry again :(

K-Felk:
> Is there a batch API for holdings records?

There is, yes.  I believe it's described by holdingsrecords.json in the ramls directory.  I did include code to ensure that batch-created holdings records have this property, and I test for it's existence in one of the tests that creates holdings records in batch.

marcjohnson-kint:
> I test for it's existence in one of the tests that creates holdings records in batch.

Cool, I didn't spot that

NEWS.md
```diff
@@ -1,12 +1,14 @@
 ## 20.2.0 IN-PROGRESS
 
+* Adds `effective location` attribute to holdings record (MODINVSTOR-669)
 * Defaults instance.previouslyHeld to false (MODINVSTOR-454)
-* Provides `instance-storage 7.7`
+* Provides `holdings-storage 4.6` (MODINVSTOR-669)
```
marcjohnson-kint, line 5: Is this the correct merge? I thought `instance-storage` was changed after the last release?

marcjohnson-kint, line 5: Is this a correct merge? I thought that `instance-storage` was changed since the last release

K-Felk: Evidently not.  Something about merging in the latest code is also crashing the tests.  I'll fix it.

K-Felk:
Okay, I rolled back the merge, re-merged master into the branch locally, ran tests and they passed.

K-Felk:
huh.  It looks like the tenant removal cleanup is timing out at the end of storage tests.  This isn't happening when I run the tests locally-they pass fine.

K-Felk:
Booyah!  I *thought* the error looked familiar-it's similar to the error I was getting when running tests in mod-circulation-storage.  We fixed that by moving to test containers-so I tried the same thing here, and it worked!

PR #606, opened 2021-04-16
MODINVSTOR-658: Make InstanceDomainEventPublisher to extend AbstractDomainEventPublisher

bohdan-suprun:


src/main/java/org/folio/services/batch/BatchOperationContextFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+import io.vertx.core.Future;
+import java.util.List;
+import java.util.function.Function;
+import org.folio.persist.AbstractRepository;
+
+public final class BatchOperationContextFactory {
+  private BatchOperationContextFactory() {}
+
+  public static <T> Future<BatchOperationContext<T>> buildBatchOperationContext(
```
hjiebsco, line 15: Consider to update `InstanceService` line 76 to use this to reduce code repeat.

hjiebsco (approved):
LGTM

src/main/java/org/folio/services/batch/BatchOperationContextFactory.java
bohdan-suprun: Thank you! Changed now.

bohdan-suprun:
Hi @marcjohnson-kint, will you or the prokopovych team be able to review the changes, because it affects `InstanceStorageAPI` class, and POST/PUT/DELETE for instances?

marcjohnson-kint:
> Hi @marcjohnson-kint, will you or the prokopovych team be able to review the changes, because it affects `InstanceStorageAPI` class, and POST/PUT/DELETE for instances?

@bohdan-suprun thanks for reaching out to me.

A review from the Prokopovych team is not required. If it is wanted, I can try to review it soon. Is this work for 2021 R1 or R2?

bohdan-suprun:
Thanks @marcjohnson-kint. It is for R2.

src/main/java/org/folio/persist/InstanceMarcRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Map;
+import org.folio.rest.jaxrs.model.MarcJson;
+
+public class InstanceMarcRepository extends AbstractRepository<MarcJson> {
+  public static final String INSTANCE_SOURCE_MARC_TABLE = "instance_source_marc";
+
+  public InstanceMarcRepository(Context context, Map<String, String> okapiHeaders) {
+    super(postgresClient(context, okapiHeaders), INSTANCE_SOURCE_MARC_TABLE, MarcJson.class);
+  }
+}
```
marcjohnson-kint, line 15: What is this used for? Was the intention that this would provide the MARC record associated with the instance?

bohdan-suprun: Right now it is used only to remove all records in the table when an instance/all instances removed. It relies on RMB methods instead of using a self written delete SQL query.

pfilippovEPAM (approved):

MikhailFokanov (approved):
Looks good.

PR #607, opened 2021-04-16
MODINVSTOR-658: Simplify publishers logic.

bohdan-suprun:


hjiebsco (approved):

src/main/java/org/folio/rest/support/CollectionUtil.java
```diff
@@ -26,6 +26,15 @@ private CollectionUtil() {}
       .collect(Collectors.toList());
   }
 
+  public static <T> T getFirst(Collection<T> collection) {
+    if (collection == null) {
+      return null;
+    }
+
+    return collection.stream()
```
pfilippovEPAM, line 34: what do you think to replace streams with iterator? it would be like:

`values!=null ? ofNullable(values.iterator().next()) : empty()`

pfilippovEPAM (approved):

src/main/java/org/folio/rest/support/CollectionUtil.java
bohdan-suprun: Make sense, changed.

PR #608, opened 2021-04-19
MODINVSTOR-716: Query search on classifications times out.

K-Felk:
I added code for the new index, modeling it on similar indexes that already 
exist in the instances table.  As requested, I removed the gindex for this field.

### DIFF
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -393,12 +393,6 @@
           "removeAccents": true
         },
         {
-          "fieldName": "classifications",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
           "fieldName": "subjects",
           "tOps": "DELETE",
           "caseSensitive": false,
@@ -462,6 +456,11 @@
         }
       ],
       "fullTextIndex": [
+        {
+          "fieldName": "classifications",
+          "arraySubfield" : "classificationNumber",
+          "arrayModifiers": ["classificationTypeId"]
+        },
         {
           "fieldName": "keyword",
           "multiFieldNames": "title,contributors[*].name,identifiers[*].value"
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -476,6 +476,50 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
   }
 
   @Test
+  public void canSearchByClassificationNumber()
+    throws MalformedURLException,
+    InterruptedException,
+    ExecutionException,
+    TimeoutException {
+    
+    UUID firstInstanceId = UUID.randomUUID();
+    UUID secondInstanceId = UUID.randomUUID();
+    UUID classificationTypeId = UUID.randomUUID();
+
+    JsonObject firstClassifications = new JsonObject();
+    firstClassifications.put("classificationTypeId", classificationTypeId.toString());
+    firstClassifications.put("classificationNumber", "K1 .M385");
+
+    JsonObject secondClassifications = new JsonObject();
+    secondClassifications.put("classificationTypeId", classificationTypeId.toString());
+    secondClassifications.put("classificationNumber", "KB1 .A437");
+
+    JsonObject firstInstanceToCreate = smallAngryPlanet(firstInstanceId)
+    .put("classifications", new JsonArray()
+    .add(firstClassifications));
+    JsonObject secondInstanceToCreate = nod(secondInstanceId)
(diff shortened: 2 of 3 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -462,6 +456,11 @@
         }
       ],
       "fullTextIndex": [
+        {
+          "fieldName": "classifications",
+          "arraySubfield" : "classificationNumber",
+          "arrayModifiers": ["classificationTypeId"]
```
marcjohnson-kint, line 462: Why have these been added? Are they necessary for using `classifications` as a full text index with the same CQL as before?

marcjohnson-kint (commented):
Personally, I would prefer if this change was associated with some tests that demonstrate that the `classifications` CQL index is supported. I understand it won't demonstrate the use of the index.

src/main/resources/templates/db_scripts/schema.json
K-Felk: I modeled this on how other full-text indexes in the same table that index arraylist fields are structured.  Contributors, for example, is structured much like classifications, as a list which contains multiple subfields.  In that case, the name is what people actually want to search.  In this case, it's the classification number.  I'm assuming you specify a subfield because indexing all subfields could lead to false matches-for example, a search on classification number which also happens to match part of the classification ID.  Arraymodifiers I'm less sure about-my assumption there is that it possibly allows you to specify searching for a classification number under a specific classification type.  Since other similar indexes have that capability, I included it here as well.  I don't think it's required for search per se.

K-Felk:
Okay, I will work on that today.

marcjohnson-kint:
@julianladisch was this the change you envisaged?

src/main/resources/templates/db_scripts/schema.json
marcjohnson-kint: Ok, this seems speculative to me, given that folks aren't using CQL with those modifiers and haven't asked to use them.

With the aim of expediency, unless @julianladisch suggests we shouldn't add them, I'm going to suggest we leave them in

julianladisch: The schema.json change is perfect.

marcjohnson-kint: > The schema.json change is perfect.

Thanks @julianladisch

marcjohnson-kint (approved):

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -475,6 +475,50 @@ public void canGetAllInstances()
...
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+    
+    client.get(instancesStorageUrl("?query=classifications%3D%22K1%20.M385%22"), StorageTestSuite.TENANT_ID,
+    ResponseHandler.json(getCompleted));
+
+    Response response = getCompleted.get(5, TimeUnit.SECONDS);
+
+    JsonObject responseBody = response.getJson();
+
+    JsonArray allInstances = responseBody.getJsonArray("instances");
```
julianladisch, line 516:
```suggestion
    JsonArray allInstances = searchForInstances("classifications =/@classificationNumber \"K1 .M385\"");
```

This checks that the arraySubfield has been declared in schema.json. If the declaration is missing or wrong the /@ will fail.

marcjohnson-kint: Thanks @julianladisch 

I agree with using `searchForInstances` to avoid the duplicated code.

@K-Felk please define separate tests for the separate CQL variants.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -392,12 +392,6 @@
           "caseSensitive": false,
           "removeAccents": true
         },
-        {
-          "fieldName": "classifications",
-          "tOps": "ADD",
```
grabsefx, line 397: I beleive, it will not be enough just to remove defenition of gin index. It will not be deleted if it already exists. So "tOps" value should be changed to "DELETE" for completly removing unnecessary gin index

marcjohnson-kint: @julianladisch @adamdickmeiss  Please could this be confirmed, does RAML Module Builder need this to be left in and marked as delete rather than removed entirely?

julianladisch: https://github.com/folio-org/raml-module-builder#removing-an-index
> When upgrading a module via the Tenant API, an index is deleted if either "tOps": "DELETE" is set or the complete index entry is removed. Note that indexes are the only elements where removing the entry in schema.json removes them from the database.

julianladisch (approved):
Looks good, thanks!

PR #609, opened 2021-04-22
MODINVSTOR-521: Implement item shelving order migration.

bohdan-suprun:
## Approach:

Stream items from DB and update them using the `ItemService` using 100 records batches. 

## Changes:
* Include migration into TenantRefApi class;
* Implement `BatchedReadStream` - wrapper around `ReadStream` but it processes records using batches.


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -122,4 +149,30 @@ String servicePointUserFilter(String s) {
...
+    List<Future> startedMigrations = javaMigrations.stream()
+      .filter(javaMigration -> javaMigration.shouldExecuteMigration(ta))
+      .peek(migration -> log.info(
+        "Following migration is to be executed [migration={}]", migration))
+      .map(BaseMigrationService::runMigration)
+      .collect(Collectors.toList());
+
+    return CompositeFuture.all(startedMigrations).<Void>map(result -> null)
+      .onSuccess(notUsed -> log.info("Java migrations has been completed"))
+      .onFailure(error -> log.error("Some java migrations failed", error));
```
marcjohnson-kint, line 171: Does this mean that if the upgrade of these records fail, the module upgrade still succeeds?

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -94,7 +94,9 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
     @SuppressWarnings("all")
     final List<Future> setHridFutures = items.stream()
       .map(item -> {
-        item.getStatus().setDate(itemStatusDate);
+        if (item.getStatus().getDate() == null) {
+          item.getStatus().setDate(itemStatusDate);
+        }
```
marcjohnson-kint, line 99: Why did this need to change during this work?

src/main/java/org/folio/services/migration/BaseMigrationService.java
```diff
@@ -0,0 +1,28 @@
...
+  private final Versioned version;
+
+  protected BaseMigrationService(String fromVersion) {
+    this.version = versioned(fromVersion);
+  }
+
+  public boolean shouldExecuteMigration(TenantAttributes tenantAttributes) {
+    return isNotBlank(tenantAttributes.getModuleFrom())
+      && version.isNewForThisInstall(tenantAttributes.getModuleFrom());
+  }
```
marcjohnson-kint, line 19: This is a neat way of doing it

src/main/java/org/folio/services/migration/BatchedReadStream.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public BatchedReadStream<T> endHandler(Handler<Void> endHandler) {
+    delegate.endHandler(notUsed -> {
+      synchronized (this) {
+        // If any records remaining - process them
+        // Vert.x does not has any way to identify end of stream right now
+        // have to process in this way
+        if (buffer.size() > 0 && handler != null) {
+          // If an error occurred on last handle do not call end handler
+          // exception handler should be called in this case
```
marcjohnson-kint, line 73: Where is the exception handler invocation that is suggested by this comment?

src/main/java/org/folio/services/migration/BatchedReadStream.java
```diff
@@ -0,0 +1,112 @@
...
+
+    return this;
+  }
+
+  private synchronized void processRecord(T record) {
+      buffer.add(record);
+      if (buffer.size() >= batchSize) {
+        handleBuffer();
+      }
+  }
```
marcjohnson-kint, line 90: Does this code add records to the buffer until it reaches its limit and then handle buffer is invoked (which invokes the provided handler on the batch of records)?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -122,4 +149,30 @@ String servicePointUserFilter(String s) {
           return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
         });
   }
```
marcjohnson-kint, line 151: Does this mean that RAML Module Builder executes both the `loadData` and `postTenant` methods?

What order are they processed in?

cc: @adamdickmeiss @julianladisch

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -77,6 +89,21 @@ String servicePointUserFilter(String s) {
     return res;
   }
 
+  @Override
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers, Handler<AsyncResult<Response>> handler, Context context) {
```
marcjohnson-kint, line 93: I thought mod-inventory-storage also initialised some Kafka publishing when a tenant is enabled, do this happen elsewhere?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -122,4 +149,30 @@ String servicePointUserFilter(String s) {
...
+
+    @SuppressWarnings("rawtypes")
+    List<Future> startedMigrations = javaMigrations.stream()
+      .filter(javaMigration -> javaMigration.shouldExecuteMigration(ta))
+      .peek(migration -> log.info(
+        "Following migration is to be executed [migration={}]", migration))
+      .map(BaseMigrationService::runMigration)
+      .collect(Collectors.toList());
+
+    return CompositeFuture.all(startedMigrations).<Void>map(result -> null)
```
julianladisch, line 169: Using org.folio.okapi.common.GenericCompositeFuture avoids `@SuppressWarnings("rawtypes")`

julianladisch, line 169: `<Void>map(result -> null)` can be replaced by `mapEmpty()`

marcjohnson-kint (commented):
Thanks @bohdan-suprun for working on this so quickly

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: No, the error is still propagated to the caller, see `shouldStopProcessingWhenCannotConvertItem` test.

src/main/java/org/folio/services/item/ItemService.java
bohdan-suprun: Otherwise all items will have the status date overwritten, I don't think this is something expected. 

I also think there was a bug in this piece if code, because the butch sync API can also handle batch updates, but the date should be populated only on create, AFAIK.

src/main/java/org/folio/services/migration/BatchedReadStream.java
bohdan-suprun: Inside the `handleBuffer` method.

src/main/java/org/folio/services/migration/BatchedReadStream.java
bohdan-suprun: Yes, exactly what it does.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: I thought the loadData is executed all the time, but when I started to implement tests, I found out that loadData is not invoked in following case:

* Create tenant with `fromModuleVersion`: `mod-inventory-storage-1.0.0` (this version specified in tests);
* Then attempt to call tenant init for `fromModuleVersion`: `20.1.0` - in this case the loadData is not executed (there is a "magic" check inside the postTenant that skips loadData if `files.length` == 1 - I assume this means no DB changes).

Maybe @adamdickmeiss, @julianladisch can advice on this, I'm happy to remove the postTenant overridden

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: It was before postTenant, but then it was moved to loadData by folks.

src/main/java/org/folio/services/item/ItemService.java
marcjohnson-kint: I agree that this is likely a bug, thank you for discovering and fixing it

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: Thanks, changed

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
```diff
@@ -0,0 +1,97 @@
...
+    log.info("Starting item shelving order migration... ");
+
+    return postgresClient.startTx()
+      .compose(con -> postgresClient.selectStream(con, selectSql())
+        .compose(this::handleUpdate)
+        .onSuccess(records -> log.info(
+          "Shelving order migration has been completed [recordsProcessed={}]", records))
+        .onFailure(error -> log.error("Unable to complete shelving order migration", error))
+        .onComplete(result -> postgresClient.endTx(con)))
+      .map(notUsed -> null);
```
julianladisch: The transaction is not closed if selectStream or selectSql throws an exception.
Consider using
```
return postgresClient.withTransaction(con -> {
        return postgresClient.selectStream(con, selectSql())
                  .compose(this::handleUpdate));
    })
    .onSuccess(records -> log.info(
          "Shelving order migration has been completed [recordsProcessed={}]", records))
    .onFailure(error -> log.error("Unable to complete shelving order migration", error))
    .mapEmpty();
```

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: loadData should always been called when updating: https://github.com/folio-org/raml-module-builder#extending-the-tenant-api

julianladisch: What value do you set for `toModuleVersion`?

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
```diff
@@ -0,0 +1,97 @@
...
+    log.info("Starting item shelving order migration... ");
+
+    return postgresClient.startTx()
+      .compose(con -> postgresClient.selectStream(con, selectSql())
+        .compose(this::handleUpdate)
+        .onSuccess(records -> log.info(
+          "Shelving order migration has been completed [recordsProcessed={}]", records))
+        .onFailure(error -> log.error("Unable to complete shelving order migration", error))
+        .onComplete(result -> postgresClient.endTx(con)))
+      .mapEmpty();
```
julianladisch, line 55: The transaction is not closed if selectStream or selectSql throws an exception.
Consider using postgresClient.withTransaction:

```suggestion
    return postgresClient.withTransaction(con -> {
            return postgresClient.selectStream(con, selectSql())
                .compose(this::handleUpdate);
        })
        .onSuccess(records -> log.info(
          "Shelving order migration has been completed [recordsProcessed={}]", records))
        .onFailure(error -> log.error("Unable to complete shelving order migration", error))
        .mapEmpty();
```

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: I've tried both `20.2.1`, `20.2.0` no result. 

But a think I mentioned, the migration seems started but with some loong delay, when the test is being destroyed.

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: The `withTransaction` uses `PgConnection`, but `selectStream` requires `SQLConnection` is there any way to convert one to another?

src/main/java/org/folio/services/migration/BatchedReadStream.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public BatchedReadStream<T> exceptionHandler(Handler<Throwable> handler) {
+    this.exceptionHandler = handler;
+    delegate.exceptionHandler(handler);
+    return this;
+  }
+
+  @Override
+  public BatchedReadStream<T> handler(Handler<List<T>> handler) {
+    if (handler == null) {
```
MikhailFokanov, line 37: so, is there a way to clear the handler property of the class?

src/test/java/org/folio/services/migration/BatchedReadStreamTest.java
```diff
@@ -0,0 +1,173 @@
...
+    var batchedReadStream = new BatchedReadStream<>(rowStream);
+
+    batchedReadStream.handler(null);
+
+    verifyNoInteractions(rowStream);
+  }
+
+  @Test
+  public void shouldDelegateCallForExceptionHandler() {
+    var delegate = mockStream();
```
marcjohnson-kint, line 73: Does this mean that if a the exception handler is invoked for the `BatchReadStream` it invokes the exception handler for the `ReadStream`?

src/test/java/org/folio/services/migration/BatchedReadStreamTest.java
```diff
@@ -0,0 +1,173 @@
...
+    var batchedReadStream = new BatchedReadStream<>(delegate);
+
+    batchedReadStream.exceptionHandler(error -> {});
+
+    verify(delegate, times(1)).exceptionHandler(any());
+  }
```
marcjohnson-kint, line 79: Does this mean that if a the exception handler is invoked for the `BatchReadStream` it invokes the exception handler for the `ReadStream`?

src/test/java/org/folio/services/migration/BatchedReadStreamTest.java
```diff
@@ -0,0 +1,173 @@
...
+      var captor = ArgumentCaptor.forClass(List.class);
+      verify(endHandler, times(1)).handle(any());
+
+      verify(handler, times(3)).handle(captor.capture());
+      assertThat(captor.getAllValues()).hasSize(3);
+      assertThat(captor.getAllValues().get(0)).hasSize(batchSize);
+      assertThat(captor.getAllValues().get(1)).hasSize(batchSize);
+      assertThat(captor.getAllValues().get(2)).hasSize(1);
+    });
+  }
```
marcjohnson-kint, line 114: How is this test different to `shouldProcessRecordsByBatch`?

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
```diff
@@ -0,0 +1,97 @@
...
+      .endHandler(notUsed -> promise.tryComplete(recordsUpdated.get()))
+      .exceptionHandler(promise::tryFail)
+      .handler(rows -> {
+        var items = rows.stream()
+          .map(this::rowToItem)
+          .map(EffectiveCallNumberComponentsUtil::calculateAndSetEffectiveShelvingOrder)
+          .collect(Collectors.toList());
+
+        // Pause stream, so that updates is executed in sequence
+        batchStream.pause();
```
marcjohnson-kint, line 73: Why is the stream paused here rather than before the items are semi-processed?

If the goal is for only one handler to be invoked at a time, shouldn't the batch be paused before the handler is invoked or at least the start of the handler (which would leave a race condition if the stream continues, would it not)?

src/test/java/org/folio/services/migration/BatchedReadStreamTest.java
bohdan-suprun: Yes. In case if an unhandled error is occurred inside record handler, the exception handler is invoked also.

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+  private IndividualResource createItem(int index) {
+    var random = new Random();
+    UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
+
+    JsonObject itemToCreate = new JsonObject()
+      .put("id", UUID.randomUUID().toString())
+      .put("status", new JsonObject().put("name", "Available"))
+      .put("holdingsRecordId", holdingsRecordId.toString())
+      .put("barcode", random.nextLong())
```
julianladisch, line 75: There is a small probability that this produces a duplicate barcode causing a failure.
Consider using `UUID.randomUUID().toString()` or `index` or `100000` + `index`.

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: As far as I understand the stream never processed concurrently.

The `pause` is needed in order to wait for update of items to be finished, if we don't pause the stream than updates will be submitted and stream finished without confirming that updates succeeded/failed.

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
bohdan-suprun: Changed to `100000` + `index`

src/test/java/org/folio/services/migration/BatchedReadStreamTest.java
bohdan-suprun: Removed this one

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
marcjohnson-kint: > As far as I understand the stream never processed concurrently.

If the stream never invokes multiple handlers concurrently, how could the updates be executed out of sequence (referred to in the comment)?

> The pause is needed in order to wait for update of items to be finished, if we don't pause the stream than updates will be submitted and stream finished without confirming that updates succeeded/failed.

I don't think I follow. The underlying stream needs to be paused because if that stream finishes before the batched stream finishes processing then updates will be lost?

src/main/java/org/folio/services/migration/BatchedReadStream.java
bohdan-suprun: Thanks, changed. Now handler is also cleared on delegate.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: Thanks.

marcjohnson-kint: Do you know why it was moved to `loadData`? Was it to only have one of either `loadData` or `postTenant`?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: > But a think I mentioned, the migration seems started but with some loong delay, when the test is being destroyed.

How was this observed? From logs during the tests?

marcjohnson-kint: > Create tenant with fromModuleVersion: mod-inventory-storage-1.0.0 (this version specified in tests);

Which tests define this, the general test setup or specific tests for this process?

julianladisch: The migration runs asynchronously: https://github.com/folio-org/raml-module-builder#extending-the-tenant-api

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: No, I don't know. Maybe loadData is designed for such thinks, but again it doesn't work for me unfortunately

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: > I don't think I follow. The underlying stream needs to be paused because if that stream finishes before the batched stream finishes processing then updates will be lost?

No, it is not related to the batched stream. Let's say we have following code:

```java
public User updateUser(User user) {
  updateUserLoginInfo(user.getLoginInfo());
  persistUser(user);

  return user;
}

private Future<Void> updateUserLoginInfo(UserLoginInfo login);
private Future<Void> persistUser(User user);
```

When we do something `var updatedUser = updateUser(user)` then at the time when we returned from the method the user most likely is not persist yet nor login info is updated because these two methods returns future, that means that result will be available after some while once the future is completed, right?

But if we still use this approach we can not guarantee that update is successful. 

That's why I pause/resume stream: I pause the stream before starting update, than wait for computable future is completed and depending on the result I resume or stop the stream.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: The postTenant is removed, loadData started to work for me.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: It works now! Thanks.

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -294,6 +295,10 @@ public static void tenantOp(String tenantId, JsonObject job) throws InterruptedE
       failureMessage = String.format("Tenant get failed: %s: %s",
           response.getStatusCode(), response.getBody());
 
+      if (response.getJson().containsKey("error")) {
+        throw new IllegalStateException(response.getJson().getString("error"));
+      }
```
marcjohnson-kint, line 300: Is the response body always JSON?

If not, could `response.getJson()` throw an exception?

julianladisch (approved):
Looks good, thanks!

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
marcjohnson-kint: > When we do something var updatedUser = updateUser(user) then at the time when we returned from the method the user most likely is not persist yet nor login info is updated because these two methods returns future, that means that result will be available after some while once the future is completed, right?

Sure

> But if we still use this approach we can not guarantee that update is successful.

If we use what approach?

> That's why I pause/resume stream: I pause the stream before starting update, then wait for computable future is completed and depending on the result I resume or stop the stream.

I'm still not sure I understand why this is important.

I understand the need to use a continuation of the future to decide on whether the process is a success or not.

Why does it matter whether the stream has been paused? If no additional handlers are invoked until this one finishes, all that will happen is the stream will grow. Is that the case? Is that the undesirable behaviour?

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
```diff
@@ -53,7 +49,7 @@ public void shouldStopProcessingWhenCannotConvertItem() throws Exception {
   private JsonObject getTenantAttributes() {
     return JsonObject.mapFrom(new TenantAttributes()
       .withModuleFrom("20.1.1")
-      .withModuleTo("20.2.0")
+      .withModuleTo("20.2." + nextPatch.incrementAndGet())
```
marcjohnson-kint, line 52: Why does this process start with version `20.2.1` rather than `20.2.0` (due to the AtomicInteger starting at 1)?

Why does this need to be incremented each time?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: How do you know it works now?

bohdan-suprun: Otherwise all tests in `org.folio.rest.api.ItemShelvingOrderMigrationServiceApiTest` are failed.

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
bohdan-suprun: Because module already on 20.2.0.

We have to increment them because the method is used by multiple tests but the tenant is for a entire test class.

When 1st test passed tenant is on 20.2.1, then second should be on 20.2.2, etc.

src/test/java/org/folio/rest/api/StorageTestSuite.java
bohdan-suprun: I've added additional status check, for 200 it is always JSON.

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
marcjohnson-kint: Oh, these tests are pretending to enable versions 20.2.1, 20.2.2 etc to mimic enabling 20.2.0?

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: > If no additional handlers are invoked until this one finishes, all that will happen is the stream will grow.

The think is that update operations are handled on vertx thread pool, so once the `updateItem` is called handler will immediately continue their processing.

At some point the stream will be finished but some item updates will be still pending. I'm not sure if it is allowed behavior or not.

For reindex we don't wait and pause stream, and no memory issues observed.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: I don't follow. Were those tests not present and passing in the first commit in this pull request?

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
bohdan-suprun: yes, correct.

marcjohnson-kint: Could they at least start from `20.2.0` so they also issue the expected instructions?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: No, they were present, but migrations called from `postTenant` but now migrations called from `loadData` method and no need to override postTenant

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: > That is the single handler that processes a batch of records?
Yes.

> Then why pause the stream here and not in that process?

Un which process, item update?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: Ok. At some intermediary point, when you tried moving the code from `postTenant` to `loadData`, it stopped working and now it works again?

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
marcjohnson-kint: > Un which process, item update?

I thought you said that the Re-index process does not pause the underlying stream during it's processing, yet this one (to update items) does. Did I misunderstand?

src/main/java/org/folio/rest/impl/TenantRefAPI.java
bohdan-suprun: Yes, seems the issue was that I did not set toModule, and it treated as current module version `20.2.0`

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
bohdan-suprun: Unfortunately no, because module already on 20.2.0 after tenant set-up by @BeforeAll method.

bohdan-suprun: However there is a unit test `org.folio.services.migration.item.ItemShelvingOrderMigrationServiceTest` that demonstrates that migration is triggered for 20.1.0 -> 20.2.0 migration.

src/main/java/org/folio/services/migration/item/ItemShelvingOrderMigrationService.java
bohdan-suprun: Reindex is fully asynchronous and it is processed in background.

If it is acceptable that the stream is processed and tenant init is marked as completed, even though there might be some pending updates, than I'm happy to remove pause/resume logic.

marcjohnson-kint: > Reindex is fully asynchronous and it is processed in background.

From your explanation, the purpose of the pause is to try to limit the memory usage during the process (rather than the ordering, which the comment referred to). I don't understand how whether the process runs synchronously or asynchronously relates to that.

I also thought that the Tenant API 2.0 runs asynchronously by default (it can be forced synchronous by a flag), with Okapi polling a status endpoint. Is that not the case?

> If it is acceptable that the stream is processed and tenant init is marked as completed, even though there might be some pending updates, than I'm happy to remove pause/resume logic.

I wasn't suggesting that the pause be removed. I was trying to understand why it was needed, as I couldn't understand the ordering comment when you also stated that only one handler instance is executed concurrently.

Anyway, given the meeting that @SergiyVSergiyenko has raised for this morning. I should probably stop asking questions. Please feel free to ignore.

PR #615, opened 2021-04-27
Release mod-inventory-storage 20.2.1

bohdan-suprun:


marcjohnson-kint (commented):
As this is on the long lived branch, is this work only to be applied in 20.20.1 and not the next mainline release?

bohdan-suprun:
I'll back-port it to master after release.

marcjohnson-kint:
> I'll back-port it to master after release.

Ok, thanks for confirming

adamdickmeiss:
> I'll back-port it to master after release.

For things going in both places.. I prefer to start with master. Then the change can be verified in snapshot-builds etc.. Then cherry-pick to branch after. Because in a way release branches are even less tested!

src/test/java/org/folio/services/migration/item/ItemShelvingOrderMigrationServiceTest.java
```diff
@@ -30,7 +31,7 @@ public void shouldTriggerMigration(String from, String to) {
   }
 
   @Parameters({
-    "20.2.0, 20.2.1",
+    "20.2.1, 20.2.2",
```
marcjohnson-kint, line 34: Is this because the initial enablement now starts at 20.2.1?

(and does this mean that these tests will need to change every time a new version is released?)

bohdan-suprun: > Is this because the initial enablement now starts at 20.2.1?

Yes

> (and does this mean that these tests will need to change every time a new version is released?)

No, unless we change from module version for the script.

bohdan-suprun:
> For things going in both places.. I prefer to start with master. Then the change can be verified in snapshot-builds etc.. Then cherry-pick to branch after. Because in a way release branches are even less tested!

I'll test it on scratch + we have integration tests for this change.

marcjohnson-kint:
> For things going in both places.. I prefer to start with master. Then the change can be verified in snapshot-builds etc.. Then cherry-pick to branch after. Because in a way release branches are even less tested!

I do too. I believe the reason for this approach here is due to urgency around the 2021 R1 release.

bohdan-suprun:
@marcjohnson-kint, @adamdickmeiss https://github.com/folio-org/mod-inventory-storage/pull/616/files has been created to port changes to master, could you please review? 

The changes was verified on scratch env and on perf env.

src/test/java/org/folio/services/migration/item/ItemShelvingOrderMigrationServiceTest.java
marcjohnson-kint: > No, unless we change from module version for the script.

Ok, thanks for confirming that

marcjohnson-kint (approved):

marcjohnson-kint (commented):
Approving to expedite the release process. I have no feedback for the code changes.

In general, I think that disabling domain events during the upgrade is rather contrary to the intent of them. In effect, the change does not exist to the consumers of those message types, meaning that those consumers could be out of sync for a undeterminable amount of time.

I understand why you have chosen to do so in this case, in order to improve performance, based upon the assumption that the only consumer is the ES based search (we cannot guarantee that). That may be ok because it does not currently care about that property and that there was a significant performance impact on the upgrade from the the ES based search making requests to inventory as a reaction to those messages.

PR #616, opened 2021-04-28
MODINVSTOR-521: Do not send domain events for item shelving order migration

bohdan-suprun:
Back port from https://github.com/folio-org/mod-inventory-storage/pull/615

### DIFF
--- a/src/main/java/org/folio/persist/AbstractRepository.java
+++ b/src/main/java/org/folio/persist/AbstractRepository.java
@@ -82,6 +82,14 @@ public abstract class AbstractRepository<T> {
     return promise.future();
   }
 
+  public Future<RowSet<Row>> update(List<T> records) {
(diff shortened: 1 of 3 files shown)


marcjohnson-kint (approved):
Approving to get the changes made during the release onto the mainline quickly. I have no feedback for the code changes.

In general, I think that disabling domain events during the upgrade is rather contrary to the intent of them. In effect, the change does not exist to the consumers of those message types, meaning that those consumers could be out of sync for a undeterminable amount of time.

I understand why you have chosen to do so in this case, in order to improve performance, based upon the assumption that the only consumer is the ES based search (we cannot guarantee that). That may be ok because it does not currently care about that property and that there was a significant performance impact on the upgrade from the the ES based search making requests to inventory as a reaction to those messages.

PR #618, opened 2021-05-11
Modinvstor 702 bound with

nielserik:
**Purpose** 
Create data structure for holding bound-withs, that is, to have a bound-with item that contains multiple titles (Instances).

Implement a database structure containing the holdings of titles that are bound together, and an API to create, update, delete, and retrieve the bound-with parts.

The table will contain a row for each holdings record that is included with a bound-with Item. The single Item with multiple holdings records will thus represent the circulating entity, like any other Item in Inventory. Outside of this new table, the Item will additionally be attached to one of the holdings records in the bound-with by it's mandatory holdingsRecordId. This particular holdings record, or rather the title that it's a holding of, will be used when naming the bound-with in displays - like "The Primary Title (and other titles)"

More details on the suggested design and the developments related to this in mod-inventory and ui-inventory: https://wiki.folio.org/display/MM/Bound-with+data+model

**Other comments**
This PR would create a raw data structure, just enough to persist bound-with information. There are issues filed with mod-inventory for making the bound-with properties conveniently accessible for clients, ie ui-inventory.


descriptors/ModuleDescriptor-template.json
```diff
@@ -1056,6 +1056,33 @@
...
+          "methods": ["PUT"],
+          "pathPattern": "/bound-with-parts/{id}",
+          "permissionsRequired": ["inventory-storage.bound-with-parts.item.put"]
+        }, {
+          "methods": ["DELETE"],
+          "pathPattern": "/bound-with-parts/{id}",
+          "permissionsRequired": ["inventory-storage.bound-with-parts.item.delete"]
+        }
+      ]
+    },
```
marcjohnson-kint, line 1085: As I understand it, this interface is specific to an aspect of inventory. I think it could be useful if the paths used in the interface (and maybe the id) reflected that e.g. `inventory-storage/bound-with-parts` if this is intended to specifically be a storage level interface.

Are there intended to be endpoints for this information in business logic interfaces or is the intention that is all hidden behind another of the inventory business logic interfaces?

ramls/bound-with-part.raml
```diff
@@ -0,0 +1,48 @@
...
+
+resourceTypes:
+  collection: !include raml-util/rtypes/collection.raml
+  collection-item: !include raml-util/rtypes/item-collection.raml
+
+/bound-with-parts:
+  type:
+    collection:
+      exampleCollection: !include examples/boundwithparts.json
+      exampleItem: !include examples/boundwithpart.json
```
marcjohnson-kint, line 30: Personally I find it easier to read multi-word file names when there are hyphens between the words

src/main/java/org/folio/rest/impl/BoundWithPartAPI.java
```diff
@@ -0,0 +1,51 @@
...
+  public void getBoundWithParts(String query, @Min(0) @Max(2147483647) int offset, @Min(0) @Max(2147483647) int limit, @Pattern(regexp = "[a-zA-Z]{2}") String lang, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.get(BOUND_WITH_TABLE, BoundWithPart.class, BoundWithParts.class, query, offset, limit,
+      okapiHeaders, vertxContext, org.folio.rest.jaxrs.resource.BoundWithParts.GetBoundWithPartsResponse.class, asyncResultHandler);
+  }
+
+  @Override
+  public void postBoundWithParts(@Pattern(regexp = "[a-zA-Z]{2}") String lang, BoundWithPart entity, Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+    PgUtil.post(BOUND_WITH_TABLE, entity, okapiHeaders, vertxContext,
+      PostBoundWithPartsResponse.class, asyncResultHandler);
+
```
marcjohnson-kint, line 29: Personally, I think this blank line (and those at the end of other methods) is redundant

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -106,6 +107,7 @@ public static void testBaseBeforeClass() throws Exception {
     instanceRelationshipsClient = ResourceClient.forInstanceRelationships(client);
     instanceRelationshipTypesClient = ResourceClient.forInstanceRelationshipTypes(client);
     precedingSucceedingTitleClient = ResourceClient.forPrecedingSucceedingTitles(client);
+    boundWithPartsClient = ResourceClient.forBoundWithParts(client);
```
marcjohnson-kint, line 110: Is this used in any tests other than src/test/java/org/folio/rest/api/BoundWithStorageTest.java?

If not, I suggest adding it that class instead of the base class.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+import static org.folio.rest.support.http.InterfaceUrls.*;
+import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@RunWith(JUnitParamsRunner.class)
+public class BoundWithStorageTest extends TestBaseWithInventoryUtil {
+
+  // see also @BeforeClass TestBaseWithInventoryUtil.beforeAny()
+
```
marcjohnson-kint, line 26: I don't think this comment is needed.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
```diff
@@ -0,0 +1,139 @@
...
+    IndividualResource mainHoldingsRecord = createHoldingsRecord(mainInstance.getId());
+    IndividualResource item = createItem(mainHoldingsRecord.getId());
+
+    IndividualResource anotherInstance = createInstance("Another Instance");
+    IndividualResource anotherHoldingsRecord = createHoldingsRecord(anotherInstance.getId());
+    IndividualResource aThirdInstance = createInstance("Third Instance");
+    IndividualResource aThirdHoldingsRecord = createHoldingsRecord(aThirdInstance.getId());
+
+    // Make 'item' a bound-with
+    boundWithPartsClient.create(createBoundWithPartJson(mainHoldingsRecord.getId(),item.getId()));
```
marcjohnson-kint, line 47: Does this mean that a bound-with item is intended to have a bound-with parts record for the primary holding as well as the secondary holdings?

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+    List<JsonObject> getAllPartsForBoundWithItem = boundWithPartsClient.getByQuery("?query=itemId==" + item.getId());
+    List<JsonObject> part2 = boundWithPartsClient.getByQuery("?query=holdingsRecordId==" + holdingsRecord2.getId()+"");
+
+    IndividualResource instance3 = createInstance("Instance 3");
+    IndividualResource holdingsRecord3 = createHoldingsRecord(instance3.getId());
+
+    Response updateResponse = boundWithPartsClient.attemptToReplace(
+      part2Created.getId(),
+      createBoundWithPartJson(holdingsRecord3.getId(), item.getId()));
```
marcjohnson-kint, line 91: What would happen if this changed the record to refer to the same holdings record as the first part?

Is it allowed that a bound-with can have multiple parts that refer to the same holdings record?

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Yes.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: That would be duplicate entries, since, besides the primary key UUID, there are only two columns in the table, `itemId`and `holdingsRecordId`.  

So it would basically state twice that this title is part of this bound-with. That would not be disastrous perhaps, but messy. If we created a unique index on the two columns, the database would prevent it from happening.

descriptors/ModuleDescriptor-template.json
nielserik: The expected usages of /inventory-storage/bound-with-parts in business logic interfaces are:

* existing API /inventory/instances  will query inventory-storage/bound-with-parts to set the intance.isBoundWith flag property (so hidden behind)
* existing API /inventory/items will query inventory-storage/bound-with-parts to set the item.isBoundWith flag property (hidden behind)
* new API /inventory/bound-with-items  will GET  bound-with item(s) for a holdingsRecordId.  It would respond with objects on the regular Inventory Item schema. I would be used by ui-inventory to populate the item list under the holdings record in the Instance detail view, since said items would not be found by the standard way of querying the /inventory/items API by holdingsRecordId   (so that's a BL endpoint of sorts but not a straight forward propagation of the storage end-point)
* I'm imagining at least one other API  /inventory/bound-withs   which would return composite bound-with objects to be used by the UI for displaying the entire bound-with or multiple elements hereof. The composite would thus hold: The item, all the holdings, and all the Instances of the bound-with. There are UI mock-ups showing lists of titles in this fashion and compiling that information would not be trivial for the UI. 
* I haven't quite thought through what APIs might be useful for editing bound-withs through the UI. The raw /inventory-storage/bound-with-parts might even suffice, I don't yet know.

ramls/bound-with-part.raml
nielserik: Yeah, I tried to follow what looked like the de-facto naming convention but I don't disagree with the preference.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Though, it really comes down to how you decide to populate the data structure. You could decide to _not_  include the primary holding, but to fetch the complete set of titles in the bound-with, you would then have to remember to go out and find that directly linked holding separately.

descriptors/ModuleDescriptor-template.json
marcjohnson-kint: > existing API /inventory/instances will query inventory-storage/bound-with-parts to set the intance.isBoundWith flag property (so hidden behind)

Why is it important that an instance be bound-with something else? Surely some copies of an instance might be bound-with copies of another instance, whilst other might not be.

> new API /inventory/bound-with-items will GET bound-with item(s) for a holdingsRecordId. It would respond with objects on the regular Inventory Item schema. I would be used by ui-inventory to populate the item list under the holdings record in the Instance detail view, since said items would not be found by the standard way of querying the /inventory/items API by holdingsRecordId (so that's a BL endpoint of sorts but not a straight forward propagation of the storage end-point)

Is this instead of introducing an API for holdings and including that information in the response to that API?

ramls/bound-with-part.raml
marcjohnson-kint: > Yeah, I tried to follow what looked like the de-facto naming convention but I don't disagree with the preference.

It's challenging because unless someone changes all of the examples at once (which folks don't have time to do), we only compound the (less readable IMO) convention

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
marcjohnson-kint: > Though, it really comes down to how you decide to populate the data structure. You could decide to _not_ include the primary holding, but to fetch the complete set of titles in the bound-with, you would then have to remember to go out and find that directly linked holding separately.

Who is `you` in these statements?

Are you suggesting that it will be up to each client to decide which way they want to do it, meaning that all clients would need to accommodate both situations when interpreting the data?

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
marcjohnson-kint: > That would not be disastrous perhaps, but messy. If we created a unique index on the two columns, the database would prevent it from happening.

Is that something you are considering as part of this work? If not, are we expecting all clients will accommodate these duplicates?

ramls/bound-with-part.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Records the relationship between a part of a bound-with (a holdings-record) and the bound-with as a whole (the circulatable item)",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "unique ID of the recorded bound-with relation; a UUID"
```
marcjohnson-kint, line 8: This comment suggests that this is intended to be a UUID, could it be worth including a constraint for this property that includes this in the machine readable part of the schema?

descriptors/ModuleDescriptor-template.json
nielserik: Yes (some copies of an instance might be bound-with while others might not). This is a question for the Inventory PO and the GBV based working group) 

No, I have not thought of it as an alternative to an Inventory holdings API. I suppose. I reckon you mean, that if there was a holdings API in Inventory, it could conceivably retrieve both bound-with items and non-bound-with items behind the seems and serve them up in the composite?  That would be possible. It would require slightly bigger changes in the Inventory UI I suppose.

ramls/bound-with-part.raml
nielserik: Right, agreed, I changed it to hyphens.

src/main/java/org/folio/rest/impl/BoundWithPartAPI.java
nielserik: Was removed after your review.

src/test/java/org/folio/rest/api/TestBase.java
nielserik: Was moved away from base class after your review.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Was removed after your review.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Well, the business logic API can take care of always including all the holdings for an item and never include the primary holdings record more than once, regardless of whether the primary holdings record has an entry in the bound-with table. If we wanted that behavior just in case.

Clients that used the storage directly would have to pull the data out with an understanding of how it populated the data in the first place. To make the same client work for different population policies, against the raw storage, that client would need to do something similar to what the BL could do as described.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: I don't think clients need to accommodate the duplicates. They are just redundant. If they were created manually in the admin, I reckon the librarian would soon discover the typo and delete the duplicate.

ramls/bound-with-part.json
nielserik: Yes, that would be tidy. Done.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: > Is that something you are considering as part of this work? If not, are we expecting all clients will accommodate these duplicates?

descriptors/ModuleDescriptor-template.json
nielserik: Hmm, I changed the API path to /inventory-storage/bound-with-parts last week, but not sure if that becomes clear. The code snippets in the conversation doesn't seem to reflect the changes made - except for the flag "Outdated".  But code was amended anyways.

marcjohnson-kint: > Yes (some copies of an instance might be bound-with while others might not). This is a question for the Inventory PO and the GBV based working group I think. They asked for flags anyway - alerts of a sort.

Do you need an answer to that question prior to designing the inventory API changes? I would think so, given that it might mean that they don't want this indicated on the instance (which describes an abstract notion that isn't actually bound to anything)

> No, I have not thought of it as an alternative to an Inventory holdings API. I reckon you mean that if there was a holdings API in Inventory, it could conceivably retrieve both bound-with items and non-bound-with items behind the seems and serve them up in a composite of holdings and items? That would be possible. It would possibly require slightly bigger changes in the Inventory UI.

Yes, that is what I was thinking. It was based upon the assumption that it was holdings where information about bound-with would be presented, given that it is these kinds of records we chose to represent that.

ramls/bound-with-part.raml
marcjohnson-kint: Thank you

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: Thank you

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
marcjohnson-kint: > Well, the business logic API can take care of always including all the holdings for an item and never include the primary holdings record more than once, regardless of whether the primary holdings record has an entry in the bound-with table. If we wanted that behavior just in case.

That assumes that the clients will only be using the business logic APIs.

> Clients that used the storage directly would have to pull the data out with an understanding of how it populated the data in the first place. 

How would those clients (which might not be the clients that stored it) know how it was populated?

> To make the same client work for different population policies, against the raw storage, that client would need to do something similar to what the BL could do as described.

Yes, that is what I was suggesting, with this ambivalent approach, every client needs (the additional complexity) to be able to interpret either approach to structuring this data.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
marcjohnson-kint: > I don't think clients need to accommodate the duplicates. They are just redundant. If they were created manually in the admin, I reckon the librarian would soon discover the typo and delete the duplicate.

We are expecting the people using the system to correct these kinds of issues?

I thought it was mentioned in the meeting we had that the first version of this functionality would not allow editing via the front-end. If that is the case, how would those folks do this?

> If RMB supported composite unique indexes I would just add it but last time I checked I couldn't find such support. We could add an SQL snippet for it but personally I don't feel it's worth the added complexity.

Something like [this](https://github.com/folio-org/mod-circulation-storage/blob/667443265f2a5d427bc8a8bc14dbd2b2f920aeaa/src/main/resources/templates/db_scripts/schema.json#L144-L148) perhaps?

ramls/bound-with-part.json
```diff
@@ -0,0 +1,30 @@
...
+      "$ref": "uuid.json"
+    },
+    "holdingsRecordId": {
+      "type": "string",
+      "description": "the ID of the holdings record representing a part of a bound-with; a UUID"
+    },
+    "itemId" : {
+      "type": "string",
+      "description": "the ID of the item representing the bind; a UUID"
+    },
```
marcjohnson-kint, line 18: Are these also expected to be UUID properties given they reference `id` properties that are intended to be UUIDs?

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
```diff
@@ -0,0 +1,139 @@
...
+    boundWithPartsClient.create(createBoundWithPartJson(mainHoldingsRecord.getId(),item.getId()));
+    IndividualResource secondPart = boundWithPartsClient.create(createBoundWithPartJson(anotherHoldingsRecord.getId(),item.getId()));
+    boundWithPartsClient.create(createBoundWithPartJson(aThirdHoldingsRecord.getId(),item.getId()));
+
+    Response boundWithGETResponseForPartById = boundWithPartsClient.getById(secondPart.getId());
+
+    List<JsonObject> getAllPartsForBoundWithItem = boundWithPartsClient.getByQuery("?query=itemId==" + item.getId());
+
+    assertThat(boundWithGETResponseForPartById.getStatusCode(), is(HttpURLConnection.HTTP_OK));
+    assertThat(getAllPartsForBoundWithItem.size(), is(3));
```
marcjohnson-kint, line 56: This assertions suggest we only care that there are 3 bound-with entries for this item and not anything else about those entries, is that the case?

ramls/bound-with-part.json
nielserik: Yes, but if you don't put UUID's in them it will fail already on the referential constraint.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Yes. there are not any properties in these, it's all primary key and foreign keys.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
nielserik: Thank you, that looks like it would do what I'd want.

nielserik: > I thought it was mentioned in the meeting we had that the first version of this functionality would not allow editing via the front-end. If that is the case, how would those folks do this?

Without the editing option folks will not create any bound-withs and thus not attach the same title twice. 

The first client populating this structure will likely be GBV and mod-inventory-update.

nielserik: Unique index done.

PR #619, opened 2021-05-13
MODINVSTOR-723: Add publication year and publication period properties.

bohdan-suprun:
Add `publicationPeriod` properties to `instance`.

As per Magda:
>  I believe that implementation can be achieved by having two fields: year1 and year2 and from your example above it would be populated as follows:

> Record A has a publication year of 1999, no publication start year and no publication end year – year1 is populated with 1999, year2 is blank
Record B has no publication year, a publication start year of 1996 and publication end year of 2001 – year1 is 1996 and year2 is 2001
Record C has no publication year, a publication start year of 1998 and no publication end year – year1 is 1998 and year2 is blank
Record D has no publication year, no publication start year and a publication end year of 2000 – year1 is blank year2 is 2000

### DIFF
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -215,6 +215,25 @@
             "type": "string",
             "description": "Date (year YYYY) of publication, distribution, etc."
           },
+          "publicationYear": {
+            "type": "integer",
+            "description": "Publication year"
+          },
+          "publicationPeriod": {
+            "type": "object",
+            "description": "Publication period",
+            "properties": {
+              "start": {
+                "type": "integer",
+                "description": "Publication start year"
+              },
+              "end": {
+                "type": "integer",
+                "description": "Publication end year"
+              }
+            },
+            "additionalProperties": false
+          },
           "role": {
             "type": "string",
             "description": "The role of the publisher, distributor, etc."
--- a/ramls/instance.json
+++ b/ramls/instance.json
@@ -215,21 +215,6 @@
             "type": "string",
             "description": "Date (year YYYY) of publication, distribution, etc."
           },
-          "publicationPeriod": {
-            "type": "object",
-            "description": "Publication period",
-            "properties": {
-              "start": {
-                "type": "integer",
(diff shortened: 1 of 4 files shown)


hjiebsco (approved):
Is there a need to update existing data with a script?

bohdan-suprun:
> Is there a need to update existing data with a script?

Not yet, it is part of https://issues.folio.org/browse/MODINVSTOR-724

ramls/instance.json
```diff
@@ -215,6 +215,25 @@
...
+            "properties": {
+              "start": {
+                "type": "integer",
+                "description": "Publication start year"
+              },
+              "end": {
+                "type": "integer",
+                "description": "Publication end year"
+              }
+            },
```
marcjohnson-kint, line 234: I'm aware that Magda referred to these ideas in a comment, yet the issue itself hasn't been updated to include them, has it been confirmed that these new properties are needed and that they are definitely only years (and hence an integer)?

ramls/instance.json
```diff
@@ -215,6 +215,25 @@
...
+            "additionalProperties": false
+          },
```
marcjohnson-kint, line 236: My interpretation of the JIRA issue is that there is only a single new field intended, that could have different values depending upon if there is a single date or a range, rather than multiple new fields. 

Have these new fields been confirmed with Magda?

bohdan-suprun: Yes, it is possible to have a specific year or a period.

marcjohnson-kint: Ok, that seemed to me to be really quite different from what the issue originally asked for. I think I'm likely missing a lot of context and likely shouldn't have got involved at all (my concern was we were adding properties that wouldn't be used).

I'll reply to Magda's comment on the issue to clarify my understanding and then leave the Falcon team to it. Apologies for barging in :-(

MikhailFokanov (approved):
Looks good.

PR #620, opened 2021-05-14
MODINVSTOR-633

alen-yao:
Add instance formats audio belt


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1035,6 +1035,11 @@
       "run": "after",
       "snippetPath": "inventory-hierarchy/addEffectiveHoldingsToItemsAndHoldingsView.sql",
       "fromModuleVersion": "20.2.0"
+	},
+	{
+      "run": "after",
+      "snippetPath": "addInstanceFormatsAudioBelt.sql",
+      "fromModuleVersion": "20.2.1"
```
marcjohnson-kint, line 1042: The next planned release for mod-inventory-storage is 20.3.0, which means I think this should be 20.3.0

```suggestion
      "fromModuleVersion": "20.3.0"
```

marcjohnson-kint (commented):
Please add a test for the upgrade process (like in #596)

src/main/resources/templates/db_scripts/schema.json
alen-yao: okay

marcjohnson-kint (approved):

PR #621, opened 2021-05-14
POST /inventory-hierarchy/items-and-holdings Not Handling Errors Gracefully

K-Felk:
This issue was happening because the sql script that assembles data for
the endpoint would crash when encountering invalid statisticalcodeid 
values, leaving the database connection in an open, aborted state.
I addressed this by making the calling code explicitly close database 
connections when encountering an error, ensuring that the open aborted 
connections would be properly closed.

https://issues.folio.org/browse/MODINVSTOR-677


src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java
```diff
@@ -99,7 +99,17 @@ protected void fetchRecordsByQuery(String sql, Supplier<Tuple> paramsSupplier, R
...
+        .exceptionHandler(e -> {
+          respondWithError(response, e, asyncResultHandler);
+          //database connections must be explicitly closed when an error is thrown 
+          //or it will be left in an open aborted state.
+          postgresClient.endTx(tx, h -> {
+            if (h.failed()) {
+              log.error("Unable to close database connection: " + h.cause().getMessage());
+              return;
+            }
+          });
```
marcjohnson-kint, line 111: Does this mean that when the `exception handler` delegate is triggered the `end handler` delegate is not?

src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java
```diff
@@ -99,7 +99,17 @@ protected void fetchRecordsByQuery(String sql, Supplier<Tuple> paramsSupplier, R
...
-        .exceptionHandler(e -> respondWithError(response, e, asyncResultHandler))
+        .exceptionHandler(e -> {
+          respondWithError(response, e, asyncResultHandler);
+          //database connections must be explicitly closed when an error is thrown 
+          //or it will be left in an open aborted state.
+          postgresClient.endTx(tx, h -> {
+            if (h.failed()) {
+              log.error("Unable to close database connection: " + h.cause().getMessage());
+              return;
+            }
```
marcjohnson-kint, line 110: Could this be extracted into a separate method / class that could be reused (in other places where this could occur) and tested independently (to improve the coverage), maybe by executing some SQL known to abort to the connection?

src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java
K-Felk: That is behavior I observed.  I traced the actual 500 error being written out to this code.  When I added code to it that explicitly closed the database connection, I stopped seeing aborted connections in PG admin.

src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java
K-Felk: I'll try.

src/main/java/org/folio/rest/impl/AbstractInstanceRecordsAPI.java
marcjohnson-kint: Thank you for confirming that.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 ## 20.3.0 IN-PROGRESS
 
+* Added code to close database connections when encountering errors from the /inventory-hierarchy endpoint (MODINVSTOR-677)
```
marcjohnson-kint, line 3: I prefer that release notes describe the way the change impacts the clients / users of the module

```suggestion
* Closes database connections when encountering errors in the /inventory-hierarchy endpoint (MODINVSTOR-677)
```

marcjohnson-kint (commented):
Thank you for extracting the error handler and testing it separately

src/main/java/org/folio/rest/impl/ConnectionErrorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+public class ConnectionErrorHandler {
+  protected static final Logger log = LogManager.getLogger();
+  protected final PostgresClient client;
+
+  public ConnectionErrorHandler(PostgresClient Pgclient) {
+    this.client = Pgclient;
+  }
+
+  public void writeErrorAndCloseConn(Throwable t, AsyncResult<SQLConnection> conn, HttpServerResponse dataResponse, Handler<AsyncResult<Response>> asyncResultHandler) {
```
marcjohnson-kint, line 22:
```suggestion
  public void writeErrorAndCloseConn(Throwable t, 
      AsyncResult<SQLConnection> conn, HttpServerResponse dataResponse, Handler<AsyncResult<Response>> asyncResultHandler) {
```

src/main/java/org/folio/rest/impl/ConnectionErrorHandler.java
```diff
@@ -0,0 +1,41 @@
...
+    respondWithError(dataResponse, t, asyncResultHandler);
+    client.endTx(conn, h -> {
+      if (h.failed()) {
+        log.error("Unable to close database connection: " + h.cause().getMessage());
+      }
+    });
+  }
+
+  private static void respondWithError(HttpServerResponse dataResponse, Throwable t, Handler<AsyncResult<Response>> asyncResultHandler) {
```
marcjohnson-kint, line 31:
```suggestion
  private static void respondWithError(HttpServerResponse dataResponse, Throwable t, 
    Handler<AsyncResult<Response>> asyncResultHandler) {
```

src/test/java/org/folio/rest/api/InventoryHierarchyViewTest.java
```diff
@@ -95,6 +95,26 @@ public void setUp() {
...
+    setFaultyStatisticalCode();    
+
+    params.put(QUERY_PARAM_NAME_SKIP_SUPPRESSED_FROM_DISCOVERY_RECORDS, "false");
+
+    List<JsonObject> instances = requestInventoryHierarchyViewUpdatedInstanceIds(params, response -> assertThat(response.getStatusCode(), is(200)));
+    UUID[] instanceIds = instances.stream()
+      .map(json -> UUID.fromString(json.getString("instanceId")))
+      .toArray(UUID[]::new);
+    
+    List<JsonObject> instancesData = requestInventoryHierarchyItemsAndHoldingsViewInstance(instanceIds, false, response -> assertThat(response.getStatusCode(), is(500)));
```
marcjohnson-kint, line 110: Can the assertion be pushed down into the method in order to shorten this line?

Is there anywhere that calls `requestInventoryHierarchyViewUpdatedInstanceIds` using a different matcher that checks something other than the status code?

K-Felk: > Can the assertion be pushed down into the method in order to shorten this line?

If I write my own version of the requestInventoryHierarchyViewUpdatedInstanceIds method, yes.  The way this one is written, you have to pass the assertion in as an argument.  I'll do so.

> Is there anywhere that calls `requestInventoryHierarchyViewUpdatedInstanceIds` using a different matcher that checks something other than the status code?

No, and I went back and forth a bit on wether I should add this-my concern here is that 500 errors are so generic-I'd like to be sure it's throwing one for the reason I expect.  I can remove if you think it's not necessary.

marcjohnson-kint: > If I write my own version of the requestInventoryHierarchyViewUpdatedInstanceIds method, yes. The way this one is written, you have to pass the assertion in as an argument. I'll do so.

Given this is the only use, I don't think there is a need to create another intermediary.

marcjohnson-kint: Please could the statement be split so that it isn't more than 100 characters per line.

marcjohnson-kint (approved):

PR #623, opened 2021-05-20
MODINVSTOR-663: Log failed domain events to internal table.

bohdan-suprun:
## Approach:

* Create new `notification_sending_error` table with following structure:
  * id;
  * topicName;
  * partitionKey;
  * payload;
  * error;
  * incidentDateTime;
* Introduce `FailureHandler` interface that injected into domain events publisher class
* Add LogToDbTable default implementation for `FailureHandler`.

### DIFF
--- a/src/main/java/org/folio/services/domainevent/LogToDbFailureHandler.java
+++ b/src/main/java/org/folio/services/domainevent/LogToDbFailureHandler.java
@@ -11,14 +11,14 @@ import java.util.UUID;
 import org.folio.persist.NotificationSendingErrorRepository;
 import org.folio.persist.entity.NotificationSendingError;
 
-public class LogToDbFailureHandler implements FailureHandler {
+final class LogToDbFailureHandler implements FailureHandler {
   private final NotificationSendingErrorRepository repository;
 
-  public LogToDbFailureHandler(NotificationSendingErrorRepository repository) {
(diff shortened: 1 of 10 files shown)


src/main/java/org/folio/services/domainevent/LogToDbFailureHandler.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Date;
+import java.util.Map;
+import java.util.UUID;
+import org.folio.persist.NotificationSendingErrorRepository;
+import org.folio.persist.entity.NotificationSendingError;
+
+public class LogToDbFailureHandler implements FailureHandler {
+  private final NotificationSendingErrorRepository repository;
+
+  public LogToDbFailureHandler(NotificationSendingErrorRepository repository) {
```
hjiebsco, line 17: Looks like `public` modifier can be omitted to reduce visibility.

hjiebsco (approved):
Assuming sys admin/operator needs to monitor the new table.

pfilippovEPAM (approved):
Looks good to me

src/main/java/org/folio/services/domainevent/LogToDbFailureHandler.java
bohdan-suprun: Changed, thanks

PR #586, opened 2021-02-25
Updates for RMB 33.0.0-SNAPSHOT

adamdickmeiss:

PR #624, opened 2021-05-20
MODINVSTOR-726, MODINVSTOR-713: Optimistic locking

julianladisch:
This configures `"withOptimisticLocking": "failOnConflict"` for instances, holdings and items: [MODINVSTOR-726](https://issues.folio.org/browse/MODINVSTOR-726), [MODINVSTOR-713](https://issues.folio.org/browse/MODINVSTOR-713)

RMB and Vert.x upgrades have been split off (#629 has upgraded RMB to 33.0.0 and Vert.x to 4.1.0.CR1 in master; these upgrade are no longer part of this PR.)

See [UXPROD-3089 "Inventory. Implementing Optimistic Locking"](https://issues.folio.org/browse/UXPROD-3089) for a list issues to co-ordinate the interface version upgrade across all affected modules.
See [FOLIO-3241 "Test inventory optimistic locking in scratch environment"](https://issues.folio.org/browse/FOLIO-3241) how we test this before merging to master.


adamdickmeiss:
I wished optimistic locking (with so many interface updates) should have been a separate commit.

And why waste time mentioning Beta.1 in news when that's going to be wrong anyway.

julianladisch:
Optimistic locking fails unless RMB is upgraded to 33.0.0, do you want a failing PR?

adamdickmeiss:
> Optimistic locking fails unless RMB is upgraded to 33.0.0, do you want a failing PR?

No. I just think it should have been two phase: update to RMB, then introduce optimistic locking. I'm not the maintainer of this module, not a big deal.

marcjohnson-kint:
> > Optimistic locking fails unless RMB is upgraded to 33.0.0, do you want a failing PR?
> 
> No. I just think it should have been two phase: update to nwer RMB, then introduce optimistic locking.

I agree, it would've been preferable for these to be separate pull requests

>  I'm not the maintainer of this module, not a big deal.

In FOLIO's process, you have more authority over this PR, as a member of the same team as Julian, than I do as the lead maintainer

descriptors/ModuleDescriptor-template.json
```diff
@@ -4,7 +4,7 @@
   "provides": [
     {
       "id": "item-storage",
-      "version": "8.9",
+      "version": "9.0",
```
marcjohnson-kint, line 7: Why is this change considered a breaking compatibility change?

Given that it is considered as such, please: 
* ensure that this work is not merged until pull requests can be coordinated for all dependent modules
* the RAML files are also updated so the versions correspond

pom.xml
```diff
@@ -59,6 +76,18 @@
...
+      <artifactId>assertj-core</artifactId>
+      <version>3.19.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <version>2.10.9</version>
+      <scope>test</scope>
+    </dependency>
```
marcjohnson-kint, line 90: As I understand it, joda time is no longer intended to be used, and instead the Java 8+ date time classes used.

Why has this dependency been introduced?

pom.xml
```diff
@@ -59,6 +76,18 @@
       <artifactId>marc4j</artifactId>
       <version>2.9.1</version>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
+      <version>3.19.0</version>
+      <scope>test</scope>
+    </dependency>
```
marcjohnson-kint, line 84: This project already uses Hamcrest for matchers, why has an alternative been introduced?

pom.xml
adamdickmeiss: It was necessary for this pr branch initially because it was used by RMB and this module used it. Perhaps not anymore as this branch is old by now

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -81,50 +83,85 @@ String servicePointUserFilter(String s) {
...
-            tl.withFilter(this::servicePointUserFilter)
-              .withPostOnly()
-              .withAcceptStatus(422)
-              .add("users", "service-points-users");
-          }
-          return tl.perform(attributes, headers, vertxContext, superRecordsLoaded);
-        }).compose(result -> runJavaMigrations(attributes, vertxContext, headers).map(result));
+        .compose(x ->super.loadData(attributes, tenantId, headers, vertxContext));
+
+    if (moduleFrom.isNew("20.0.0")) {
```
marcjohnson-kint, line 128: Why is this check for the version necessary in order to decide to load reference and sample data in addition to invoking `super.loadData`?

adamdickmeiss:
Ran this today and tests still fail:

```sh

Results :

Tests in error: 
  ItemShelvingOrderMigrationServiceApiTest.shouldStopProcessingWhenCannotConvertItem:41->MigrationTestBase.executeSql:81->TestBase.get:73 » Runtime

Tests run: 846, Failures: 0, Errors: 1, Skipped: 1

```

julianladisch:
I've extracted the RMB and Vert.x upgrade into a separate pull request #629 by continuing #586 and the updates-for-rmb-33-0-0 branch.
Those upgrade commits that have been included into this pull request to make it succeed will disappear from the "Files changed" section a soon as #629 is merged to master.

pom.xml
julianladisch: mod-inventory-storage has been using joda-time in tests since 2019:
https://github.com/folio-org/mod-inventory-storage/commit/2d0737522f0c1681411b0143d17558b3c13ff30e
https://github.com/folio-org/mod-inventory-storage/search?q=joda
Jodat-time usage is a technical depth of mod-inventory-storage: MODINVSTOR-733
Removing joda-time usage from mod-inventory-storage is out of scope of this pull request.
The explicit joda-time dependency in pom.xml is needed because RMB domain-models-runtime before 33.0.0 shipped with joda-time. Since 33.0.0 joda-time is no longer included.

marcjohnson-kint: @julianladisch Thank you for explaining why these dependencies were introduced. At no point did I suggest including the removal of joda time in the scope of this pull request.

(this comment was made redundant by the other pull request)

pom.xml
julianladisch: The https://github.com/folio-org/raml-module-builder/blob/master/doc/upgrading.md#version-330 says: "Add org.assertj test dependency to pom.xml if Assertj is used. RMB no longer ships with Assertj."

This pull request doesn't add any additional Assertj tests. The pom.xml entry is needed for the existing Assertj tests.

I've created https://issues.folio.org/browse/MODINVSTOR-734 "Replace Assertj matchers by Hamcrest matchers" = https://github.com/folio-org/mod-inventory-storage/pull/630 to fix this technical debt.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: TenantLoading doesn't support PUT for tables where optimistic locking is enabled.
The isNew check ensures that TenantLoading is called on the initial tenant install but not during any upgrade. I assume that people upgrade from Iris or later so that the previous version is always >= 20.0.0 on upgrades and this data load task is skipped.
If upgrading from Honeysuckle should be supported we need to add an upgrade task for https://github.com/folio-org/mod-inventory-storage/commit/a2b152596a93b5924887e0b8bbf52317eba5a69d to TenantRefAPI.java.

src/test/java/org/folio/rest/impl/TenantRefAPITest.java
```diff
@@ -0,0 +1,28 @@
...
+    "1.3.2 , 1.22.3, true",
+    "1.22.2, 1.22.3, true",
+    "1.22.3, 1.22.3, false",
+    "1.22.4, 1.22.3, false",
+  })
+  @Test
+  public void isNew(String migratingFromVersion, String featureVersion, boolean expected) {
+    if ("null".equals(migratingFromVersion)) {
+      migratingFromVersion = null;
+    }
```
bohdan-suprun, line 24: You can annotate `String migrationFromVersion` with `@Nullable` annotation

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -81,50 +83,85 @@ String servicePointUserFilter(String s) {
     return res;
   }
 
+  /**
+   * TODO: Move to okapi-common for re-use
+   */
+  static class ModuleFrom {
```
bohdan-suprun, line 89: Migration classes uses `Versioned` for similar logic, see `org.folio.services.migration.BaseMigrationService#versioned`. Should we use `versioned` here as well, or replace with `ModuleFrom` there?

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -611,7 +611,19 @@ public void tenantIsRequiredForGettingAllHoldings() throws InterruptedException,
...
+  @Test
+  public void optimisticLockingVersion() {
+    UUID holdingId = createInstanceAndHolding(mainLibraryLocationId);
+    JsonObject holding = getById(holdingId.toString()).getJson();
+    holding.put(PERMANENT_LOCATION_ID_KEY, annexLibraryLocationId);
+    // updating with current _version 1 succeeds and increments _version to 2
+    assertThat(update(holding).getStatusCode(), is(204));
+    holding.put(PERMANENT_LOCATION_ID_KEY, secondFloorLocationId);
+    // updating with outdated _version 1 fails, current _version is 2
+    assertThat(update(holding).getStatusCode(), is(409));
```
bohdan-suprun, line 623: Should the consumer increment it or it is autoincremented if difference is 1?

For example, I've read an item from database, that has `_version: 3`, than I updated item shelving order. Assuming no concurrent updates, should I increment `_version` to `4` or it is incremented by the locking mechanism? 

I'm asking, because there is a migration class `org.folio.services.migration.item.ItemShelvingOrderMigrationService` that does exactly what I described.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
marcjohnson-kint: > TenantLoading doesn't support PUT for tables where optimistic locking is enabled.

Could this have been changed to avoid all modules that use RAML Module Builder and will support optimistic concurrency control to make similar changes?

> The isNew check ensures that TenantLoading is called on the initial tenant install but not during any upgrade. I assume that people upgrade from Iris or later so that the previous version is always >= 20.0.0 on upgrades and this data load task is skipped.

You are assuming that no sample or reference data loading is needed, irrespective of the parameters folks use, if it isn't a first enablement of a module version in this family?

> If upgrading from Honeysuckle should be supported we need to add an upgrade task for [a2b1525](https://github.com/folio-org/mod-inventory-storage/commit/a2b152596a93b5924887e0b8bbf52317eba5a69d) to TenantRefAPI.java.

Have you asked folks if this is needed?

src/test/java/org/folio/rest/impl/TenantRefAPITest.java
julianladisch: Thanks, fixed.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
julianladisch: The consumer/client should never increment it. Only the database trigger created by RMB should increment it to ensure that the check and the increment are in the same transaction to avoid any race condition.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: Using PUT overwrites the reference/sample data, but this is not what people want because they have altered the reference data.
In most cases existing reference data should not been altered. This is accomplished by skipping data loading when the version check shows that the data has already been loaded.
Therefore the TenantLoading PUT doesn't support overwriting with optimistic locking yet.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: Thanks, I've replaced ModuleFrom class by Versioned class.

descriptors/ModuleDescriptor-template.json
julianladisch: It was decided that enabling "failOnConflict" is a breaking change with major interface bump: https://github.com/folio-org/raml-module-builder#api-and-schema-versioning-for-optimisticlocking
The corresponding mod-inventory changes are in https://issues.folio.org/browse/MODINV-398 = https://github.com/folio-org/mod-inventory/pull/344

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
bohdan-suprun: Sounds great, thanks!

bohdan-suprun:
No comments/questions from my side, leaving up to Core teams to approve.

adamdickmeiss (approved):

descriptors/ModuleDescriptor-template.json
julianladisch: All RAML files have corresponding versions.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
julianladisch: >> TenantLoading doesn't support PUT for tables where optimistic locking is enabled.

> Could this have been changed to avoid all modules that use RAML Module Builder and will support optimistic concurrency control to make similar changes?

No, because we don't know whether fields that have been changed since the inital load should
* be kept or
* be overwritten with the inital values.

Therefore upgrading has always been the task of the module, not the task of RMB's methods for initial data loading.

marcjohnson-kint: > This is accomplished by skipping data loading when the version check shows that the data has already been loaded.

How can a version check determine if the data has been loaded or not?

PR #627, opened 2021-05-25
Fix shouldStopProcessingWhenCa.. MODINVSTOR-725

adamdickmeiss:
Fix
ItemShelvingOrderMigrationServiceApiTest#shouldStopProcessingWhenCannotConvertItem
depends on other test (can not be run in isolation - when Vert.x
4.1.0.Beta1 is in use).

### DIFF
--- a/src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
+++ b/src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
@@ -37,7 +37,7 @@ public class ItemShelvingOrderMigrationServiceApiTest extends MigrationTestBase 
   @Test
   public void shouldStopProcessingWhenCannotConvertItem() throws Exception {
     var item = createItem(0);
-    executeSql("UPDATE item SET jsonb = '{\"a\":\"b\"}'::jsonb WHERE id = '" + item.getId() + "'");
+    executeSql("UPDATE " + getSchemaName() + ".item SET jsonb = jsonb || '{\"a\":\"b\"}'::jsonb WHERE id = '" + item.getId() + "'");
 
     var ta = getTenantAttributes();
 


src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
```diff
@@ -37,7 +37,7 @@ public void shouldPopulateShelvingOrder() throws Exception {
   @Test
   public void shouldStopProcessingWhenCannotConvertItem() throws Exception {
     var item = createItem(0);
-    executeSql("UPDATE item SET jsonb = '{\"a\":\"b\"}'::jsonb WHERE id = '" + item.getId() + "'");
+    executeSql("UPDATE " + getSchemaName() + ".item SET jsonb = jsonb || '{\"a\":\"b\"}'::jsonb WHERE id = '" + item.getId() + "'");
```
marcjohnson-kint, line 40: What has changed with vert.x 4.1.x meaning that the schema name now needs to be explicit (though, I think that's a good thing)?

adamdickmeiss: It was not the intention.

marcjohnson-kint: > It was not the intention.

It was not the intention of RMB moving to Vert.x 4.1.0 to require explicit schema in SQL?

marcjohnson-kint: > Exactly. That's why I don't like this. I also think explicit schema name would be better, but that's not for RMB 33. At this point, I think we should analyze this - perhaps just a problem with the test at hand.

Agreed on all points

PR #629, opened 2021-05-25
MODINVSTOR-728: Upgrade to RAML Module Builder 33.0.0 and Vert.x 4.1.0.CR1

julianladisch:


pom.xml
```diff
@@ -59,6 +76,18 @@
...
+      <artifactId>assertj-core</artifactId>
+      <version>3.19.0</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>joda-time</groupId>
+      <artifactId>joda-time</artifactId>
+      <version>2.10.9</version>
+      <scope>test</scope>
+    </dependency>
```
marcjohnson-kint, line 90: Are these needed because RAML Module Builder uses them?

marcjohnson-kint: Damn, that's unfortunate :-/

marcjohnson-kint (approved):

pom.xml
julianladisch: RMB before 33.0.0 shipped with (production) dependency on assertj-core and joda-time.
This has been fixed since RMB 33.0.0.
mod-inventory-storage uses both assertj-core and joda-time in tests. Therefore these test dependencies have been added.
I've created https://issues.folio.org/browse/MODINVSTOR-734 "Replace Assertj matchers by Hamcrest matchers" = #630 to fix the Assertj technical debt.
I've created https://issues.folio.org/browse/MODINVSTOR-733 "Replace joda-time by java.time" to fix the joda-time technical debt.

PR #631, opened 2021-05-26
MODINVSTOR-724: Populate publication period property for instance.

bohdan-suprun:
## Changes:
* Implement dateOfPublication parsing logic as per requirements;
* Populate the publicationPeriod on instance create/update if the dateOfPublication changed and the period is not set yet;
* Move stream processing logic for java migrations to base class.
* Add java migration for the new property.

### DIFF
--- a/src/main/java/org/folio/services/instance/InstanceEffectiveValuesService.java
+++ b/src/main/java/org/folio/services/instance/InstanceEffectiveValuesService.java
@@ -1,0 +1,74 @@
+package org.folio.services.instance;
+
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.toMap;
+import static org.folio.services.instance.PublicationPeriodParser.parsePublicationPeriod;
+
+import java.util.List;
+import java.util.Objects;
+import org.folio.rest.jaxrs.model.Instance;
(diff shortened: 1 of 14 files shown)


pfilippovEPAM (approved):

src/main/java/org/folio/services/instance/InstanceEffectiveValuesService.java
```diff
@@ -0,0 +1,74 @@
...
+      return true;
+    }
+
+    if (newPublications.size() != oldPublications.size()) {
+      return true;
+    }
+
+    if (newPublications.size() == 0) {
+      return false;
+    }
```
hjiebsco, line 59: Seems to me you can adjust the logic to return false if newPublications is null or size == 0.

hjiebsco (approved):
Left a minor comment for consideration. Other than that, looks fine to me.

src/main/java/org/folio/services/instance/InstanceEffectiveValuesService.java
bohdan-suprun: It is already returns false if newPublications is empty. If any is null than we still have to recalculate the property to remove the old value.

PR #637, opened 2021-06-30
MODINVSTOR-722: Remove update effective location database functions

K-Felk:
I wrote a script to remove the functions and put an entry to run it in the
schema file.  I didn't put this in the news file, it didn't seem significant
enough to warrant inclusion.

I had problems with some of the tests that looked at times failing 
because the test was a bit too exact. So I adjusted a few of the 
tests to have longer margins for error when comparing times.
Also found and removed a redundant library import in one of the 
test files.

### DIFF
--- a/src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
+++ b/src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
@@ -1,0 +1,2 @@
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_holding_update CASCADE;
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_item_update CASCADE;
--- a/src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
+++ b/src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
@@ -1,2 +1,2 @@
-DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_holding_update CASCADE;
-DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_item_update CASCADE;
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_holding_update;
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_item_update;
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -1067,6 +1067,11 @@
       "run": "after",
       "snippetPath": "addInstanceFormatsAudioBelt.sql",
       "fromModuleVersion": "20.3.0"
+    },
+    {
+      "run": "after",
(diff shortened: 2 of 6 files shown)


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1067,6 +1067,11 @@
       "run": "after",
       "snippetPath": "addInstanceFormatsAudioBelt.sql",
       "fromModuleVersion": "20.3.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "dropLegacyItemEffectiveLocationFunctions.sql",
+      "fromModuleVersion": "21.0.0"
```
marcjohnson-kint, line 1074: Version 21.0.0 has already been released. I think that means that this won't be executed when an upgrade is done.

As the next planned release is 21.1.0, I think that is what the from module version should be.

src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
```diff
@@ -0,0 +1,2 @@
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_holding_update CASCADE;
+DROP FUNCTION IF EXISTS ${myuniversity}_${mymodule}.update_effective_location_on_item_update CASCADE;
```
marcjohnson-kint, line 2: Why are these using `CASCADE`?

marcjohnson-kint (commented):
How did you verify that the script removes the functions?

If possible, please write a test to check this. I believe there is a base class for migration testing that could help with this.

src/main/resources/templates/db_scripts/schema.json
K-Felk: Fixed.

src/main/resources/templates/db_scripts/dropLegacyItemEffectiveLocationFunctions.sql
K-Felk: I found that other scripts in this module that were dropping database functions were doing it using cascade.  It seemed odd, but I wasn't sure if there was a reason I wasn't aware of.

marcjohnson-kint: I suggest we don't do that. 

I'd rather it fail if there were unexpected uses of these functions, rather than for those uses to be arbitrarily dropped as well.

marcjohnson-kint (approved):

src/test/java/org/folio/rest/api/LegacyItemEffectiveLocationMigrationScriptTest.java
```diff
@@ -0,0 +1,40 @@
...
+import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowSet;
+
+public class LegacyItemEffectiveLocationMigrationScriptTest extends MigrationTestBase {
+
+  private static final String CREATE_FUNCTIONS_SCRIPT = loadScript("effectiveLocationFunctionCreator.sql");
+  private static final String DROP_EFFECTIVE_LOCATION_FUNCTION_SCRIPT = loadScript("dropLegacyItemEffectiveLocationFunctions.sql");
+
+  @Test
+  public void canDropLegacyEffectiveItemLocationFunctions() throws Exception {
```
marcjohnson-kint, line 20: If lombok is part of the module, this could use `@SneakyThrows` (which preserves the specific types whilst also removing noise) and get rid of the `throws Exception`

PR #638, opened 2021-07-09
Include tenant and environment in topic names MODINVSTOR-738

marcjohnson-kint:
## Purpose

Publish all inventory messages to different topics, whose name starts with the environment, then the tenant ID e.g. `test.foo_tenant.inventory.instance`

None of the existing topics have been removed, as I wasn't sure about the requirements or the impact on pre-existing consumers.

This change is **implementation compatibility breaking** and needs to be deployed together with changes to consuming modules.

## Approach
Usually, I prefer to refactor the code to make the change easier before making a change. Given the urgency of this change, the need to back port it and my lack of familiarity with the code, I chose to try to make the least amount of changes.

There are changes to the tests in order to be able to change the topic names in them independently of the production code. I find this useful for checking that test fail in the expected way before making the production code change.

More information can be found in some of the commit messages.

I may issue a follow up pull request to this with further refactoring of this code, after these changes have been tested, back ported and released.

### DIFF
--- a/src/main/java/org/folio/services/kafka/topic/KafkaAdminClientService.java
+++ b/src/main/java/org/folio/services/kafka/topic/KafkaAdminClientService.java
@@ -3,6 +3,7 @@ package org.folio.services.kafka.topic;
 import static io.vertx.core.Future.succeededFuture;
 import static io.vertx.kafka.admin.KafkaAdminClient.create;
 import static org.apache.logging.log4j.LogManager.getLogger;
+import static org.folio.Environment.getEnvironmentName;
 import static org.folio.services.kafka.KafkaProperties.getReplicationFactor;
 
 import java.util.List;
@@ -57,8 +58,8 @@ public class KafkaAdminClientService {
 
   private Future<Void> createKafkaTopics(String tenantId, KafkaAdminClient kafkaAdminClient) {
     final List<NewTopic> expectedTopics = readTopics()
-      .map(topic -> prefixWithTenant(topic, tenantId))
-      .map(this::prefixWithEnvironment)
+      .map(topic -> prefixWith(tenantId, topic))
+      .map(topic -> prefixWith(getEnvironmentName(), topic))
       .collect(Collectors.toList());
 
     return kafkaAdminClient.listTopics().compose(existingTopics -> {
@@ -77,14 +78,8 @@ public class KafkaAdminClientService {
     });
   }
 
-  private NewTopic prefixWithTenant(NewTopic topic, String tenantId) {
-    return topic.setName(tenantId + "." + topic.getName());
-  }
-
-  private NewTopic prefixWithEnvironment(NewTopic topic) {
-    final var environmentName = System.getenv().getOrDefault("ENV", "folio");
-
-    return topic.setName(environmentName + "." + topic.getName());
+  private NewTopic prefixWith(String value, NewTopic topic) {
+    return topic.setName(value + "." + topic.getName());
   }
 
   private Stream<NewTopic> readTopics() {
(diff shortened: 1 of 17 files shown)


marcjohnson-kint:
@MikhailFokanov @hjiebsco @pfilippovEPAM 

Given my unfamiliarity with both the existing code and the requirements for this work, I'd appreciate a review of this to check the scope of my work and the changes I've made. Any feedback appreciated.

I likely still have some changes to make, however given the urgency of the changes I wanted to issue the pull request as soon as possible.

src/main/java/org/folio/services/kafka/topic/KafkaAdminClientService.java
```diff
@@ -37,30 +38,36 @@ public KafkaAdminClientService(Supplier<KafkaAdminClient> clientFactory) {
...
   }
 
-  private Future<Void> createKafkaTopics(KafkaAdminClient kafkaAdminClient) {
-    final List<NewTopic> newTopics = readTopics();
+  private Future<Void> createKafkaTopics(String tenantId, String environmentName,
+    KafkaAdminClient kafkaAdminClient) {
+
+    final List<NewTopic> expectedTopics = readTopics()
+      .map(topic -> prefixWith(tenantId, topic))
+      .map(topic -> prefixWith(environmentName, topic))
```
hjiebsco, line 64: Maybe do a string concat in one call?

src/main/java/org/folio/services/kafka/topic/KafkaTopic.java
```diff
@@ -1,11 +1,20 @@
...
+  }
 
-public enum KafkaTopic {
-  INVENTORY_INSTANCE("inventory.instance"),
-  INVENTORY_ITEM("inventory.item"),
-  INVENTORY_HOLDINGS_RECORD("inventory.holdings-record");
+  public static KafkaTopic item(String tenantId, String environmentName) {
+    return new KafkaTopic(prefixWith(environmentName,
+      prefixWith(tenantId, "inventory.item")));
+  }
```
hjiebsco, line 17: Maybe inline string concat will be sufficient.

src/test/java/org/folio/services/domainevent/CommonDomainEventPublisherTest.java
```diff
@@ -40,7 +43,7 @@
   @Before
   public void setUpPublisher() {
     eventPublisher = new CommonDomainEventPublisher<>(
-      Map.of(), INVENTORY_INSTANCE, producerManager, failureHandler);
+      Map.of(), KafkaTopic.instance("", environmentName()), producerManager, failureHandler);
```
hjiebsco, line 46: Topic without tenant id on purpose?

hjiebsco (approved):
Thanks for getting this done so quickly!

src/main/java/org/folio/services/kafka/topic/KafkaAdminClientService.java
marcjohnson-kint: I deliberately kept them separate to try to emphasise the separate aspects of the name qualification. 

I'll change this to a single string statement.

src/main/java/org/folio/services/kafka/topic/KafkaTopic.java
marcjohnson-kint: I deliberately kept them separate to try to emphasise the separate aspects of the name qualification. 

I'll change this to a single string statement.

src/test/java/org/folio/services/domainevent/CommonDomainEventPublisherTest.java
marcjohnson-kint: I will add a tenant here. 

It doesn't seem to make any difference to the test (which I'm a little surprised by given it's test for publishing, probably related to mocking).

marcjohnson-kint:
> Thanks for getting this done so quickly!

Thank you for the quick review. I'll try to get a review from within my team on Monday and get this merged.

K-Felk (approved):
I didn't see anything that seemed obviously problematic, but I also don't have a lot of experience working with Kafka.

PR #639, opened 2021-07-09
MODINVSTOR-719: Add indexTitle, alternativeTitles, series to keyword search index

K-Felk:
I made the modifications to the index outlined in the jira issue.

I then wrote a test for the new search behavior.

I believe this is supposed to be backported, so I added
The change under the last release in the NEWS file.
Once this is approved, I'll hotfix and backport.

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -577,6 +577,42 @@ public class InstanceStorageTest extends TestBaseWithInventoryUtil {
   }
 
   @Test
+  public void canSearchAlternateTitlesUsingKeywordIndexAll()
+    throws MalformedURLException,
+    InterruptedException,
+    ExecutionException,
+    TimeoutException {
+
+    UUID firstInstanceId = UUID.randomUUID();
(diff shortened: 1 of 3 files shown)


marcjohnson-kint:
> I believe this is supposed to be backported, so I added the change under the last release in the NEWS file.

It is intended to be back ported. However, as far as the mainline is concerned, this should be included in the news for the next unreleased minor version of the module.

As we cannot change existing releases, when this is back-ported, it will need to be as 21.0.1 (even though I don't consider this a bug).

Does that make sense?

K-Felk:
> Does that make sense?

I think so.  I'll amend the news.

marcjohnson-kint (approved):

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -576,6 +576,42 @@ public void canSearchUsingKeywordIndexAll()
     assertThat(allInstances.size(), is(1));
   }
 
+  @Test
+  public void canSearchAlternateTitlesUsingKeywordIndexAll()
```
marcjohnson-kint, line 580:
```suggestion
  public void canSearchForAlternateTitlesUsingKeywordIndexAll()
```

PR #643, opened 2021-07-19
Use environment and tenant in topic names MODINVSTOR-744

marcjohnson-kint:
## Purpose

Publish all inventory messages to different topics, whose name starts with the environment, then the tenant ID e.g. `test.foo_tenant.inventory.instance`

None of the existing topics have been removed, as I wasn't sure about the requirements or the impact on pre-existing consumers.

This change is **implementation compatibility breaking** and needs to be deployed together with changes to consuming modules.

## Approach
Usually, I prefer to refactor the code to make the change easier before making a change. Given the urgency of this change, that it is being made to an existing release and my lack of familiarity with the code, I chose to try to make the least amount of changes.

There are changes to the tests in order to be able to change the topic names in them independently of the production code. I find this useful for checking that test fail in the expected way before making the production code change.

The re-indexing test uses mocks in this version rather than publishing messages, so it isn't possible to add the same checks as in the mainline change.

More information can be found in some of the commit messages.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -2392,6 +2392,7 @@
       { "name": "JAVA_OPTIONS",
         "value": "-XX:MaxRAMPercentage=66.0"
       },
+      { "name": "ENV", "value": "folio" },
       { "name": "DB_HOST", "value": "postgres" },
       { "name": "DB_PORT", "value": "5432" },
       { "name": "DB_USERNAME", "value": "folio_admin" },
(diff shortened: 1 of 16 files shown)


marcjohnson-kint:
@MikhailFokanov @hjiebsco @pfilippovEPAM 

Given my unfamiliarity with both the existing code and the requirements for this work, I'd appreciate a review of this to check the scope of my work and the changes I've made. Any feedback appreciated.

K-Felk (approved):

hjiebsco (approved):
LGTM. Thanks for quick turnaround.

PR #647, opened 2021-07-30
MODINVSTOR-742 Add PUT title collection endpoint

psmagin:
## Purpose 
Add PUT title collection endpoint

## Approach
* extend `instance-preceding-succeeding-titles` API with new endpoint
* add tests

### DIFF
--- a/ramls/instance-preceding-succeeding-titles.raml
+++ b/ramls/instance-preceding-succeeding-titles.raml
@@ -1,6 +1,6 @@
 #%RAML 1.0
 title: Preceding/succeeding Titles API
-version: v0.1
+version: v0.2
 protocols: [ HTTP, HTTPS ]
 baseUri: http://localhost
 
@@ -67,6 +67,40 @@ resourceTypes:
           body:
             text/plain:
               example: "Internal server error, contact administrator"
+    put:
+      description: Update preceding/succeeding titles related to the instance
+      queryParameters:
+        instanceId:
+          type: string
+          required: true
+          pattern: "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
+      is: [validate]
+      body:
+        application/json:
+          type: precedingSucceedingTitles
+          example:
+            strict: false
+            value: !include examples/precedingsucceedingtitles_get.json
+      responses:
+        204:
+          description: "Successfully updated"
+        404:
+          description: "Instance with a given ID not found"
+          body:
+            text/plain:
+              example: |
+                "Instance with a given ID not found"
+        400:
+          description: "Bad request: malformed request body"
+          body:
+            text/plain:
+              example: |
+                "unable to update preceding/succeeding titles"
+        500:
+          description: "Internal server error, e.g. due to misconfiguration"
+          body:
+            text/plain:
+              example: "internal server error, contact administrator"
     /{precedingSucceedingTitleId}:
       type:
         collection-item:
--- a/ramls/instance-preceding-succeeding-titles.raml
+++ b/ramls/instance-preceding-succeeding-titles.raml
@@ -67,40 +67,6 @@ resourceTypes:
           body:
             text/plain:
               example: "Internal server error, contact administrator"
-    put:
-      description: Update preceding/succeeding titles related to the instance
-      queryParameters:
-        instanceId:
-          type: string
-          required: true
-          pattern: "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
-      is: [validate]
-      body:
-        application/json:
-          type: precedingSucceedingTitles
-          example:
-            strict: false
-            value: !include examples/precedingsucceedingtitles_get.json
-      responses:
-        204:
-          description: "Successfully updated"
-        404:
-          description: "Instance with a given ID not found"
-          body:
-            text/plain:
-              example: |
-                "Instance with a given ID not found"
-        400:
-          description: "Bad request: malformed request body"
-          body:
-            text/plain:
-              example: |
-                "unable to update preceding/succeeding titles"
-        500:
-          description: "Internal server error, e.g. due to misconfiguration"
-          body:
-            text/plain:
-              example: "internal server error, contact administrator"
     /{precedingSucceedingTitleId}:
       type:
         collection-item:
@@ -150,3 +116,33 @@ resourceTypes:
             body:
               text/plain:
                 example: "Internal server error, contact administrator"
+    /instances/{instanceId}:
+      put:
+        description: Update preceding/succeeding titles related to the instance
+        is: [validate]
+        body:
+          application/json:
+            type: precedingSucceedingTitles
+            example:
+              strict: false
+              value: !include examples/precedingsucceedingtitles_get.json
+        responses:
+          204:
+            description: "Successfully updated"
+          404:
+            description: "Instance with a given ID not found"
+            body:
+              text/plain:
+                example: |
+                  "Instance with a given ID not found"
+          400:
+            description: "Bad request: malformed request body"
+            body:
+              text/plain:
+                example: |
+                  "unable to update preceding/succeeding titles"
+          500:
+            description: "Internal server error, e.g. due to misconfiguration"
+            body:
+              text/plain:
+                example: "internal server error, contact administrator"
--- a/ramls/instance-storage.raml
+++ b/ramls/instance-storage.raml
@@ -14,11 +14,14 @@ types:
   marcJson: !include marc.json
   instanceRelationship: !include instancerelationship.json
   instanceRelationships: !include instancerelationships.json
+  precedingSucceedingTitles: !include instanceprecedingsucceedingtitles.json
+  errors: !include raml-util/schemas/errors.schema
 
 traits:
   language: !include raml-util/traits/language.raml
   pageable: !include raml-util/traits/pageable.raml
   searchable: !include raml-util/traits/searchable.raml
+  validate: !include raml-util/traits/validation.raml
 
 resourceTypes:
   collection: !include raml-util/rtypes/collection.raml
@@ -92,6 +95,36 @@ resourceTypes:
           Instances are stored and accessed by a hash of key properties. The rules which govern
           how instance hashes are computed are business rules and defined in the service layer.
           the storage layer only knows how to insert or retrieve instance records by ID.
+      /preceding-succeeding-titles:
+        put:
+          description: Update preceding/succeeding titles related to the instance
+          is: [validate]
+          body:
+            application/json:
+              type: precedingSucceedingTitles
+              example:
+                strict: false
+                value: !include examples/precedingsucceedingtitles_get.json
+          responses:
+            204:
+              description: "Successfully updated"
+            404:
+              description: "Instance with a given ID not found"
+              body:
+                text/plain:
+                  example: |
+                    "Instance with a given ID not found"
+            400:
+              description: "Bad request: malformed request body"
+              body:
+                text/plain:
+                  example: |
+                    "unable to update preceding/succeeding titles"
+            500:
+              description: "Internal server error, e.g. due to misconfiguration"
+              body:
+                text/plain:
+                  example: "internal server error, contact administrator"
       /source-record:
         delete:
           is: [language]
--- a/ramls/instance-storage.raml
+++ b/ramls/instance-storage.raml
@@ -14,14 +14,11 @@ types:
   marcJson: !include marc.json
   instanceRelationship: !include instancerelationship.json
   instanceRelationships: !include instancerelationships.json
-  precedingSucceedingTitles: !include instanceprecedingsucceedingtitles.json
-  errors: !include raml-util/schemas/errors.schema
 
 traits:
   language: !include raml-util/traits/language.raml
   pageable: !include raml-util/traits/pageable.raml
   searchable: !include raml-util/traits/searchable.raml
-  validate: !include raml-util/traits/validation.raml
 
 resourceTypes:
   collection: !include raml-util/rtypes/collection.raml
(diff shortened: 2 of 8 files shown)


ramls/instance-storage.raml
```diff
@@ -92,6 +95,36 @@ resourceTypes:
           Instances are stored and accessed by a hash of key properties. The rules which govern
           how instance hashes are computed are business rules and defined in the service layer.
           the storage layer only knows how to insert or retrieve instance records by ID.
+      /preceding-succeeding-titles:
```
marcjohnson-kint, line 98: Why has this endpoint been added to the instance-storage interface when there is a separate preceding / succeeding titles interface?

psmagin: Its because `/instance/<id>/preceding-succeeding-titles` endpoint has direct relation to the instance. It will be incorrect to add this endpoint to the preceding / succeeding titles interface because it operates more narrow resources.

marcjohnson-kint: > Its because `/instances/<id>/preceding-succeeding-titles` endpoint has direct relation to the instance. 

All [preceding / succeeding titles](https://github.com/folio-org/mod-inventory-storage/blob/16ecdd1b28d55e30ca8b4135eb84080d9adda755/ramls/instanceprecedingsucceedingtitle.json#L11) already have a direct relationship with instances

Could an endpoint be added to the `preceding / succeeding titles` interface that was also specific to an instance instead?

bq. It will be incorrect to add this endpoint to the preceding / succeeding titles interface because it operates more narrow resources.

What do you mean by narrow resources?

At the moment, it is possible to implement the `instance storage interface` without any recognition of preceding / succeeding titles. If this endpoint is added to the instance storage API this would no longer be true, and I think would undermine the intent of `preceding / succeeding titles` being a separate interface.

cc: @nielserik

psmagin: Hey @marc-johnson. I moved it to another interface.

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -68,6 +94,56 @@ public void putPrecedingSucceedingTitlesByPrecedingSucceedingTitleId(String prec
       PutPrecedingSucceedingTitlesByPrecedingSucceedingTitleIdResponse.class, asyncResultHandler);
   }
 
+  private boolean validatePrecedingSucceedingTitles(List<PrecedingSucceedingTitle> precedingSucceedingTitles,
```
oleksandr-dekin, line 97: Maybe it will be better to move this logic and  saveCollection into service layer?

oleksandr-dekin (approved):

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
psmagin: It's is a good point. But in this case, we should move all logic of this interface implementation to the service layer, not only one endpoint. I think there could be a separate story for this if it is needed.

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -68,6 +94,56 @@ public void putPrecedingSucceedingTitlesByPrecedingSucceedingTitleId(String prec
...
+                                                    Handler<AsyncResult<Response>> asyncResultHandler) {
+    boolean areValidTitles = true;
+    for (PrecedingSucceedingTitle precedingSucceedingTitle : precedingSucceedingTitles) {
+      if (titleIsLinkedToInstanceId(precedingSucceedingTitle, instanceId)) {
+        var validationErrorMessage =
+          createValidationErrorMessage("precedingInstanceId or succeedingInstanceId", "",
+            String.format("The precedingInstanceId or succeedingInstanceId should contain instanceId [%s]", instanceId));
+        asyncResultHandler.handle(Future.succeededFuture(PutPrecedingSucceedingTitlesResponse
+          .respond422WithApplicationJson(validationErrorMessage)));
+        areValidTitles = false;
```
julianladisch, line 108: If there are multiple invalid titles the asyncResultHandler is called multiple times resulting in an error because HTTP can return a single 422 message only.
Consider replacing `areValidTitles = false;` by `return false;` and `return areValidTitles;` by `return true;`

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -68,6 +94,56 @@ public void putPrecedingSucceedingTitlesByPrecedingSucceedingTitleId(String prec
...
+      }
+    }
+    return areValidTitles;
+  }
+
+  private boolean titleIsLinkedToInstanceId(PrecedingSucceedingTitle precedingSucceedingTitle, String instanceId) {
+    return !instanceId.equals(precedingSucceedingTitle.getPrecedingInstanceId())
+      && !instanceId.equals(precedingSucceedingTitle.getSucceedingInstanceId());
```
julianladisch, line 116: The more negations code has the more difficult to understand it is.
I suggest 
```
    return instanceId.equals(precedingSucceedingTitle.getPrecedingInstanceId())
      || instanceId.equals(precedingSucceedingTitle.getSucceedingInstanceId());
```
and using
```
if (! titleIsLinkedToInstanceId(precedingSucceedingTitle, instanceId))
```

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -39,6 +50,21 @@ public void postPrecedingSucceedingTitles(String lang, PrecedingSucceedingTitle
       PostPrecedingSucceedingTitlesResponse.class, asyncResultHandler);
   }
 
+  @Override
+  public void putPrecedingSucceedingTitles(String instanceId, PrecedingSucceedingTitles entity,
```
julianladisch, line 54: `@Validate`

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
```diff
@@ -68,6 +94,56 @@ public void putPrecedingSucceedingTitlesByPrecedingSucceedingTitleId(String prec
...
+          promise.complete(response);
+        } else {
+          promise.complete(PutPrecedingSucceedingTitlesResponse.respond204());
+        }
+      };
+      postgresClient.saveBatch(PRECEDING_SUCCEEDING_TITLE_TABLE, entities, replyHandler);
+    } catch (ReflectiveOperationException e) {
+      promise.complete(PutPrecedingSucceedingTitlesResponse.respond500WithTextPlain(e.getMessage()));
+    }
+    return promise.future();
```
julianladisch, line 144: [Futurisation](https://github.com/folio-org/raml-module-builder/blob/master/doc/futurisation.md) allows to greatly reduce the boilerplate code:

```suggestion
  private Future<Response> saveCollection(List<PrecedingSucceedingTitle> entities,
                                          Map<String, String> okapiHeaders, Context vertxContext) {
    try {
      MetadataUtil.populateMetadata(entities, okapiHeaders);
      return postgresClient(vertxContext, okapiHeaders)
          .saveBatch(PRECEDING_SUCCEEDING_TITLE_TABLE, entities)
          .<Response>map(x -> PutPrecedingSucceedingTitlesResponse.respond204())
          .otherwise(cause -> {
            var errorMessage = cause.getMessage();
            if (ValidationHelper.isFKViolation(errorMessage)) {
              return PutPrecedingSucceedingTitlesResponse.respond404WithTextPlain("Instance not found");
            } else {
              return PutPrecedingSucceedingTitlesResponse.respond500WithTextPlain(errorMessage);
            }
          });
    } catch (Throwable e) {
      return Future.succeededFuture(PutPrecedingSucceedingTitlesResponse.respond500WithTextPlain(e.getMessage()));
    }
  }
```

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
psmagin: Great catch. Thanks

src/main/java/org/folio/rest/impl/PrecedingSucceedingTitleAPI.java
psmagin: This is really better. Thanks.

ramls/instance-preceding-succeeding-titles.raml
```diff
@@ -67,6 +67,40 @@ resourceTypes:
           body:
             text/plain:
               example: "Internal server error, contact administrator"
+    put:
+      description: Update preceding/succeeding titles related to the instance
+      queryParameters:
+        instanceId:
+          type: string
+          required: true
+          pattern: "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
```
marcjohnson-kint, line 76: Why was a query parameter chosen instead of including the instance in the path?

ramls/instance-storage.raml
marcjohnson-kint: @psmagin 

> I moved it to another interface.

Thank you

ramls/instance-preceding-succeeding-titles.raml
psmagin: That's because in the case of adding it to `instance-preceding-succeeding-titles` interface it will violate the REST principle to go from general to more specific in the url.

marcjohnson-kint: > That's because in the case of adding it to `instance-preceding-succeeding-titles` interface it will violate the REST principle to go from general to more specific in the url.

I don't think I'm aware of that principal, where can I learn more about it?

How does using a query parameter avoid this challenge?

psmagin: https://restful-api-design.readthedocs.io/en/latest/urls.html
Titles are more specific then instance. So instance should come first and titles after. But it's not possible in current interface. That's why instance id is in query params

marcjohnson-kint: > https://restful-api-design.readthedocs.io/en/latest/urls.html

Thanks for the link, I wasn't aware of this guidance at all.

I had a quick read of that document and couldn't find anything about going from general to specific ideas in a URL. Which section of the document talks about this that I missed?

> Titles are more specific then instance. So instance should come first and titles after. 

What do you mean by `more specific`? Do you mean that a preceding / succeeding title is beneath an instance in a hierarchy?

Does that mean that `/instances/{id}/preceding-succeeding-titles` would be ok however `/preceding-succeeding-titles/instances/{id}/` is not?

And so `preceding-succeeding-titles?instanceId={id}` was chosen instead to try to remove the hierarchy from the path?

> But it's not possible in current interface. That's why instance id is in query params

Why is it not possible? Is it because Okapi stops the use of instance in the path in this interface because it's used in another interface?

psmagin: > I had a quick read of that document and couldn't find anything about going from general to specific ideas in a URL. Which section of the document talks about this that I missed?

There is about collection and sub-collection hierarchy:
image

> What do you mean by more specific? Do you mean that a preceding / succeeding title is beneath an instance in a hierarchy?

Yes, you understood me right. Preceding / succeeding title couldn't exist without instance so it is beneath an instance in a hierarchy.

> Does that mean that /instances/{id}/preceding-succeeding-titles would be ok however /preceding-succeeding-titles/instances/{id}/ is not?

Yes.

> Why is it not possible? Is it because Okapi stops the use of instance in the path in this interface because it's used in another interface?

It's not possible because we can't use `/preceding-succeeding-titles/instances/{id}` because of collection/sub-collection principle. And we can't use smth like `/instances/{id}/preceding-succeeding-titles` in this interface because all paths in the interface should be consistent (in our case starts with `/preceding-succeeding-titles`

marcjohnson-kint: Thank you for sharing your reasoning with me.

> > I had a quick read of that document and couldn't find anything about going from general to specific ideas in a URL. Which section of the document talks about this that I missed?
> 
> There is about collection and sub-collection hierarchy:
> image

Thank you for referring to specifically to that. I think I interpreted that less strongly than you did.

> > What do you mean by more specific? Do you mean that a preceding / succeeding title is beneath an instance in a hierarchy?
> 
> Yes, you understood me right. Preceding / succeeding title couldn't exist without instance so it is beneath an instance in a hierarchy.

Ok, 

> > Does that mean that /instances/{id}/preceding-succeeding-titles would be ok however /preceding-succeeding-titles/instances/{id}/ is not?
> 
> Yes.
> 
> > Why is it not possible? Is it because Okapi stops the use of instance in the path in this interface because it's used in another interface?
> 
> It's not possible because we can't use `/preceding-succeeding-titles/instances/{id}` because of collection/sub-collection principle. 

Like the above, I think I didn't interpret that guidance that sub-collections _must_ be nested under a non-collection resource.

bq. And we can't use smth like `/instances/{id}/preceding-succeeding-titles` in this interface because all paths in the interface should be consistent (in our case starts with `/preceding-succeeding-titles`

Why do all of the paths need to be consistent and start with the same path?

psmagin: The pros of REST approach is that developer could expect uniform interface from API. If paths are not consistent then it becomes chaos. 

So, why are you against the approach to have instance id as a query parameter? Could you provide some cons of it?

marcjohnson-kint: > The pros of REST approach is that developer could expect uniform interface from API. If paths are not consistent then it becomes chaos.

My understanding is the uniform interface principle refers more to how a client interacts with resources, rather than the structure of specific URLs.

> So, why are you against the approach to have instance id as a query parameter? Could you provide some cons of it?

I don't think I've argued against the use of a query parameter. I wanted to understand the reasoning behind the decision.

I am biased for paths. I find paths easier to comprehend and easier to construct than query parameters. And that query parameters can easily be missed or ignored.

I found the idea of a [clean URL](https://en.wikipedia.org/wiki/Clean_URL) as a good explanation of my bias (that admittedly, I developed prior to knowing about that concept).

psmagin: This query parameter is required so it couldn't be missed or ignored.

But I heard you. You are the owner of this module, so I will agree with any your decision. What URL do you prefer to have for this endpoint?

marcjohnson-kint: > This query parameter is required so it couldn't be missed or ignored.

Fair enough.

> But I heard you. You are the owner of this module, so I will agree with any your decision. What URL do you prefer to have for this endpoint?

I'm not trying to make an argument from authority (though it seems I have). I wanted to understand the reasoning. 

Neither am I'm not attempting to block the (already approved) work or request changes. I don't think there is a need to change the code again at this point. I will resolve this comment.

As for what I would do, there are some challenges with how the preceding / succeeding titles are already structured. Given that, my preference would be for a path something like `preceding-succeeding-titles/instances/{id}`.
If there had been a prefix for this interface, I might have gone with `preceding-succeeding-titles-storage/instances/{id}/preceding-succeeding-title`.

Each of these have challenges with long paths and are somewhat clunky. If FOLIO didn't have the business logic split and we were starting from scratch, I'd probably use something like `instances/{id}/preceding-titles` and `instances/{id}/succeeding-titles` to represent these two collections (though that is somewhat misleading as these resources aren't really independent, due to how preceding / succeeding titles can be associated with more than one instance.

psmagin: It's not a big deal to change it. I have done this.

natali-zaitseva (approved):

PR #648, opened 2021-07-31
MODINVSTOR-523: Item barcode should be unique

julianladisch:
Add unique index for item barcode.

Remove all other indexes for barcode: non-unique b-tree index, ginIndex, fullTextIndex.
"tOps": "DELETE" is not needed since RMB v30.0.0: https://issues.folio.org/browse/RMB-552
Using "tOps": "DELETE" while changing index type is broken: https://issues.folio.org/browse/RMB-864

Unit tests:
* Can create multiple items without barcode.
* Cannot create multiple items with the same barcode (respects accents, ignores case).

Fix existing unit tests that use multiple items with the same barcode.

TODO:

"Changes and Required Actions" of Kiwi Release Notes
https://wiki.folio.org/display/REL/R3+2021+(Kiwi)+Release+Notes
have a note about upgrading mod-inventory-storage when
there are duplicate item barcodes.

Update that note as needed.


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -788,6 +788,12 @@
 
       ],
       "uniqueIndex": [
+        {
+          "fieldName": "barcode",
+          "tOps": "ADD",
+          "caseSensitive": false,
+          "removeAccents": false
```
marcjohnson-kint, line 795: The values for these parameters are different to the existing indexes, is that intentional? 

My understanding was that database indexes won't be used unless these parameters match up with how the CQL is interpreted, which would suggest that either the old indexes weren't being used or the new one won't. What have I missed / misunderstood?

src/test/java/org/folio/rest/api/InventoryViewTest.java
```diff
@@ -108,6 +108,13 @@ public void shouldReturnInstanceEvenIfNoHoldings() {
     }
   }
 
+  /**
+   * nod without barcode to allow multiple items
+   */
+  static JsonObject nod(UUID holdingsRecordId) {
```
marcjohnson-kint, line 114: Given the intention is this item should not have a barcode, I suggest this be described in the method name rather than a comment (meaning a consumer does not need to check the comments / implementation).

```suggestion
  static JsonObject nodWithNoBarcode(UUID holdingsRecordId) {
```

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
```diff
@@ -288,6 +287,15 @@ private IndividualResource createHoldingsWithPropertySetAndInstance(
...
   }
 
+  /**
+   * ItemStorageTest.nod with random id and without barcode
+   */
+  private JsonObject nod(UUID holdingsRecordId) {
+    JsonObject nod = ItemStorageTest.nod(holdingsRecordId);
+    nod.remove("barcode");
+    return nod;
+  }
```
marcjohnson-kint, line 297: Given the intention is this item should not have a barcode, I suggest this be described in the method name rather than a comment (meaning a consumer does not need to check the comments / implementation).

```suggestion
  static JsonObject nodWithNoBarcode(UUID holdingsRecordId) {
```

src/test/java/org/folio/rest/api/ItemEffectiveCallNumberComponentsTest.java
```diff
@@ -288,6 +287,15 @@ private IndividualResource createHoldingsWithPropertySetAndInstance(
     return holdings;
   }
 
+  /**
+   * ItemStorageTest.nod with random id and without barcode
+   */
+  private JsonObject nod(UUID holdingsRecordId) {
```
marcjohnson-kint, line 293: Given the intention is this item should not have a barcode, I suggest this be described in the method name rather than a comment (meaning a consumer does not need to check the comments / implementation).

```suggestion
  private JsonObject nodWithNoBarcode(UUID holdingsRecordId) {
```

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
```diff
@@ -38,7 +38,7 @@ public void shouldPopulateShelvingOrder() throws Exception {
 
   @Test
   public void shouldStopProcessingWhenCannotConvertItem() throws Exception {
-    var item = createItem(0);
+    var item = createItem(300);
```
marcjohnson-kint, line 41: Rather than pass in an index, could this be changed to make it take the whole barcode. I think that would make the intent more explicit.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1005,6 +1003,8 @@ public void cannotSyncPostWithDuplicateId() {
   public Response postSynchronousBatchWithExistingId(String subPath) {
     JsonArray itemsArray1 = threeItems();
     JsonArray itemsArray2 = threeItems();
+    itemsArray2.getJsonObject(0).put("barcode", "new0");
+    itemsArray2.getJsonObject(2).put("barcode", "new2");
```
marcjohnson-kint, line 1007: Are these details important to the scenario this test is intending to cover? 

If not, could barcode generation be included in the `threeItems` method instead of the main body of the test.

src/main/resources/templates/db_scripts/schema.json
julianladisch: This is intentional, see discussion on https://issues.folio.org/browse/MODINVSTOR-523

RMB automatically uses the case and accents settings for both creating the indexes and when running a CQL query. This ensures that the indexes are always been used. Please suggest a better wording for the documentation if it is unclear: https://github.com/folio-org/raml-module-builder#cql-relations

marcjohnson-kint: > This is intentional, see discussion on https://issues.folio.org/browse/MODINVSTOR-523

Thank you for clarifying that and helping me change my out of date understanding.
 
> RMB automatically uses the case and accents settings for both creating the indexes and when running a CQL query. This ensures that the indexes are always been used. Please suggest a better wording for the documentation if it is unclear: https://github.com/folio-org/raml-module-builder#cql-relations

I was not aware of that

src/test/java/org/folio/rest/api/ItemShelvingOrderMigrationServiceApiTest.java
julianladisch: The passed index is not only used for the barcode but also for the call number components.

marcjohnson-kint: Ok, those could be split, however it's likely out of scope of this work

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2427,26 +2423,48 @@ static JsonObject nod(UUID holdingsRecordId) {
...
     CompletableFuture<Response> searchCompleted = new CompletableFuture<>();
 
     client.get(itemsStorageUrl("?query=") + urlEncode(searchQuery),
       StorageTestSuite.TENANT_ID, ResponseHandler.json(searchCompleted));
 
-    return searchCompleted.get(5, TimeUnit.SECONDS).getJson()
-      .mapTo(Items.class);
+    Response response = searchCompleted.get(5, TimeUnit.SECONDS);
+    assertThat(response.getStatusCode(), is(200));
+    return response.getJson().mapTo(Items.class);
```
marcjohnson-kint, line 2467: I find it easier to notice assertions when they are separated from the statements around them

```suggestion
    Response response = searchCompleted.get(5, TimeUnit.SECONDS);
    
    assertThat(response.getStatusCode(), is(200));
    
    return response.getJson().mapTo(Items.class);
```

src/test/java/org/folio/rest/api/ItemStorageTest.java
julianladisch: threeItems() has always returned the same barcodes, now it returns random barcodes to prevent a duplicate barcode error.

marcjohnson-kint (commented):
Please add an entry into the news that describes the change in behaviour

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1573,78 +1573,74 @@ public void canPageAllItems()
...
-    TimeoutException {
+  /**
+   * Assert that the cql query returns items with the expected barcodes.
+   */
+  private void assertCqlFindsBarcodes(String cql, String ... expectedBarcodes) throws Exception {
+    Items items = findItems(cql);
+    String [] barcodes = items.getItems().stream().map(Item::getBarcode).toArray(String []::new);
+    assertThat(cql, barcodes, is(expectedBarcodes));
+    assertThat(cql, items.getTotalRecords(), is(barcodes.length));
+  }
```
marcjohnson-kint, line 1584: Given that these helper methods are intended to aid the readability of the tests, please move them to the bottom of the test file so they do not distract from the tests themselves.

marcjohnson-kint:
Why have the interface versions been changed for this new constraint, as it is not part of the interface definition?

src/test/java/org/folio/rest/api/ItemStorageTest.java
julianladisch: Moved.

julianladisch:
> Why have the interface versions been changed for this new constraint, as it is not part of the interface definition?

item.json says:
```
    "barcode": {
      "type": "string",
      "description": "Unique inventory control number for physical resources, used largely for circulation purposes"
    },
```
Therefore uniqueness is part of the interface.

Or do you consider this a bug fix because mod-inventory-storage should have been rejecting duplicate barcodes from the very beginning?

marcjohnson-kint:
> > Why have the interface versions been changed for this new constraint, as it is not part of the interface definition?
> 
> item.json says:
> 
> ```
>     "barcode": {
>       "type": "string",
>       "description": "Unique inventory control number for physical resources, used largely for circulation purposes"
>     },
> ```
> 
> Therefore uniqueness is part of the interface.

I disagree that a description is sufficient for something to be part of the interface rather than a policy.

I think that is a quirk of how we use descriptions in interface definitions (and not implementations). I don't think that constraint is actually manifest in the interface itself.

Given that we disagree, I leave it to the team making the change to make the decision (Core Platform in this case). 

In practice, I imagine that you will be applying the same thinking to this as to the [statistical code changes](https://github.com/folio-org/mod-inventory-storage/pull/650#discussion_r691073443) and be clumping multiple (independent) changes under the banner of a single compatibility break. Which would rather make this conversation academic.

> Or do you consider this a bug fix because mod-inventory-storage should have been rejecting duplicate barcodes from the very beginning?

I can understand the perspective behind that framing.

I think that a side effect of splitting business logic and storage into separate modules leads to an ambiguity as to what belongs in each module.

My understanding was that the storage modules were intended to be actively ignorant of business logic, as part of the motivation for the separation was that they could be replaced independently. It was with this reasoning, that a policy (that I consider business logic).

Since then however, I think practical concerns have taken over that, and we have increasing found business logic in the storage modules for pragmatic purposes e.g. some constraints should be unavoidable, given that clients can go direct to storage.

julianladisch:
I've tested migration from mod-inventory-storage-21.0.3 with sample data to mod-inventory-storage-22.0.0-SNAPSHOT (= this PR).
Indexes before:

```
    "item_barcode_idx" btree ("left"(lower(f_unaccent(jsonb ->> 'barcode'::text)), 600))
    "item_barcode_idx_ft" gin (get_tsvector(f_unaccent(jsonb ->> 'barcode'::text)))
    "item_barcode_idx_gin" gin (lower(f_unaccent(jsonb ->> 'barcode'::text)) public.gin_trgm_ops)
```
Upgrades succeeds. Indexes after:
```
    "item_barcode_idx_unique" UNIQUE, btree (lower(jsonb ->> 'barcode'::text))
```

This is correct and as expected.

steveellis (approved):

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -860,10 +854,6 @@
         }
       ],
       "fullTextIndex": [
-        {
```
mtraneis, line 863: @julianladisch This index needs to be maintained because when creating a new request, the query 
`SELECT jsonb,id FROM <tenantID>_mod_inventory_storage.item WHERE get_tsvector(f_unaccent(item.jsonb->>?)) @@ tsquery_phrase(f_unaccent(?)) LIMIT ? OFFSET ?`
is used. Without this index creating requests takes minutes in a large DB.

julianladisch: See UIREQ-639 "Use == not = for item barcode search": https://issues.folio.org/browse/UIREQ-639

PR #649, opened 2021-08-02
MODINVSTOR-754: Sporadic await timeout in DomainEventAssertions

julianladisch:
Builds on Jenkins have sporadic failures:

https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fmod-inventory-storage/detail/PR-648/1/pipeline/

canPostSynchronousBatchWithExistingIdUpsertTrue(org.folio.rest.api.ItemStorageTest) Time elapsed: 1.101 sec <<< ERROR!

org.awaitility.core.ConditionTimeoutException: Lambda expression in org.folio.rest.support.matchers.DomainEventAssertions that uses java.lang.String, java.lang.Stringjava.lang.String: expected a value greater than <0> but <0> was equal to <0> within 1 seconds.

 at org.folio.rest.api.ItemStorageTest.canPostSynchronousBatchWithExistingIdUpsertTrue(ItemStorageTest.java:1051)

https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fmod-inventory-storage/detail/PR-648/2/pipeline/

canDeleteAnItem(org.folio.rest.api.ItemStorageTest) Time elapsed: 1.074 sec <<< ERROR!

org.awaitility.core.ConditionTimeoutException: Lambda expression in org.folio.rest.support.matchers.DomainEventAssertions that uses java.lang.String, java.lang.Stringjava.lang.String: expected a value greater than <1> but <1> was equal to <1> within 1 seconds.

 at org.folio.rest.api.ItemStorageTest.canDeleteAnItem(ItemStorageTest.java:1530)

https://jenkins-aws.indexdata.com/blue/organizations/jenkins/folio-org%2Fmod-inventory-storage/detail/folijet-rancher/18/pipeline/

canCreateAnItemViaCollectionResource(org.folio.rest.api.ItemStorageTest) Time elapsed: 1.034 sec <<< ERROR!

org.awaitility.core.ConditionTimeoutException: Lambda expression in org.folio.rest.support.matchers.DomainEventAssertions that uses java.lang.String, java.lang.Stringjava.lang.String: expected a value greater than <0> but <0> was equal to <0> within 1 seconds.

 at org.folio.rest.api.ItemStorageTest.canCreateAnItemViaCollectionResource(ItemStorageTest.java:291)

The timeout of 1 second might be too small, therefore setting the timeout is moved into a function that sets it to 2 seconds.


marcjohnson-kint:
> The timeout of 1 second might be too small, therefore it is removed to use the default await() timeout of 10 seconds.

I would prefer that this wasn't changed back to the default of 10 seconds as that means that failures take 10 times longer to resolve. Given 1 second seems to be insufficient, I would prefer we either to figure out why that is the case or increase it by a smaller increment e.g. 2 seconds to find a more acceptable value.

I also prefer specifying the timeout as that makes the policy explicit and does not rely on folks knowing what the default is.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
```diff
@@ -90,42 +91,38 @@ private static void assertHeaders(List<KafkaHeader> headers) {
...
     final JsonObject updateMessage  = getLastInstanceEvent(instanceId).value();
     assertThat(updateMessage.getString("type"), not(is("DELETE")));
   }
 
-  public static void assertNoCreateEvent(String instanceId) {
-    await().atLeast(1, SECONDS);
-
-    assertThat(getInstanceEvents(instanceId).size(), is(0));
+  public static void assertNoEvent(String instanceId) {
+    await().during(1, SECONDS)
```
marcjohnson-kint, line 118: According to the Awaitility documentation, [during](https://github.com/awaitility/awaitility/wiki/Usage#assert-that-a-value-is-maintained) checks that a value is maintained for a period of time.

My interpretation of that documentation is that this change will wait for 10 seconds and during that 10 seconds, there must be no events for 1 seconds worth of that time. Is that what you intended?

julianladisch: No, the Awaitility documentation explicitly says that it "will wait for at most" the 10 seconds.
"at most" means that it may stop earlier.
And it actually stops earlier when the conditions is true for 1 second because that assert cannot fail when waiting longer.
I've tested it and it actually exits after 1 second.

julianladisch:
I've created a method that explicitly sets the timeout to 2 seconds.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: Ok. Let me check my understanding.

This await relies on the Awaitility _default_ of waiting for at least 10 seconds, however it will override that when the state is as expected for a whole second during that (up to) 10 second wait?

marcjohnson-kint: Ok. Let me check my understanding.

This await relies on the waiting at most 2 seconds (as configured at the top of the file) however it will override that when the state is as expected for a whole second during that (up to) 10 second wait?

julianladisch: No.

The await() timeout is 10 seconds by default. However, this doesn't apply to DomainEventAssertions class because I've changed it to 2 seconds as you requested.

Therefore the await() timeout is always 2 seconds in DomainEventAssertions.

`await().until(someCondition);` will succeed if the condition is or becomes true before the await() timeout, otherwise it fails. A single point of time where condition is true is sufficient. The await() timeout is 2 seconds.

`await().during(1, SECOND).until(someCondition);` will succeed if the condition is true for at least 1 second before the await() timeout, otherwise it fails. A single point of time where condition is true is not sufficient, multiple tests that cover a timespan of 1 second are needed. The await() timeout is 2 seconds.
.5 seconds true and 1.5 seconds false fails.
.5 seconds false and 1 second true succeeds.
1.5 seconds false and 1 second true fails at the time when the timeout stops the execution (2 seconds from start) because the condition was true for only .5 seconds.

marcjohnson-kint: Thank you for the thorough explanation.

> The await() timeout is 10 seconds by default. However, this doesn't apply to DomainEventAssertions class because I've changed it to 2 seconds as you requested.
> 
> Therefore the await() timeout is always 2 seconds in DomainEventAssertions.

Ok, I think this fits with what I said.

> `await().until(someCondition);` will succeed if the condition is or becomes true before the await() timeout, otherwise it fails. A single point of time where condition is true is sufficient. The await() timeout is 2 seconds.

Sure, `until` is eager which tends to make it inappropriate for checking for the absence of conditions.

> `await().during(1, SECOND).until(someCondition);` will succeed if the condition is true for at least 1 second before the await() timeout, otherwise it fails. A single point of time where condition is true is not sufficient, multiple tests that cover a timespan of 1 second are needed. The await() timeout is 2 seconds.
> .5 seconds true and 1.5 seconds false fails.
> .5 seconds false and 1 second true succeeds.

Why does this succeed when the condition cannot have been satisfied for a whole second? At most, it can have been satisfied for 0.5 a second (1 - 0.5).

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
```diff
@@ -90,42 +101,38 @@ private static void assertHeaders(List<KafkaHeader> headers) {
...
     final JsonObject updateMessage  = getLastHoldingEvent(instanceId, hrId).value();
     assertThat(updateMessage.getString("type"), not(is("UPDATE")));
   }
 
   public static void assertNoRemoveEvent(String instanceId) {
-    await().atLeast(1, SECONDS);
-    await().atMost(1, SECONDS)
-      .until(() -> getInstanceEvents(instanceId).size(), greaterThan(0));
+    await()
+      .until(() -> getInstanceEvents(instanceId), is(not(empty())));
```
marcjohnson-kint, line 121: > await().until(someCondition); will succeed if the condition is or becomes true before the await() timeout, otherwise it fails. A single point of time where condition is true is sufficient. The await() timeout is 2 seconds.

I think this eagerness makes `until` (without a minimum wait) to be inappropriate for checking for the absence of something, as it will falsely succeed immediately.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
julianladisch: If someCondition holds true for .5 seconds and then someCondition holds false for 1.5 seconds then the check will fail.
If someCondition holds false for .5 seconds and then someCondition holds true for 1 second then the check will succeed because the condition has been satisfied for a while second. This succeeds 1.5 seconds after start (.5 + 1.5 seconds), no need to wait longer.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
julianladisch: This test doesn't check the absence of a delete event, but the presence of an event, and that that event is not a delete event.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: > If someCondition holds false for .5 seconds and then someCondition holds true for 1 second then the check will succeed because the condition has been satisfied for a while second.

I don't understand. 

If the observation after 0.5 second is that the condition _failed_ and the observation after 1 second is that the condition _succeeded_ the longest period of time where the condition could have continuously succeeded is 0.5 (1 - 0.5, because the earliest the condition could shift is immediately after the 0.5 second observation) a second which isn't sufficient for the check to succeed.

marcjohnson-kint: > This succeeds 1.5 seconds after start (.5 + 1.5 seconds), no need to wait longer.

This can only be relevant if the observation at 1.5 seconds succeeded. The previous statement make no mention of the observation at 1.5 seconds.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
marcjohnson-kint: Hmm, I don't think the logic of this assertion makes sense (and that might be my mistake as I may have changed it previously).

The first event not being of type `DELETE` does not mean that no message of type `DELETE` has been published, which is what the name `assertNoRemoveEvent`.

If this is beyond the scope of this pull request, that is ok, it isn't caused by your change.

marcjohnson-kint (approved):
Approving to get this merged to validate if it improves build stability.

Concerns with whether the checks demonstrate the intended situation are separate and need to be addressed separately.

src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
julianladisch: `await().until(someCondition);` succeeds if someCondition is true for a single point in time.

`await().during(1, SECOND).until(someCondition);` succeeds if someCondition is true for a time span, the time span specified in the `during` method. Think of a stopwatch that is started whenever someCondition becomes true and is stopped whenever someCondition becomes false. The await succeeds if the stopwatch reaches the timespan specified in the `during` method before timeout.

marcjohnson-kint: > `await().during(1, SECOND).until(someCondition);` succeeds if someCondition is true for a time span, the time span specified in the `during` method. Think of a stopwatch that is started whenever someCondition becomes true and is stopped whenever someCondition becomes false. The await succeeds if the stopwatch reaches the timespan specified in the `during` method before timeout.

I agree with that understanding.

I think that statement means this statement is not the case:

> If someCondition holds false for .5 seconds and then someCondition holds true for 1 second then the check will succeed because the condition has been satisfied for a while second.

As that stopwatch for the condition holding true can only have been going for 0.5 seconds (because it was false at 0.5 seconds and the latest time interval is at 1 second).

Anyway. Thank you for taking the time to patiently explain this to me. I have learnt more about the awaitility library.

PR #650, opened 2021-08-02
MODINVSTOR-755: Reject non-UUID statistical code IDs

julianladisch:
[MODINVSTOR-676](https://issues.folio.org/browse/MODINVSTOR-676) rejects non-existing statistical code IDs. This includes statistical code IDs that are not a UUID.

This issue is about a better error message.

Current response:

```
HTTP/1.1 500 Internal Server Error
transfer-encoding: chunked
Content-Type: text/plain

ERROR: invalid input syntax for type uuid: "07" (22P02)
```

Response with this change:
```
HTTP/1.1 422 Unprocessable Entity
transfer-encoding: chunked
Content-Type: application/json

{"errors":[
  { "message":"elements in list must match pattern",
    "type":"1",
    "code":"elements in list must match pattern",
    "parameters":[ { "key":"statisticalCodeIds","value":"[07]" } ]
  }
]}
```

422 status code errors already exist in the affected interfaces. Therefore this is not a breaking change but a bug fix to be released with mod-inventory-storage 24.0.0.


ramls/item.json
```diff
@@ -377,7 +377,8 @@
       "type": "array",
       "description": "List of statistical code IDs",
       "items": {
-        "type": "string"
+        "type": "string",
+        "pattern": "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[1-5][a-fA-F0-9]{3}-[89abAB][a-fA-F0-9]{3}-[a-fA-F0-9]{12}$"
```
marcjohnson-kint, line 381: This is a compatibility breaking change to the interfaces that use this schema. Please change their major versions accordingly and update the dependents.

julianladisch: Changed.

The same new interface version is used for enabling optimistic locking and the dependents get updated for optimistic locking by https://issues.folio.org/browse/UXPROD-3089 and https://issues.folio.org/browse/FOLIO-3241 . Therefore no additional work is needed to update dependents regarding this change (MODINVSTOR-755: Reject non-UUID statistical code IDs).

marcjohnson-kint: Ok, whilst I think it doesn't match how compatibility breaking changes are typically made, I understand that this is a pragmatic decision to avoid the impact of breaking compatibility twice for a change that is likely to have limited impact in practice.

K-Felk (approved):

PR #653, opened 2021-08-05
Remove sample and reference data scripts (MODINVSTOR-758)

nielserik:
Also: remove broken link to `doc` directory.


marcjohnson-kint (approved):
Thanks @nielserik

PR #656, opened 2021-08-10
MODINVSTOR-761: Return 400, not 500, for service-points invalid CQL

julianladisch:
Replace code by PgUtil.get.

The code has been written before PgUtil.get.


marcjohnson-kint (approved):
@julianladisch  Thank you for making this change

PR #660, opened 2021-09-03
Release 21.0.4

julianladisch:
This backports "Enforce item barcode uniqueness" ([MODINVSTOR-523](https://issues.folio.org/browse/MODINVSTOR-523)) = #648 from master to b21.0 for Juniper (R2 2021) Hot Fix # 2

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 21.0.3 2021-08-16
 
 * Includes missing changes to `get_items_and_holdings_view` function during upgrade (MODINVSTOR-759)
+* Enforce item barcode uniqueness (MODINVSTOR-523)
 
 ## 21.0.2 2021-07-20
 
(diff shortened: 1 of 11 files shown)


adamdickmeiss (commented):
Enforcing restrictions doesn't sound like a bug fix to me. But I guess everyone thinks overwise. What happens when upgrading to this version and there is non-uniqueness already??

MikhailFokanov-ebsco (approved):
Looks good.

julianladisch:
> Enforcing restrictions doesn't sound like a bug fix to me. But I guess everyone thinks otherwise. What happens when upgrading to this version and there is non-uniqueness already??

This has been discussed in https://issues.folio.org/browse/MODINVSTOR-523 and in the original pull request #648 this pull requests back-ports from master.

In short: Uniqueness has been enforced by mod-inventory. Only circumventing mod-inventory by directly accessing mod-inventory-storage could have created non-unique barcodes. Release notes are in place for Kiwi and will be provided for Juniper HotFix 2.

PR #661, opened 2021-09-07
MODINVSTOR-771 Fix issue with destination service point for updating item 

roman-barannyk:
In transit report does not show destination service point if item is edited.

## Purpose
- save  InTransitDestinationServicePointId field for item updating;
- create tests;

Resolves: [MODINVSTOR-771](https://issues.folio.org/browse/MODINVSTOR-771)

### DIFF
--- a/src/main/java/org/folio/services/item/ItemService.java
+++ b/src/main/java/org/folio/services/item/ItemService.java
@@ -36,6 +36,7 @@ import org.apache.logging.log4j.Logger;
 import org.folio.persist.ItemRepository;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
 import org.folio.rest.jaxrs.model.Item;
+import org.folio.rest.jaxrs.model.Status;
 import org.folio.rest.persist.PostgresClient;
 import org.folio.rest.persist.SQLConnection;
 import org.folio.rest.support.HridManager;
@@ -111,7 +112,7 @@ public class ItemService {
       .compose(oldItem -> refuseWhenHridChanged(oldItem, newItem))
       .compose(oldItem -> {
         var oldDestServicePointId = oldItem.getInTransitDestinationServicePointId();
-        if (oldDestServicePointId != null &&
+        if (newItem.getStatus().getName() == Status.Name.IN_TRANSIT &&
           newItem.getInTransitDestinationServicePointId() == null) {
 
           newItem.setInTransitDestinationServicePointId(oldDestServicePointId);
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -109,6 +109,16 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
+      .compose(oldItem -> {
+        var oldDestServicePointId = oldItem.getInTransitDestinationServicePointId();
+        if (oldDestServicePointId != null &&
+          newItem.getInTransitDestinationServicePointId() == null) {
+
+          newItem.setInTransitDestinationServicePointId(oldDestServicePointId);
+        }
+
+        return succeededFuture(oldItem);
+      })
```
marcjohnson-kint, line 121: This suggests that the in transit destination service point should be retained even if it is removed by the client making a change.

Why is this needed?

Does this make it impossible for a client to remove the destination service point?

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -109,6 +110,16 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
       .compose(oldItem -> refuseWhenHridChanged(oldItem, newItem))
+      .compose(oldItem -> {
+        var oldDestServicePointId = oldItem.getInTransitDestinationServicePointId();
+        if (newItem.getStatus().getName() == Status.Name.IN_TRANSIT &&
+          newItem.getInTransitDestinationServicePointId() == null) {
+
+          newItem.setInTransitDestinationServicePointId(oldDestServicePointId);
+        }
+
+        return succeededFuture(oldItem);
```
marcjohnson-kint, line 121: Why are we trying to preserve this on the server rather than relying on the client to provide a representation that includes it?

src/main/java/org/folio/services/item/ItemService.java
roman-barannyk: Thanks @marcjohnson-kint, you are right.
When an item is becoming "Available" mod-circulation on check-in sends InTransitDestinationServicePointId == null.
I've updated the condition and now the fix should work correctly for such case.

Another solution can be implemented on UI side: check InTransitDestinationServicePointId of old item and if nothing changes, send this field in new item.

marcjohnson-kint: > When an item is becoming "Available" mod-circulation on check-in sends InTransitDestinationServicePointId == null.
> I've updated the condition and now the fix should work correctly for such case.

Has it been decided by a PO that if an item is In-Transit it should *always* retain the destination service point irrespective of whether a client wants to remove it?

> Another solution can be implemented on UI side: check InTransitDestinationServicePointId of old item and if nothing changes, send this field in new item.

Personally, unless this is a deliberate rule, I think this would be better resolved in the client, which is the UI.

I think this kind of logic can easily lead to folks being surprised when they think they've removed a property from a record representation and instead, as if by magic, it comes back.

roman-barannyk: Thanks @marcjohnson-kint, we discussed it with FE team, it will be fixed on UI side. I'll close this PR

PR #662, opened 2021-09-07
MODINVSTOR-769:  Introduce new item API to include related records

K-Felk:
I named the endpoint "item-storage-dereferenced" to differentiate from the regular
item-storage endpoint.  

The principle for making this work turned out to be surprisingly simple: when a CQL
query is translated to SQL using the CQL2PgJSON utility, you can call its "toString" method,
which will dump a textual representation of the cql search translated to an sql WHERE clause.  
Then I just tacked that on to a monster SQL query with all of the fields and joins needed.  
That search will return a  collection of item records with their related instance, holdings, 
and other records.  

That's then run through a mapping function to restructure the data and map it to the 
correct model class.  I tried to keep the structure of the item records returned by this 
class very similar to the ones returned by the item-storage endpoint, except relevant Id 
numbers for related records are replaced with the full representation of the record.  
The case does not specifically ask for the endpoint to provide a way to retrieve item records
by ID number, but it was trivial to implement, I thought it might be useful, and it makes this 
endpoint behave more like the original item-storage endpoint.

Currently, the endpoint returns full records for every location and loan type, even temporary ones 
if present.  I reasoned as I  would already have to do joins on those tables, more likely would not matter.  
If that's not correct, I can just remove those sections of the query and leave the IDs in place.


K-Felk:
I have no idea what's up with the code coverage report-most of the code it's telling me is not covered actually have specific tests written for it in the test file, which are passing. ????

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
```diff
@@ -0,0 +1,198 @@
...
+  public void getItemStorageDereferencedItems(
+    int offset, int limit, String query, String lang,
+    Map<String, String> okapiHeaders,
+    Handler<AsyncResult<Response>> asyncResultHandler,
+    Context vertxContext) {
+
+    List<DereferencedItem> mappedResults = new ArrayList<>();
+    String whereClause = "";
+
+    if (query != null) {
```
marcjohnson-kint, line 92: For this particular endpoint, maybe we want to require the CQL query and not allow the entire result set to be fetched by default.

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
```diff
@@ -0,0 +1,198 @@
...
+    } else {
+      whereClause = "LIMIT " + limit + " OFFSET " + offset;
+    }
+
+    PostgresClient postgresClient = PgUtil.postgresClient(vertxContext, okapiHeaders);
+    postgresClient.select(sqlQuery + whereClause, asyncResult -> {
+
+      handleSelectFailure(asyncResult, asyncResultHandler);
+
+      asyncResult.result().forEach(row -> {mappedResults.add(mapToDereferencedItem(row));});
```
marcjohnson-kint, line 110: If the SQL query fails, won't this line get executed anyway (as there is no guard provided by `handleSelectFailure`) and cause a null pointer exception?

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
```diff
@@ -0,0 +1,198 @@
...
+    AsyncResult<RowSet<Row>> asyncResult,
+    Handler<AsyncResult<Response>> asyncResultHandler) {
+    
+    if (asyncResult.failed()) {
+      String errorMessage = asyncResult.cause().getMessage();
+      respondWith500Error("Can't retrive item records: " + errorMessage, asyncResultHandler);
+    }
+    if (asyncResult.result().size() == 0) {
+      respondWith404Error("No item records found.", asyncResultHandler);
+    }
```
marcjohnson-kint, line 152: Why is getting no record back a failure?

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
```diff
@@ -0,0 +1,198 @@
...
+        itemJson.remove(fieldName);
+      }
+    });   
+    List<String> fieldList = new ArrayList<String>(fields.values());
+    fieldList.forEach(fieldName -> {
+      int pos = row.getColumnIndex(fieldName.toLowerCase());
+      if (row.getJson(pos) != null) {
+        itemJson.put(fieldName, row.getJson(pos));
+      }
+    });
```
marcjohnson-kint, line 193: What is the purpose of this mapping?

If the item representation in a dereferenced item is the same as a regular item, why not copy the JSON completely?

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  private DereferencedItem mapToDereferencedItem(Row row) {
+    
+    JsonObject itemJson = (JsonObject) row.getJson("item");
+    idFields.forEach(fieldName -> {
+      if (itemJson.containsKey(fieldName)) {
+        itemJson.remove(fieldName);
+      }
+    });
```
marcjohnson-kint, line 186: Why do the ID fields need to be removed?

ramls/dereferenceditem.json
```diff
@@ -0,0 +1,407 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "An item record with related record data",
+  "type": "object",
+  "properties": {
+    "id": {
+      "type": "string",
+      "description": "Unique ID of the item record"
+    },
+    "instanceRecord":{
```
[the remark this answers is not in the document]
marcjohnson-kint: Having said all that, the instance and the holdings record aren't the only records that are de-referenced.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,245 @@
...
+    UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
+
+    JsonObject smallAngryPlanet = smallAngryPlanet(smallAngryPlanetId, holdingsRecordId);
+    JsonObject nod = nod(UUID.randomUUID(), holdingsRecordId);
+    JsonObject uprooted = uprooted(UUID.randomUUID(), holdingsRecordId);
+
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+    getCompleted = client.post(itemsStorageUrl(""), smallAngryPlanet, StorageTestSuite.TENANT_ID);
+    Response response = getCompleted.get(5, SECONDS);
+    assertThat(response.getStatusCode(), is(201));
```
marcjohnson-kint, line 56: This could be extracted into a method to remove the duplication for each record being created.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,245 @@
...
+    StorageTestSuite.deleteAll(holdingsStorageUrl(""));
+    StorageTestSuite.deleteAll(instancesStorageUrl(""));
+
+    UUID holdingsRecordId = createInstanceAndHolding(mainLibraryLocationId);
+
+    JsonObject smallAngryPlanet = smallAngryPlanet(smallAngryPlanetId, holdingsRecordId);
+    JsonObject nod = nod(UUID.randomUUID(), holdingsRecordId);
+    JsonObject uprooted = uprooted(UUID.randomUUID(), holdingsRecordId);
+
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
```
marcjohnson-kint, line 53: The request isn't a get, so I think this might be inappropriately named.
```suggestion
    CompletableFuture<Response> postCompleted = new CompletableFuture<>();
```

```suggestion
    CompletableFuture<Response> createCompleted = new CompletableFuture<>();
```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,245 @@
...
+    assertThat(items.getTotalRecords(), is(1));
+
+    DereferencedItem item = items.getDereferencedItems().get(0);
+
+    assertThat(item.getBarcode(), is("036000291452"));
+    assertThat(item.getId(), is(smallAngryPlanetId.toString()));
+    assertThat(item.getInstanceRecord().getTitle(), is("Long Way to a Small Angry Planet"));
+    assertThat(item.getPermanentLoanType().getName(), is("Can Circulate"));
+    assertThat(item.getMaterialType().getName(), is("journal"));
+    assertThat(item.getHoldingsRecord().getInstanceId(), is(item.getInstanceRecord().getId()));
```
marcjohnson-kint, line 87: What about the other dereferenced records?

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,245 @@
...
+  private DereferencedItems getAll() {
+    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
+    client.get(dereferencedItemStorage(""), TENANT_ID, json(getCompleted));
+    try {
+      return getCompleted.get(5, SECONDS).getJson()
+      .mapTo(DereferencedItems.class);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RuntimeException(e);
+    }
+  }
```
marcjohnson-kint, line 243: I suggest using sneaky throwing instead of this

(and to provide whitespace between method definitions)

```suggestion

  @SneakyThrows
  private DereferencedItems getAll() {
    CompletableFuture<Response> getCompleted = new CompletableFuture<>();
    client.get(dereferencedItemStorage(""), TENANT_ID, json(getCompleted));

    return getCompleted.get(5, SECONDS).getJson()
      .mapTo(DereferencedItems.class);
  }
```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,245 @@
...
+      String barcode) {
+
+    return createItemRequest(id, holdingsRecordId, barcode, journalMaterialTypeID);
+  }
+
+  private static JsonObject createItemRequest(
+    UUID id,
+    UUID holdingsRecordId,
+    String barcode,
+    String materialType) {
```
marcjohnson-kint, line 152: Why is every parameter on a separate line?

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,19 @@
   "id": "${artifactId}-${version}",
   "name": "Inventory Storage Module",
   "provides": [
+    {
+      "id": "item-storage-dereferenced",
+      "version": "0.1",
+      "handlers": [{
+        "methods": ["GET"],
+        "pathPattern": "/item-storage-dereferenced/items",
+        "permissionsRequired": ["inventory-storage.items.collection.get"]
```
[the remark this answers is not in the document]
K-Felk: Yes.  It made sense to me that if you have permission to get item data through the existing item-storage endpoint, then you should also have permission to get data through this API.  I couldn't think of a good reason to have a separate set of permissions-if you can, please tell me.

descriptors/ModuleDescriptor-template.json
```diff
@@ -2,6 +2,19 @@
...
+      },{
+        "methods": ["GET"],
+        "pathPattern": "/item-storage-dereferenced/items/{id}",
+        "permissionsRequired": ["inventory-storage.items.item.get"]
+      }]
+    },
```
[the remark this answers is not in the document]
K-Felk: See general comments above.  The existing item storage endpoint has the capacity for retrieving item records by Id-is there a good reason this one shouldn't?  It was trivial to implement once I had the code for CQL searching worked out.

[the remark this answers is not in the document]
marcjohnson-kint: > The existing item storage endpoint has the capacity for retrieving item records by Id-is there a good reason this one shouldn't? It was trivial to implement once I had the code for CQL searching worked out.

Ok. I tend to want to minimise the API surface area. I'm ok with leaving this in. I do think it has slightly complicated the implementation code (because the mapping and other behaviour is shared).

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
marcjohnson-kint: > I couldn't think of a good reason to have a separate set of permissions-if you can, please tell me.

We have an (admittedly informal) policy that each endpoint should have it's own permission.

It's to allow folks the flexibility of granting access independently, even if that doesn't make sense.

In this case, there might be organisations who want folks to have access to item information but not holdings or instance information, which is a discrepancy between these APIs.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
marcjohnson-kint: Thank you, that formatting style was likely something I did a long time ago and was a mistake.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
marcjohnson-kint: @K-Felk This has been resolved without the code changing, does that mean it's an un-pushed commit or a rejected suggestion?

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
K-Felk: The only one required not tested here is location-I added an assertion for that.

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
K-Felk: Okay, that makes sense.  I'll create separate permissions.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
marcjohnson-kint: What about the optional ones?

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
marcjohnson-kint: Thanks

K-Felk:
> I have no idea what's up with the code coverage report-most of the code it's telling me is not covered actually have specific tests written for it in the test file, which are passing. ????

Figured this out-I left off the "@test" annotation from some of the tests.  Fixed in this update, and the code coverage is now passing.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
K-Felk: If we're going to test for that, then we should also test for the absence of fields that should be null.  Added tests for both.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,270 @@
...
+    if (includeOptionalFields) {
+      itemToCreate.put("temporaryLoanTypeId", nonCirculatingLoanTypeID);
+      itemToCreate.put("temporaryLocationId", secondFloorLocationId);
+    }
+
+    return itemToCreate;
+  }
+
+
+
```
marcjohnson-kint, line 207:
```suggestion

```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,270 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+
+import io.vertx.core.json.JsonObject;
+import junitparams.JUnitParamsRunner;
+import lombok.SneakyThrows;
+
+
+@RunWith(JUnitParamsRunner.class)
```
marcjohnson-kint, line 36: There doesn't seem to be any parameterised tests, could this be removed?

```suggestion
```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
marcjohnson-kint: Thanks

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
marcjohnson-kint: I think we should be consistent with the other APIs and return 404 for the single record endpoint and 200 with zero records for the collection endpoint.

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
K-Felk: Well, it will still default limit to ten records.  I dunno, I tend to want endpoints to conform to established design patterns-and from what I can see, the standard design pattern for these endpoints is "If you don't give me a CQL query, I'll give you everything, starting with the first ten records."

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
K-Felk: I had assumed calling a response also terminated execution?  I haven't seen an example of an explicit break statement after those are called in the existing code-but they also tend to be used in a syntactically different way than I'm calling them here.  If not, then I'll add some kind of break statement to the error-handling function.

ramls/dereferenceditem.json
```diff
@@ -0,0 +1,407 @@
...
+            "type": "object",
+            "folio:$ref": "itemnotetype.json",
+            "javaType": "org.folio.rest.jaxrs.model.itemNoteTypeVirtual",
+            "readonly": true,
+            "folio:isVirtual": true,
+            "folio:linkBase": "item-note-types",
+            "folio:linkFromField": "itemNoteTypeId",
+            "folio:linkToField": "id",
+            "folio:includedElement": "itemNoteTypes.0"
+          },
```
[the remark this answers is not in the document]
K-Felk: Removed.

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
K-Felk: Apologies, I was under the impression that the standard response for "I didn't find anything" was a 404.  I'll fix.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,266 @@
...
+  @AfterClass
+  public static void cleanUpDatabase() {
+    StorageTestSuite.deleteAll(itemsStorageUrl(""));
+    StorageTestSuite.deleteAll(holdingsStorageUrl(""));
+    StorageTestSuite.deleteAll(instancesStorageUrl(""));
+  }
+
+  
+  public void CanGetRecordByCQLSearch() {
+    String queryString = "barcode=036000291452";
```
marcjohnson-kint, line 67: If this is intended to demonstrate an exact match on barcode (which I believe is the use case), the relation should be `==`

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,266 @@
...
+    String queryStringOptionalFields = "657670342075";
```
marcjohnson-kint, line 68: What is the intent of using a CQL query that does not include an index (e.g. `barcode`) or relation (e.g. `=`)?

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,266 @@
...
+    return createItemRequest(id, holdingsRecordId, barcode, journalMaterialTypeID, includeOptionalFields);
+  }
+
+  private static JsonObject createItemRequest(
+    UUID id, UUID holdingsRecordId, String barcode, 
+    String materialType, Boolean includeOptionalFields) {
+
+    JsonObject itemToCreate = new JsonObject();
+
+    if(id != null) {
```
marcjohnson-kint, line 185:
```suggestion
    if (id != null) {
```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,266 @@
...
+    assertThat(items.getTotalRecords(), is(3));
+  }
+
+  @Test
+  public void Returns404WhenNoItemsFound() {
+    String queryString = "barcode=647671342075";
+    
+    Response response = attemptFindByCql(queryString);
+
+    assertThat(response.getStatusCode(), is(404));
```
marcjohnson-kint, line 100: I thought this was going to be changed to return 200 and an empty record set?

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
```diff
@@ -0,0 +1,266 @@
...
+    String queryString = "barcode&647671342075";
+    
+    Response response = attemptFindByCql(queryString);
+
+    assertThat(response.getStatusCode(), is(400));
+  }
+
+  @Test
+  public void CanGetRecordById() {
+
```
marcjohnson-kint, line 114:
```suggestion
```

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
K-Felk: I don't follow-this is an assignment, I'm putting the query string for the search into a variable.

K-Felk: Fixed.

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
marcjohnson-kint: Ok, let's leave it as is.

src/main/java/org/folio/rest/impl/ItemStorageDereferencedAPI.java
marcjohnson-kint: > I had assumed calling a response also terminated execution?

It does not. Vert.x sending the response to the client has no effect on what further code the server executes.

This happens frequently enough that Vert.x has an error message specifically for circumstances where code attempts to write a second response after the first.

bq. I haven't seen an example of an explicit break statement after those are called in the existing code-but they also tend to be used in a syntactically different way than I'm calling them here. If not, then I'll add some kind of break statement to the error-handling function.

If the sending of the failure response is intended to be the final action, there needs to be some form of return / break to stop further execution.

src/test/java/org/folio/rest/api/DereferencedItemStorageTest.java
K-Felk: I think part of the query string got accidentally removed while I was doing other work.  I've fixed it.

descriptors/ModuleDescriptor-template.json
```diff
@@ -2468,7 +2493,7 @@
       { "name": "DB_QUERYTIMEOUT", "value": "60000" },
       { "name": "DB_CHARSET", "value": "UTF-8" },
       { "name": "DB_MAXPOOLSIZE", "value": "5" },
-      { "name": "KAFKA_HOST", "value": "kafka" },
+      { "name": "KAFKA_HOST", "value": "localhost" },
```
marcjohnson-kint, line 2496: Why was this value changed?

marcjohnson-kint (approved):

descriptors/ModuleDescriptor-template.json
K-Felk: This is my fault-I have a separate instance of this module I've been trying to get to work with okapi-I must have changed  the descriptor on this one by mistake.  Reverting.

PR #665, opened 2021-09-15
Remove code owners MODINVSTOR-775

marcjohnson-kint:
As the module is maintained by multiple teams the review requests generated by the code owners definition leads to unnecessary notifications that slow down the review process as folks start to ignore them

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-* @folio-org/inventory-backend-code-owners


K-Felk:
This may be a dumb question, but will this change the review process when someone on our team makes changes to this module?  Will we need to manually request reviews from those on our team?  It's fine if it does, I just want to know.

K-Felk (approved):

PR #666, opened 2021-09-16
MODINVSTOR-778: Explain local deployment with java -jar

julianladisch:


README.MD
```diff
@@ -93,6 +93,55 @@ Navigate to the folder `infrastructure/local` directory within the clone of the
...
+## Build jar
+
+    mvn clean package -DskipTests
+
+## Set environment variables
+
+    export DB_DATABASE=folio
+    export DB_PASSWORD=folio
+    export DB_USERNAME=folio
+    export KAFKA_PORT=29092
```
marcjohnson-kint, line 115: Why is any of this needed? 

It seems to directly contradict the instructions provided above for running a module locally using Docker.

marcjohnson-kint (commented):
Can you help me understand the intent of this pull request?

There are already instructions (directly above the additions) and a docker compose file (that avoids the need to duplicate infrastructure set up and set environment variables in a shell) for doing this.

marcjohnson-kint (changes requested):
@julianladisch 

I think this work is unnecessary and confusing as the module already contains a docker compose file for running the module locally. There are also instructions provided (directly above the ones added).

The approach already taken does not require environment variables to be set in the shell and does not require duplicated definitions of the infrastructure in each module repository. At the small trade off that folks need to run two separate docker compose commands (one for infrastructure and one for the module). This approach also makes it easier if folks want to run multiple modules locally, as less modifications to the docker compose files are needed to do so.

I think this PR should be either closed without merge. If the existing documentation is not satisfactory, a separate PR could be raised to improve that documentation e.g. if we think folks don't know how to build a local jar artefact.

README.MD
julianladisch: I've added an introduction that explains the differences between the two instructions.

Both have their con: Additional docker build step versus environment variables in the shell.

I've added the env variables to my .bashrc to avoid the additional docker build step.

If you don't like the "java -jar without container" instructions being mentioned in mod-inventory-storage repository I can put them at some other place like dev.folio.org or a new subdirectory of https://github.com/folio-org/folio-tools/tree/master/infrastructure .

marcjohnson-kint: Why does your new approach require a new docker compose file at all? 

There is already a docker compose file that can be used to set up infrastructure that can be used with a module deployed either with docker or as a jar. 

Why replicate that in every module repository?

README.MD
```diff
@@ -93,6 +97,60 @@ Navigate to the folder `infrastructure/local` directory within the clone of the
...
+For a deployment that runs a Docker container that wraps mod-inventory-storage-fat.jar
+see "Local Deployment using Docker" above.
+
+## Start PostgreSQL, Kafka and Zookeeper
+
+This starts them as Docker containers.
+
+    cd run-locally
+    docker-compose up -d
+    cd ..
```
marcjohnson-kint, line 113: Why can this not use the existing, shared, docker compose file for spinning up the infrastructure needed for the module?

run-locally/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+  kafka:
+    image: wurstmeister/kafka
+    container_name: kafka
+    restart: always
+    ports:
+      - "9092:9092"
+      - "29092:29092"
+    environment:
+      KAFKA_LISTENERS: INTERNAL://:9092,LOCAL://:29092
+      KAFKA_ADVERTISED_LISTENERS: INTERNAL://:9092,LOCAL://localhost:29092
```
julianladisch, line 26: @marc-johnson 
The entry `LOCAL://localhost:29092` in KAFKA_ADVERTISED_LISTENERS is needed to make the jar (running without being wrapped in a Docker container) to work.

Can I replace `LOCAL://:29092` by `LOCAL://localhost:29092` in KAFKA_ADVERTISED_LISTENERS in https://github.com/folio-org/folio-tools/blob/master/infrastructure/local/docker-compose.yml to avoid yet another docker-compose.yml?

Currently the port 29092 duplicates 9092 in infrastructure/local/docker-compose.yml and therefore has no use.

README.MD
marcjohnson-kint: > If you don't like the "java -jar without container" instructions being mentioned in mod-inventory-storage repository

I'm more than ok with having instructions for how to run the module as a jar locally.

I think it would be better if those instructions:
* used what we already have e.g. use the docker compose file for spinning up infrastructure rather than defining a new one
* are added alongside a review of the existing documentation we have for running the module locally (which uses Okapi and some scripts and may well be out of date or redundant)

run-locally/docker-compose.yml
marcjohnson-kint: Yes, I'm totally happy with that change. I'm glad you suggested it. Thank you.

When I put shared docker compose file together I was rushing to get schema upgrade testing done and did not (and still don't) fully understand Kafka configuration. Please do make improvements to it.

If you would also like to upgrade PostgreSQL to version 12 as well, that would be great too.

adamdickmeiss:
> @julianladisch
> 
> I think this work is unnecessary and confusing as the module already contains a docker compose file for running the module locally. There are also instructions provided (directly above the ones added).
> 

IMHO letting Docker running a module is not local. And most certainly is a major blow in being able to start and stop, debug a module like this.

(That Docker is local, does not make the module local.. It's in a container, to some extend isolated from a development environment).. And yes, there are tools to make these things easier)

So a procedure that will allow running it completely local is welcomed. IMHO.

marcjohnson-kint:
I'm closing this pull request because it has had no activity for a year. Please re-open if it is still wanted.

PR #667, opened 2021-09-17
MODINVSTOR-763  "Instance Iteration" API

dmtkachenko:
## Purpose 
Introduce "Instance Iteration API" for the purpose of contributing instance catalog to external system (INN-Reach). The functionality provided by the interface will be used during [Initial record contribution](https://wiki.folio.org/display/DD/D2IR+Record+Contribution+flow#D2IRRecordContributionflow-Initialrecordcontribution). In depth analysis and solution was provided in [SPIKE: [MODINREACH-80]](https://wiki.folio.org/display/FOLIJET/SPIKE%3A+%5BMODINREACH-80%5D+Record+Contribution%3A+Analyze+Re-Index+job+implementation+usage+in+mod-inventory-storage) 

JIRA: [MODINVSTOR-763](https://issues.folio.org/browse/MODINVSTOR-763), [MODINVSTOR-764](https://issues.folio.org/browse/MODINVSTOR-764), [MODINVSTOR-765](https://issues.folio.org/browse/MODINVSTOR-765)

## Approach
* implementation is based on existing Instance Re-indexing mechanism (see https://github.com/folio-org/mod-inventory-storage/pull/581) 
* implement instance iteration API: POST - submit a job, DELETE - cancel a job, GET - get job status.
* store the job statistic in DB table `iteration_job`;
* the iteration logic is in `IterationJobRunner`


README.MD
```diff
@@ -321,3 +321,48 @@ x-okapi-url: <okapi-url>
   "tenant": "<the-tenant-name>"
 }
 ```
+
+## Iteration of instances
```
marcjohnson-kint, line 325: Thank you for adding some documentation

README.MD
```diff
@@ -321,3 +321,48 @@ x-okapi-url: <okapi-url>
...
+
+Once iteration job has been started, it initiates streaming of all instance IDs and publishing domain events for them. 
+The domain event has the following structure:
+* Topic: <target-topic-name>
+* Partition key: `The instance id`
+* Payload:
+```javascript
+x-okapi-tenant: <tenant-id>
+x-okapi-url: <okapi-url>
+iteration-job-id: <job-id>
```
marcjohnson-kint, line 363: Are these represented differently because they are not part of the JSON and instead another part of the message?

ramls/instance-iteration.raml
```diff
@@ -0,0 +1,72 @@
...
+/instance-storage/instances/iteration:
+  displayName: Iterate instances
+  post:
+    description: Submit an iteration job
+    body:
+      application/json:
+        description: Iteration job parameters
+        type: iterationJobParams
+    responses:
+      200:
```
marcjohnson-kint, line 29: Given this creates a job, could it be more specific to use 201 to denote that a new resource has been created?

ramls/instance-iteration.raml
```diff
@@ -0,0 +1,72 @@
...
+        500:
+          description: Internal server error
+          body:
+            text/plain:
+              example: Internal server error
+    delete:
+      description: Cancel iteration job by id
+      responses:
+        204:
+          description: Iteration job has been cancelled
```
marcjohnson-kint, line 62: Does the cancellation also delete the job?

ramls/instance-iteration/iteration-job-params.json
```diff
@@ -0,0 +1,24 @@
...
+  "type": "object",
+  "javaType": "org.folio.rest.jaxrs.model.IterationJobParams",
+  "properties":
+  {
+    "eventType":
+    {
+      "description": "Type of events to be published",
+      "type": "string",
+      "default": "ITERATE"
+    },
```
marcjohnson-kint, line 13: What other possible event types are allowed? The DomainEventType class suggests that ITERATE is the only one possible.

If there is only one valid option, why give the client the choice?

If there are more, could they be explicitly documented as part of the schema using an enumeration?

ramls/instance-iteration/iteration-job.json
```diff
@@ -0,0 +1,40 @@
...
+    "jobStatus": {
+      "description": "Overall job status",
+      "type": "string",
+      "enum": [
+        "In progress",
+        "Id publishing failed",
+        "Ids published",
+        "Pending cancel",
+        "Id publishing cancelled"
+      ]
```
marcjohnson-kint, line 26: One thing I learnt from using enumerations for item status is that if the enumeration values take the form of presentable text, folks tend to think they can be presented as is and forget about any need for localization.

To avoid that, I would probably use formatting that makes them obviously a constant value e.g. uppercase with underscores, rather than directly presentable text.

ramls/instance-iteration/iteration-job.json
```diff
@@ -0,0 +1,40 @@
...
+    },
+    "jobStatus": {
+      "description": "Overall job status",
+      "type": "string",
+      "enum": [
+        "In progress",
+        "Id publishing failed",
+        "Ids published",
+        "Pending cancel",
+        "Id publishing cancelled"
```
marcjohnson-kint, line 25: The context is already provided by this being part of an instance iteration job (rather than a ID publishing job).

Some of the values include the context that it is IDs that are being published, others do not. I think it might be easier to interpret them if they were consistently named.

```suggestion
        "In progress",
        "Failed",
        "Completed",
        "Cancellation pending",
        "Cancelled"
```

src/main/java/org/folio/persist/IterationJobRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+import io.vertx.core.Context;
+import io.vertx.core.Future;
+import java.util.Map;
+import java.util.function.UnaryOperator;
+import org.folio.rest.jaxrs.model.IterationJob;
+
+public class IterationJobRepository extends AbstractRepository<IterationJob> {
+
+  public static final String TABLE_NAME = "iteration_job";
```
marcjohnson-kint, line 13: Why is this public? Is it used externally to the repository?

src/main/java/org/folio/persist/IterationJobRepository.java
```diff
@@ -0,0 +1,26 @@
...
+
+  public IterationJobRepository(Context context, Map<String, String> okapiHeaders) {
+    super(postgresClient(context, okapiHeaders), TABLE_NAME, IterationJob.class);
+  }
+
+  public Future<IterationJob> fetchAndUpdate(String id, UnaryOperator<IterationJob> builder) {
```
marcjohnson-kint, line 19: Why does a job need to be fetched in order to be updated?

src/main/java/org/folio/rest/impl/InstanceIterationAPI.java
```diff
@@ -0,0 +1,49 @@
...
+import io.vertx.core.Handler;
+import java.util.Map;
+import javax.ws.rs.core.Response;
+import org.folio.persist.IterationJobRepository;
+import org.folio.rest.jaxrs.model.IterationJob;
+import org.folio.rest.jaxrs.model.IterationJobParams;
+import org.folio.rest.jaxrs.resource.InstanceStorageInstancesIteration;
+import org.folio.services.iteration.IterationService;
+
+public class InstanceIterationAPI implements InstanceStorageInstancesIteration {
```
marcjohnson-kint, line 17: I like how succinct this implementation is

ramls/instance-iteration/iteration-job.json
```diff
@@ -0,0 +1,40 @@
...
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Iteration job",
+  "type": "object",
+  "properties": {
+    "id": {
+      "description": "Job id",
+      "type": "string",
+      "$ref": "../raml-util/schemas/uuid.schema"
+    },
+    "published": {
```
marcjohnson-kint, line 11: Does this refer to the number of instance ID messages published?

```suggestion
    "messagesPublished": {
```

src/main/java/org/folio/services/iteration/IterationJobRunner.java
```diff
@@ -0,0 +1,201 @@
...
+
+  public IterationJobRunner(Context vertxContext, Map<String, String> okapiHeaders) {
+    this(new PostgresClientFuturized(PgUtil.postgresClient(vertxContext, okapiHeaders)),
+      new IterationJobRepository(vertxContext, okapiHeaders),
+      vertxContext,
+      okapiHeaders);
+  }
+
+  public IterationJobRunner(PostgresClientFuturized postgresClient, IterationJobRepository repository,
+                          Context vertxContext, Map<String, String> okapiHeaders) {
```
marcjohnson-kint, line 57: Personally, I think there s no need to indent method signatures this much.

```suggestion
  public IterationJobRunner(PostgresClientFuturized postgresClient, IterationJobRepository repository,
      Context vertxContext, Map<String, String> okapiHeaders) {
```

src/main/java/org/folio/services/iteration/IterationJobRunner.java
```diff
@@ -0,0 +1,201 @@
...
+
+          logReindexCompleted(published, context);
+        }
+      });
+  }
+
+  private Future<RowStream<Row>> selectInstanceIds(IterationContext ctx) {
+    return postgresClient.selectStream(ctx.connection,
+      "SELECT id FROM " + postgresClient.getFullTableName(INSTANCE_TABLE));
+  }
```
marcjohnson-kint, line 105: Could this be part of the repository, allowing the details to be hidden from this class (and allow the table name to be private to the repository)?

src/main/java/org/folio/services/kafka/topic/KafkaTopic.java
```diff
@@ -13,6 +14,10 @@ public static KafkaTopic item(String tenantId, String environmentName) {
     return new KafkaTopic(qualifyName("inventory.item", environmentName, tenantId));
   }
 
+  public static KafkaTopic forName(String name, String tenantId, String environmentName) {
+    return new KafkaTopic(qualifyName(name, environmentName, tenantId));
+  }
```
marcjohnson-kint, line 19: Why is a general method needed for creating any topic name needed?

If it is, could the existing methods use this method as well?

src/main/resources/kafka-topics.json
```diff
@@ -11,6 +11,10 @@
     {
       "name": "inventory.holdings-record",
       "numPartitions": 50
+    },
+    {
+      "name": "inventory.instance-contribution",
```
marcjohnson-kint, line 16: What is an instance contribution? I thought the job published messages to an iteration topic?

src/test/java/org/folio/rest/api/IterationJobRunnerTest.java
```diff
@@ -0,0 +1,134 @@
...
+    assertThat(job.getJobStatus(), is(IDS_PUBLISHED));
+    assertThat(job.getSubmittedDate(), notNullValue());
+
+    // Should be a single iteration message for each instance ID generated in the row stream
+    await().atMost(5, SECONDS)
+      .until(FakeKafkaConsumer::getAllPublishedInstanceIdsCount, is(numberOfRecords));
+  }
+
+  @Test
+  public void canCancelReindex() {
```
marcjohnson-kint, line 78: Should this be re-indexing, when this test is for instance iteration?

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -118,6 +121,7 @@ public static void testBaseBeforeClass() throws Exception {
     kafkaConsumer = new FakeKafkaConsumer().consume(vertx);
     kafkaConsumer.removeAllEvents();
     instanceReindex = new InstanceReindexFixture(client);
+    instanceIteration = new InstanceIterationFixture(client);
```
marcjohnson-kint, line 124: Is this fixture needed by many test classes? If not, please consider defining it in the test classes that need it rather than here.

src/test/java/org/folio/services/iteration/IterationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+  @SuppressWarnings("unchecked")
+  public void canCancelIteration() {
+    var jobId = UUID.randomUUID().toString();
+
+    when(repository.fetchAndUpdate(eq(jobId), isA(UnaryOperator.class)))
+      .thenReturn(Future.succeededFuture());
+
+    var result = get(service.cancelIteration(jobId));
+
+    assertThat(result, nullValue());
```
marcjohnson-kint, line 69: Is there any need to verify the interactions with the repository?

How does the repository know that the update should be to cancel a job?

src/test/java/org/folio/services/iteration/IterationServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    var job = get(service.submitIteration(jobParams));
+
+    assertThat(job.getId(), notNullValue());
+    assertThat(job.getJobStatus(), is(IN_PROGRESS));
+    assertThat(job.getPublished(), is(0));
+    assertThat(job.getSubmittedDate(), notNullValue());
+    assertThat(job.getJobParams(), is(jobParams));
+
+    verify(runner, times(1)).startIteration(any());
```
marcjohnson-kint, line 56: Does it not matter what parameters the iteration job the parameters is started with?

src/test/java/org/folio/rest/api/IterationJobRunnerTest.java
```diff
@@ -0,0 +1,134 @@
...
+    var numberOfRecords = 1100;
+    var rowStream = new TestRowStream(numberOfRecords);
+    var iterationJob = iterationJob();
+    var postgresClientFuturized = spy(getPostgresClientFuturized());
+
+    doReturn(succeededFuture(rowStream))
+      .when(postgresClientFuturized).selectStream(any(), anyString());
+
+    get(repository.save(iterationJob.getId(), iterationJob).toCompletionStage()
+      .toCompletableFuture());
```
marcjohnson-kint, line 56: Extracting an overload of the `get` method that takes the return type of the `save` method and does the `.completionStage().toCompletableFuture()` could help reduce some of the noise from asynchronous operations on this line.

README.MD
dmtkachenko: yes, the above are Kafka message headers

ramls/instance-iteration.raml
dmtkachenko: Physically, no, it only stops the job. Job record is retained in DB

ramls/instance-iteration/iteration-job-params.json
dmtkachenko: Well, ideally eventType value should not be strictly defined, it's supposed to be whatever caller wants it to be. `ITERATE` just serves as a default value, in case the event type is not explicitly specified. That is why the schema doesn't contain any value restrictions, like enum.

But `DomainEvent` class accepts only enum values as event type. I had to define `ITERATE` in `DomainEventType`, otherwise there would be an exception during transformation from String representation to enum value. 
So there isinconsistency between flexible event type in interface definition and its representation in Java code that is hard to overcome.

src/main/java/org/folio/persist/IterationJobRepository.java
dmtkachenko: Yes, it's used in `InstanceIterationAPI`

ramls/instance-iteration.raml
dmtkachenko: Thanks, updated.

ramls/instance-iteration/iteration-job.json
dmtkachenko: Thanks, updated

src/main/java/org/folio/persist/IterationJobRepository.java
dmtkachenko: I've followed the pattern established by similar repository and method: [ReindexJobRepository.fetchAndUpdate()](https://github.com/folio-org/mod-inventory-storage/blob/11f09665e26bf47ebd841d47f860885cc278ff92/src/main/java/org/folio/persist/ReindexJobRepository.java#L18)

Why it's fetched first? Probably to get the latest representation of data and update it.

ramls/instance-iteration/iteration-job.json
dmtkachenko: Thanks, updated

src/main/java/org/folio/persist/IterationJobRepository.java
marcjohnson-kint: Could that code be moved to within the repository, meaning this can be kept private?

src/main/java/org/folio/services/kafka/topic/KafkaTopic.java
dmtkachenko: It's needed because topic name comes from the request and it could be arbitrary name (see iteration-job-params.json)

> could the existing methods use this method as well?
ok, i'll do that

README.MD
marcjohnson-kint: Maybe the documentation could be tweaked to make the difference between the headers and the body more explicit

ramls/instance-iteration.raml
marcjohnson-kint: > Physically, no, it only stops the job. Job record is retained in DB

In which case, maybe this could respond with 200 and an updated representation given that it doesn't actually delete the resource rather it changes it's state.

ramls/instance-iteration/iteration-job-params.json
marcjohnson-kint: > Well, ideally eventType value should not be strictly defined, it's supposed to be whatever caller wants it to be. 

What is the value of allowing the caller to change the event type?

As I understand it, event types are a way of conveying to a consumer the semantic meaning (what happened, though that isn't strictly true for these messages as they aren't really events) and what information should be provided.

> But `DomainEvent` class accepts only enum values as event type. I had to define `ITERATE` in `DomainEventType`, otherwise there would be an exception during transformation from String representation to enum value.

What will happen if the client provides a different event type?

This suggests to me that it will fail.

> So there is inconsistency between flexible event type in interface definition and its representation in Java code that forces all the values to be defined. This is hard to overcome unless making event type in `DomainEvent` of String type instead of enum, which is also probably not desirable.

Or we stop the client from providing the name of the event type. The type itself doesn't really change, as the same semantic meaning and structure is used whatever name is provided.

src/test/java/org/folio/rest/api/IterationJobRunnerTest.java
dmtkachenko: Updated. Thanks

src/main/java/org/folio/persist/IterationJobRepository.java
dmtkachenko: Updated. Thanks

README.MD
dmtkachenko: not sure how headers can be separated from the body
any suggestions?

ramls/instance-iteration.raml
dmtkachenko: It could, but 200 is not a standard response for `DELETE`. I'd leave as is

src/main/resources/kafka-topics.json
dmtkachenko: No, for contribution we need another, dedicated topic. Please review once again
[Initial record contribution](https://wiki.folio.org/display/DD/D2IR+Record+Contribution+flow#D2IRRecordContributionflow-Initialrecordcontribution) and [SPIKE: [MODINREACH-80]](https://wiki.folio.org/display/FOLIJET/SPIKE%3A+%5BMODINREACH-80%5D+Record+Contribution%3A+Analyze+Re-Index+job+implementation+usage+in+mod-inventory-storage)

marcjohnson-kint: I don't understand. What is this topic used for?

dmtkachenko: It's used by mod-inn-reach module to perform initial contribution of inventory to external system (Inn-Reach). Details about Initial Contribution, what it is for etc, can be found in the provided wiki pages (above)

marcjohnson-kint: Ok, I don't have the capacity to do that research and review that aspect.

It seems odd to me that is part of the same work as the new iteration API. And I don't understand why either iteration or the regular events aren't sufficient for that.

src/test/java/org/folio/rest/api/TestBase.java
dmtkachenko: Updated. Thanks

src/test/java/org/folio/rest/api/IterationJobRunnerTest.java
dmtkachenko: Updated. Thanks

src/main/java/org/folio/services/iteration/IterationJobRunner.java
dmtkachenko: Updated. Thanks

oliinyko (approved):
Looks good to me

andrew-khodyka (approved):

PR #668, opened 2021-09-20
MODINVSTOR-753: Sporadic ReindexJobRunnerTest.canReindexInstances failures

K-Felk:
in addition to fixing the noted test, I've noticed several other
tests in the itemStorageTest file sporadically fail.  It looks like
sometimes, the expected date stamp in the item status object 
will fall after the current datetime, rather than before.  My theory
is that because operations are asynchronous, the update sometimes won't 
have completed until after the current date is set in the test.

So for those tests, instead of comparing the date to the current 
datetime, I instead compare the datetime to the datetime of an 
earlier status, ensuring that the updated date, whatever it is,
is after the initial creation datetime.  I wrote a small function to 
do this comparison for tests that were dealing with string
representations, since there did not seem to be one in the 
matcher package.


marcjohnson-kint:
> in addition to fixing the noted test, I've noticed several other tests in the itemStorageTest file sporadically fail.

Thank you for identifying other test failures.

> instead of comparing the date to the current datetime, I instead compare the datetime to the datetime of an earlier status

What I've sometimes done is check that the actual date and time is within a certain tolerance e.g. 1 or 2 seconds of the expected date and time.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1282,12 +1282,13 @@ public void checkIfStatusDateExistsWhenItemStatusUpdated()
     assertThat(getResponse.getStatusCode(), is(HttpURLConnection.HTTP_OK));
 
     Item item = getResponse.getJson().mapTo(Item.class);
+    Instant finalStatusDate = item.getStatus().getDate().toInstant();
 
     assertThat(item.getId(), is(id.toString()));
 
     assertThat(item.getStatus().getName().value(), is("Checked out"));
 
-    assertThat(item.getStatus().getDate().toInstant(), withinSecondsBeforeNow(seconds(5)));
```
marcjohnson-kint, line 1290: Why would the item status date be after the current date and time? 

Even if the operations are done asynchronously, the date chosen should still be earlier than at the point of the assertion.

Have I missed something?

src/test/java/org/folio/rest/api/ReindexJobRunnerTest.java
```diff
@@ -78,7 +78,7 @@ public void canReindexInstances() {
 
     // Should be a single reindex message for each instance ID generated in the row stream
     await().atMost(5, SECONDS)
-      .until(FakeKafkaConsumer::getAllPublishedInstanceIdsCount, is(numberOfRecords));
+      .until(FakeKafkaConsumer::getAllPublishedInstanceIdsCount, greaterThanOrEqualTo(numberOfRecords));
```
marcjohnson-kint, line 81: I think it could be useful to explain why this is greater than rather than exact.

As I understand it, it should be exact, however it isn't in practice and we don't know why.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: That seems very odd, given that receiving the response _should_ mean that the API operation has completed and the `now` used in the assertion should be determined after that.

Thank you for investigating it.

PR #674, opened 2021-09-29
MODINVSTOR-794. Memory Leaks: io.vertx.core.impl.DuplicatedContext

SerhiiNosko:
Based on this picture reported in issue there was created more that 100K of DuplicatedContext objects that could not be garbage collected:
image

If we look into the internal state of Vertx Kafka producer in debug mode we see that for every producer's usage new Vertx write stream created(marked as blue) with link to shared producer(marked as green). In this particular example there was 3 usages, if we run 100K times - 100K write stream will be created and each object contains own DuplicatedContext object and all these objects will not be cleared by Garbage Collector. 
image

To prevent it if we use shared producers we need to invoke producer's close() method (in close's implementation the corresponding Vertx write stream is deleted from map and so it would be accessible for Garabge Collector and when all write stream deleted - physical invocation of producer's Kafka close() method is triggered)
Also the same we see in Vertx documentation for shared producer usage:
`When you are done with the producer, just close it, when all shared producers are closed, the resources will be released for you.`
Link: https://vertx.io/docs/vertx-kafka-client/java/#_sharing_a_producer


marcjohnson-kint:
@siarhei-charniak Thank you for asking me for a review. I don't think I have sufficient context or knowledge about the Kafka producer to provide a useful review.

KaterynaSenchenko (approved):

RuslanLavrov (approved):

Aliaksandr-Fedasiuk (approved):

PR #681, opened 2021-10-01
MODINVSTOR-774: Adds sql migration for publication period (back port)

julianladisch:
Back port from master to b21.0


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1078,6 +1078,11 @@
       "run": "after",
       "snippetPath": "inventory-hierarchy/addEffectiveHoldingsToItemsAndHoldingsViewDuringUpgrade.sql",
       "fromModuleVersion": "21.0.3"
+    },
+    {
+      "run": "after",
+      "snippetPath": "populatePublicationPeriod.sql",
+      "fromModuleVersion": "21.0.7"
```
marcjohnson-kint, line 1085: Why is this 21.0.7 when it is to be first included in 21.0.9?

Won't this mean that if folks upgrade from 21.0.8 to 21.0.9 this won't run (because 21.0.7 has already passed)?

julianladisch: Yes, good catch, thanks! Changed to 21.0.9.

marcjohnson-kint (approved):

julianladisch:
Closing, because even this SQL based migration takes to long for 10 million instances.

PR #683, opened 2021-10-05
MODINVSTOR-774: Disable automatic publicationPeriod migration

julianladisch:
Keep populatePublicationPeriod.sql for manual execution.

### DIFF
--- a/src/main/resources/templates/db_scripts/populatePublicationPeriod.sql
+++ b/src/main/resources/templates/db_scripts/populatePublicationPeriod.sql
@@ -36,15 +36,21 @@ CREATE OR REPLACE FUNCTION parse_publication_period(jsonb) RETURNS jsonb AS $$
   END;
 $$ LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT;
 
-ALTER TABLE instance DISABLE TRIGGER USER;
-
 UPDATE instance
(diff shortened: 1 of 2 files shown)


src/main/resources/templates/db_scripts/populatePublicationPeriod.sql
```diff
@@ -28,19 +36,19 @@ CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.parse_publication_period(
   END;
 $$ LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT;
 
-ALTER TABLE ${myuniversity}_${mymodule}.instance DISABLE TRIGGER USER;
+ALTER TABLE instance DISABLE TRIGGER USER;
```
marcjohnson-kint, line 39: If this script is going to be run whilst the system is operational and accessible, is it still appropriate to disable all of the triggers whilst it is being executed?

julianladisch: No.

marcjohnson-kint: That answer surprises me, I would have thought that disabling triggers could interfere with the integrity of the data in the system (presumably that's why they are there).

adamdickmeiss (approved):

PR #684, opened 2021-10-05
MODINVSTOR-809: Disable optimistic locking failures for v22.0.0

julianladisch:
Optimistic locking of instances, holdings and items will be set to "logOnConflict" for v22.0.0.


adamdickmeiss (approved):
So effectively no version enforcement. Good luck getting that back.

steveellis:
@julianladisch  and @adamdickmeiss was sad to see this considering all the work that has gone into it (mostly Julian's!). But it if it's only a couple of modules holding us back at least we're most of the way there and it will get into the next release.

PR #695, opened 2021-11-09
MODINVSTOR-24: Instance storage APIs do not validate

julianladisch:
Many APIs are missing the `@Validate` annotation on the
implementing method (on the method, not on the method parameters).

Adding it makes RMB validate all method parameters against the
RAML definition.

If a method has a `@Validate` annotation added to a method
parameter like `@Min(0) @Max(2147483647) int offset`
it must be removed because it causes this error:

`HV000151: A method overriding another method must not redefine
the parameter constraint configuration`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 23.0.0 in progress
+
+* Introduces validation against RAML API specs for 147 APIs (MODINVSTOR-24)
+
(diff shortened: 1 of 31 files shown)


adamdickmeiss (approved):
Interesting to see if this breaks some clients (that sent faulty requests that is).

steveellis (approved):

PR #696, opened 2021-11-11
Removed item parameters to collection get template MODINVSTOR-824

marcjohnson-kint:


marcjohnson-kint:
@MikeTaylor  @sduvvuri-ebsco  

Given that the RAML tools (api-lint in folio-tools) that FOLIO officially uses do not complain (at info level), it's difficult for me to assess whether change resolves the reported issue. I would appreciate your review of this change.

MikeTaylor (approved):
This looks good to me in so far as when I made the same change myself in my local reproduction of the problem, it made the problem go away. But surely the real issue is what the author of this code _meant_ by passing those two parameters, and if that intention is being thwarted in a way that he or she has not been aware of?

marcjohnson-kint:
> surely the real issue is what the author of this code _meant_ by passing those two parameters, and if that intention is being thwarted in a way that he or she has not been aware of?

I don't think so. 

I suspect it's likely a combination of:
* the typical copy-paste process that folks tend to use with RAML
* that the two common resource types (`collection` and `collection-get`) are very similar (yet take different parameters) 
* none of our official tooling identifies concerns with this

What I'm baffled about is why this particular case caused an outage and yet mod-graphql reports what seems to me to be identical errors elsewhere without failure:
```
 warning:Unused parameter: 'exampleItem' at instance-storage.raml:82:10
```

MikeTaylor:
The second except in the original description was generated by running in a situation where the `raml-utils` submodule had not been pulled in.

PR #699, opened 2021-11-15
MODINVSTOR-827 Clean kafka topics on tenant delete

OleksiiKuzminov:
### Purpose/Overview:
There are a lot of unused topics that can be on envs with several tenants. To decrease pressure on consumers modules it’s nice to clean unused topics in Kafka.

### Approach:

Delete topics for the tenant on the tenant delete operation

### DIFF
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -164,7 +164,7 @@ public class TenantRefAPI extends TenantAPI {
                          Handler<AsyncResult<Response>> handler, Context context) {
     // delete Kafka topics if tenant purged
     Future.succeededFuture()
-      .compose(x -> tenantAttributes.getPurge()
+      .compose(x -> tenantAttributes.getPurge() != null && tenantAttributes.getPurge()
         ? new KafkaAdminClientService(context.owner()).deleteKafkaTopics(TenantTool.tenantId(headers), environmentName())
         : Future.succeededFuture())
       .onComplete(x -> super.postTenant(tenantAttributes, headers, handler, context));
--- a/src/main/java/org/folio/rest/impl/TenantRefAPI.java
+++ b/src/main/java/org/folio/rest/impl/TenantRefAPI.java
@@ -1,12 +1,11 @@
 package org.folio.rest.impl;
 
-import static org.folio.Environment.environmentName;
-
+import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
-import java.util.stream.Collectors;
 import org.apache.commons.io.IOUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -15,8 +14,12 @@ import org.folio.okapi.common.GenericCompositeFuture;
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.TenantAttributes;
 import org.folio.rest.tools.utils.TenantLoading;
+import org.folio.rest.tools.utils.TenantTool;
 import org.folio.services.kafka.topic.KafkaAdminClientService;
+import org.folio.services.migration.BaseMigrationService;
+import org.folio.services.migration.item.ItemShelvingOrderMigrationService;
 
+import javax.ws.rs.core.Response;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URISyntaxException;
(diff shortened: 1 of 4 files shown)


marcjohnson-kint:
My primary concern with this change is the impact it may have on consumers.

As I understand it, consuming modules are configured to automatically create a topic if it isn't present at the time the consumer is defined.

That leads me to think that consumers are expecting a topic to be there.

Would this change mean that modules consuming these topics have the topic removed from underneath them? What impact will that have on those modules?

OleksiiKuzminov:
@marcjohnson-kint All consumers use subscription by pattern. In this case, there are no problems from mo-search side.

pfilippov-epam (approved):
Looks good.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -154,8 +159,19 @@ static boolean isNew(TenantAttributes attributes, String featureVersion) {
     return future.compose(result -> runJavaMigrations(attributes, vertxContext, headers).map(result));
   }
 
+  @Validate
+  public void postTenant(TenantAttributes tenantAttributes, Map<String, String> headers,
+                         Handler<AsyncResult<Response>> handler, Context context) {
+    // delete Kafka topics if tenant purged
+    Future.succeededFuture()
+      .compose(x -> tenantAttributes.getPurge()
```
julianladisch, line 167: `Future.succeededFuture().compose(x -> foo)` is a noop and should be replaced by `foo`.

julianladisch, line 167: A tenant can be disabled with purge=true or without purge=true: https://github.com/folio-org/okapi/blob/master/doc/guide.md#disabling

For me the intention of https://issues.folio.org/browse/MODINVSTOR-827 is to delete the kafka topics in both cases. Disabling without purge=true allows to upgrade to a new version without the need for keeping the kafka topics in the old version.

src/main/java/org/folio/services/kafka/topic/KafkaAdminClientService.java
```diff
@@ -53,11 +53,34 @@ public KafkaAdminClientService(Supplier<KafkaAdminClient> clientFactory) {
...
+      } else {
+        log.error("Unable to delete topics", res.cause());
+      }
+      kafkaAdminClient.close().onComplete(closeResult -> {
+        if (closeResult.failed()) {
+          log.error("Failed to close kafka admin client", closeResult.cause());
+        }
+      });
     });
+    return result.future();
```
julianladisch, line 79: Avoiding the Promise and using dedicated Future methods:
```
    return kafkaAdminClient.deleteTopics(topicsToDelete)
        .onSuccess(x -> log.info("Topics deleted successfully"))
        .onFailure(e -> log.error("Unable to delete topics", e))
        .eventually(x -> 
            kafkaAdminClient.close()
            .onFailure(e -> log.error("Failed to close kafka admin client", e)));
```

Creating and closing the AdminKafkaClient is duplicated code (createKafkaTopics) and can be extracted into a `withAdminKafkaClient` method:

```
  public Future<Void> deleteKafkaTopics(String tenantId, String environmentName) {
    List<String> topicsToDelete = readTopics()
      .map(topic -> qualifyName(topic, environmentName, tenantId))
      .map(NewTopic::getName)
      .collect(Collectors.toList());
    return withKafkaAdminClient(kafkaAdminClient -> kafkaAdminClient.deleteTopics(topicsToDelete))
        .onSuccess(x -> log.info("Topics deleted successfully"))
        .onFailure(e -> log.error("Unable to delete topics", e));
  }

  private <T> Future<T> withKafkaAdminClient(Function<KafkaAdminClient, Future<T>> function) {
    final KafkaAdminClient kafkaAdminClient = clientFactory.get();
    return function.apply(kafkaAdminClient)
        .eventually(x ->
            kafkaAdminClient.close()
            .onFailure(e -> log.error("Failed to close kafka admin client", e)));
  }
```

src/main/java/org/folio/rest/impl/TenantRefAPI.java
OleksiiKuzminov: @julianladisch  Can you please share an example of such requests for disabling without prge=true

MikhailFokanov (approved):
Looks good.

src/main/java/org/folio/rest/impl/TenantRefAPI.java
OleksiiKuzminov: @julianladisch we decided to delete topics only on purge

marcjohnson-kint:
@OleksiiKuzminov 

> All consumers use subscription by pattern. In this case, there are no problems from mo-search side.

What about other consumers other than mod-search?

Does that mean that the consumers don't automatically create the topics in the pattern themselves (and expect them to be there)?

Has mod-search been tested for tolerating the removal of a topic after consumption has started?

PR #700, opened 2021-11-15
MODINVSTOR-676: Add statistical code ids exists constraint for item create/update

wwelling:
https://issues.folio.org/browse/MODINVSTOR-676

Note that this is a workaround to supporting database constraints on arrays. This application layer validation is only done on create and update of a single item. I do not recommend this approach for batch item create as it may incur an undesirable performance issue requiring to iterate over item batch and performing service points lookup per item.


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2373,6 +2376,192 @@ public void canSearchByPurchaseOrderLineIdentifierProperty() throws Exception {
...
+      .withMaterialType(journalMaterialTypeId)
+      .withPermanentLoanType(canCirculateLoanTypeId)
+      .withStatus(status)
+      .withStatisticalCodeIds(Arrays.asList(statisticalCodeId))
+      .create();
+
+    final Response createdItem = itemsClient.attemptToCreate(itemToCreate);
+
+    assertThat(createdItem.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+  }
```
marcjohnson-kint, line 2405: Does this need to be a separate test? Could it be added to the general test for creating an item?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2373,6 +2376,192 @@ public void canSearchByPurchaseOrderLineIdentifierProperty() throws Exception {
...
+      ))
+      .create();
+
+    final Response createdItem = itemsClient.attemptToCreate(itemToCreate);
+
+    assertThat(createdItem.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
+  }
+
+  @Test
+  public void cannotCreateItemWithAllNonExistentStatisticalCodeId() throws Exception {
```
marcjohnson-kint, line 2468: What about the scenario where some of the statistical codes exist and others do not.

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +185,27 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
+  private Future<Item> refuseWhenServicePointIdDoesNotExist(final Item item) {
+    if (item.getStatisticalCodeIds().isEmpty()) {
+      return succeededFuture(item);
+    }
+
+    return statisticalCodeRepository.getById(item.getStatisticalCodeIds())
+      .compose(statCodeMap -> {
+        if (statCodeMap.size() == item.getStatisticalCodeIds().size()) {
+          return succeededFuture(item);
+        }
```
marcjohnson-kint, line 198: Does this mean that if the size of the result set is the same as the number of statistical codes in the item record, it is assumed they are all valid?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -107,6 +109,7 @@ public void beforeEach() {
     StorageTestSuite.deleteAll(itemsStorageUrl(""));
     StorageTestSuite.deleteAll(holdingsStorageUrl(""));
     StorageTestSuite.deleteAll(instancesStorageUrl(""));
+    statisticalCodeFixture.removeTestStatisticalCodes();
```
marcjohnson-kint, line 112: Is this fixture only used by tests?

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +185,27 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
   private Function<SQLConnection, Future<RowSet<Row>>> updateSingleItemBatchFactory(Item item) {
     return connection -> itemRepository.update(connection, item.getId(), item);
   }
+
+  private Future<Item> refuseWhenServicePointIdDoesNotExist(final Item item) {
```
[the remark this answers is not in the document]
wwelling: Yes, will update. Thanks.

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: That is this test.

src/main/java/org/folio/services/item/ItemService.java
wwelling: If the number of statistical code ids on the item to create or update matched the number of results of get statistical codes by ids they all exist.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: The title suggests to me that all of the statistical codes don't exist. Have I misunderstood the use of _all_ in the name?

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: I found it in another test. Will check where else it is used.

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: Sure, I can update name.

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: I do not see a general can create item method to move this to. What would you suggest?

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: This fixture is part of TestBase and only used in StatisticalCodeTest and now ItemStorageTest.

wwelling: If you are asking if it is used in multiple tests in ItemStorageTest, then yes it is.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: I suggest this one - https://github.com/folio-org/mod-inventory-storage/blob/a3eb856b8006d3e2246f142b0f271539c760cfde/src/test/java/org/folio/rest/api/ItemStorageTest.java#L202

wwelling: Done.

wwelling: Moving to that method causes other tests to fail! 

```
Failed tests: 

  ItemEffectiveCallNumberComponentsTest.canCalculateEffectiveCallNumberPropertyOnUpdate:261 

Expected: is <{"id":"b4e32eee-0b16-4fa9-997c-d52aa7bb3789","_version":2,"hrid":"it00000000036","holdingsRecordId":"aa5e032e-0917-4e4f-b7fa-34215da30d8f","formerIds":[],"effectiveShelvingOrder":"","itemLevelCallNumber":"initItCN","effectiveCallNumberComponents":{"callNumber":"initItCN"},"yearCaption":[],"notes":[],"circulationNotes":[],"status":{"name":"Available","date":"2021-11-16T15:35:51.035+00:00"},"materialTypeId":"c052063c-73ee-4692-8f32-ac412cdb3793","permanentLoanTypeId":"77d8bc76-46b3-441b-813b-b47920ca57f0","temporaryLocationId":"b2c78c43-f4ab-4fe5-9fc5-76f79a79cadd","effectiveLocationId":"b2c78c43-f4ab-4fe5-9fc5-76f79a79cadd","electronicAccess":[],"statisticalCodeIds":[],"metadata":{"createdDate":"2021-11-16T15:35:51.035+00:00","updatedDate":"2021-11-16T15:35:51.035+00:00"},"instanceId":"2eaa2452-5ee0-4ac9-97f8-73e7bd698eea"}>

     but: was <{"instanceId":"2eaa2452-5ee0-4ac9-97f8-73e7bd698eea","id":"b4e32eee-0b16-4fa9-997c-d52aa7bb3789","_version":1,"hrid":"it00000000036","holdingsRecordId":"aa5e032e-0917-4e4f-b7fa-34215da30d8f","formerIds":[],"effectiveShelvingOrder":"","itemLevelCallNumber":"initItCN","effectiveCallNumberComponents":{"callNumber":"initItCN"},"yearCaption":[],"notes":[],"circulationNotes":[],"status":{"name":"Available","date":"2021-11-16T15:35:51.035+00:00"},"materialTypeId":"c052063c-73ee-4692-8f32-ac412cdb3793","permanentLoanTypeId":"77d8bc76-46b3-441b-813b-b47920ca57f0","temporaryLocationId":"b2c78c43-f4ab-4fe5-9fc5-76f79a79cadd","effectiveLocationId":"b2c78c43-f4ab-4fe5-9fc5-76f79a79cadd","electronicAccess":[],"statisticalCodeIds":[],"metadata":{"createdDate":"2021-11-16T15:35:51.035+00:00","updatedDate":"2021-11-16T15:35:51.035+00:00"}}>
```

wwelling: I moved that requested change to another branch, https://github.com/folio-org/mod-inventory-storage/tree/MODINVSTOR-676-failure. There appears to be that tests are not independent.

marcjohnson-kint: You might find those are unrelated / false failures.

Those tests are very strict on the output check (they check the whole JSON), often the only difference is the `_version` property.

It is useful that it reveals something strange about how that property is incremented, however it makes the suite generally unstable.

wwelling: Well, now failing on any change. Something wrong with the tests. Going to pull this request change back in.

wwelling: Okay, it passed this time. I think there needs to be an issue to resolve the sporadic failing tests. Certainly beyond scope of this issue.

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: I opted for as much isolation as I could achieve. If you want I can create the statistical codes in before all and remove them in after all. Or are they already be removed by some abstraction?

wwelling: Or would you prefer I just move that remove call to each test creating stat codes?

marcjohnson-kint: I'm not sure what you mean, I haven't mentioned anything about isolation or moving code. 

I assume that method already deletes all statistical codes. The only thing I'd want to do is rename it to remove the redundant `test` from it's name.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: By all means raise the issue (one may already exist). As is often the case, these issues tend not to get prioritised.

src/test/java/org/folio/rest/api/ItemStorageTest.java
wwelling: Oh, I didn't name that method. Sorry for the misunderstanding. Do you think I should rename it in the context of this PR?

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +185,27 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
   private Function<SQLConnection, Future<RowSet<Row>>> updateSingleItemBatchFactory(Item item) {
     return connection -> itemRepository.update(connection, item.getId(), item);
   }
+
+  private Future<Item> refuseWhenStatisticalCodeIdDoesNotExist(final Item item) {
```
julianladisch, line 189: Can you add a comment to this pull request's description whether there will be an automated migration for non-existing statistical code ids? Or state that a warning will be added to the release notes that SysOps need to manually search for non-existing statistical code ids and replace them by an existing one, otherwise updating an item will be rejected?

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +185,27 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
+    if (item.getStatisticalCodeIds().isEmpty()) {
```
julianladisch, line 190: Can `item.getStatisticalCodeIds()` be null because statisticalCodeIds is not a required field?
https://github.com/folio-org/mod-inventory-storage/blob/v22.0.2/ramls/item.json#L428

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +185,27 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
+      return succeededFuture(item);
+    }
+
+    return statisticalCodeRepository.getById(item.getStatisticalCodeIds())
```
julianladisch, line 194: All other foreign keys are implemented using an PostgreSQL trigger.
The array can be checked with a trigger like this:
```
SET search_path TO diku_mod_inventory_storage;

CREATE OR REPLACE FUNCTION check_statistical_code_references()
RETURNS TRIGGER AS $$
DECLARE
  invalid text;
BEGIN
  SELECT ref
  INTO invalid
  FROM jsonb_array_elements_text(NEW.jsonb->'statisticalCodeIds') ref
  LEFT JOIN statistical_code ON id=ref::uuid
  WHERE id IS NULL
  LIMIT 1;
  IF FOUND THEN
    RAISE foreign_key_violation USING
      MESSAGE='statistical code doesn''t exist: ' || invalid,
      DETAIL='foreign key violation in statisticalCodeIds array of ' || TG_TABLE_NAME || ' with id=' || NEW.id,
      SCHEMA=TG_TABLE_SCHEMA,
      TABLE=TG_TABLE_NAME;
  END IF;
  RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER check_statistical_code_references_on_insert
  BEFORE INSERT ON item
  FOR EACH ROW
  WHEN (NEW.jsonb->'statisticalCodeIds' IS NOT NULL AND NEW.jsonb->'statisticalCodeIds' <> '[]')
  EXECUTE FUNCTION check_statistical_code_references();

CREATE TRIGGER check_statistical_code_references_on_update
  BEFORE UPDATE ON item
  FOR EACH ROW
  WHEN (NEW.jsonb->'statisticalCodeIds' IS NOT NULL AND NEW.jsonb->'statisticalCodeIds' <> '[]'
             AND OLD.jsonb->'statisticalCodeIds' IS DISTINCT FROM NEW.jsonb->'statisticalCodeIds')
  EXECUTE FUNCTION check_statistical_code_references();
```

What are the advantages of this Java-based code that requires yet another database query round-trip?

src/main/java/org/folio/services/item/ItemService.java
julianladisch, line 189: Can you add documentation what happens when a code is deleted from statistical_code table that is still in use in statisticalCodeIds field of an instance, holding or item, or open a Jira for that task?

src/main/java/org/folio/services/item/ItemService.java
wwelling: Why does RMB not do that?

wwelling: The advantage is to not subject module development to implement database triggers for every case where the "framework" does not support.

I am not in favor of application layer solution and will gladly close this PR for your solution.

src/main/java/org/folio/services/item/ItemService.java
wwelling: I suppose it could. Could add a null check. I think the generate model instantiates the array but probably shouldn't rely on that.

src/main/java/org/folio/services/item/ItemService.java
wwelling: Preferably, it would not allow for it to be deleted if in use and constrained so at the database level. Since that is unlikely the case, I assume some UI feature will break.

wwelling: I think this is a question for the product owner. I would suggest a migration task to remove the non existent ids in use.

src/main/java/org/folio/services/item/ItemService.java
wwelling: What are the advantages of a database trigger over a well designed database?

wwelling: I suppose that argument could be the same as not require module development to implement application layer validation for every case the "framework" does not support the database constraint.

Anyhow, I opted for this to hopefully reduce future technical debt and additional requirements to deliver feature. This may not be the case though.

src/main/java/org/folio/services/item/ItemService.java
wwelling: Moved to issue.

src/main/java/org/folio/services/item/ItemService.java
wwelling: Moved to issue, removed application layer validation, added sql script above and update schema.json.

src/main/java/org/folio/services/item/ItemService.java
wwelling: Moved to issue.

wwelling:
@julianladisch now have some interesting seemingly unrelated test failures. Any ideas? Will be digging into them tomorrow.

julianladisch (approved):
Looks good to me, thanks!

src/test/java/org/folio/rest/support/builders/ItemRequestBuilder.java
```diff
@@ -85,10 +85,7 @@ public JsonObject create() {
     put(itemRequest, "itemLevelCallNumberSuffix", itemLevelCallNumberSuffix);
     put(itemRequest, "itemLevelCallNumberTypeId", itemLevelCallNumberTypeId);
     put(itemRequest, "discoverySuppress", discoverySuppress);
-
-    JsonArray statisticalCodeIds = new JsonArray();
-    this.statisticalCodeIds.forEach(statCodeId -> statisticalCodeIds.add(statCodeId));
-    put(itemRequest, "statisticalCodeIds", statisticalCodeIds);
+    put(itemRequest, "statisticalCodeIds", new JsonArray(statisticalCodeIds));
```
wwelling, line 88: Well that is convenient. Thanks.

src/test/java/org/folio/rest/api/InventoryHierarchyViewTest.java
```diff
@@ -98,24 +97,22 @@ public void setUp() {
...
-    removeFaultyStatisticalCode();
-
+        .map(json -> UUID.fromString(json.getString("instanceId")))
+        .toArray(UUID[]::new);
+
+    requestInventoryHierarchyItemsAndHoldingsViewInstance(instanceIds, false, response -> {
+      undoFaultyViewFunction();
+      assertThat(response.getStatusCode(), is(500));
+      String message = new JsonObject(response.getBody()).getString("message");
+      assertThat(message, is("function get_items_and_holdings_view(unknown, unknown) does not exist"));
```
marcjohnson-kint, line 114: Why did these changes result in the behaviour of the hierarchy view (which is purely a read) changing?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -119,6 +121,11 @@ public void checkIdsAfterEach() {
     StorageTestSuite.checkForMismatchedIDs("item");
   }
 
+  @After
+  public void removeStatisticalCodes() {
+    statisticalCodeFixture.removeTestStatisticalCodes();
```
marcjohnson-kint, line 126: Why is this being done after and not before?

wwelling: Should not require cleanup of statistical codes before they are created. If other test class creates them, it should be responsible for cleaning them up. If the cleanup is done before there is a chance the last test of this class will create something that makes the next test class non isolated.

marcjohnson-kint: Ok. I'm biased towards resetting the state before a test starts. It means that the initial state is always cleaned up irrespective of what ran before and not resetting it after can make manual diagnostics easier (as the state is still present)

wwelling: I see. That is interesting. That puts the onus on developers to cleanup all other tests remnants while writing a test class for them to be isolated. I can see the benefit in ability to diagnose but not sure how as when test fails another will likely run and the before method will do the cleanup anyhow.

marcjohnson-kint: > That puts the onus on developers to cleanup all other tests remnants while writing a test class for them to be isolated

I think of it more as preparing the state I want. I can understand framing it the other way.

src/test/java/org/folio/rest/api/InventoryHierarchyViewTest.java
julianladisch: Because the test setup of serverErrorWrittenOutOnDatabaseError() changed from setFaultyStatisticalCode() to setFaultyViewFunction() - for obvious reasons.

PR #701, opened 2021-11-17
Describe how mod-source-record-storage  source records are not automatically deleted MODINVSTOR-828

marcjohnson-kint:
The presence of source records in mod-inventory-storage creates ambiguity between them and source records defined in mod-source-record-storage.

This has led to folks thinking that documentation in this module about automatic source record deletion, that refers to source records, refers to source records in mod-source-record-storage rather than in this module.

It has been requested that we change the wording to remove that ambiguity. The new wording has been provided by @a-mbreaux 

(Personally, I don't think these kinds of reference to other modules were intended in an interface definition, however I'm changing it at folks request)

### DIFF
--- a/ramls/instance-storage.raml
+++ b/ramls/instance-storage.raml
@@ -115,7 +115,7 @@ resourceTypes:
                   example: "Internal server error, contact administrator"
           description: |
             Delete the source record.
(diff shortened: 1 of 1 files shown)


julianladisch (approved):
Looks good, thanks!

K-Felk (approved):

PR #706, opened 2021-11-23
MODINVSTOR-841: Spurious failures canCalculateEffectiveCallNumberPropertyOnUpdate

julianladisch:
The spurious test failures indicate that the _version property
is still 1 but is expected to be 2.


adamdickmeiss (approved):
Yes. So we can close https://issues.folio.org/browse/MODINVSTOR-841

PR #708, opened 2021-11-24
MODINVSTOR-839: SampleDataTest TopicExistsException

julianladisch:


adamdickmeiss (commented):
Looks like you are fixing the symptom, not the problem. Perhaps there's a problem with the purge operation and Kafka topics. Or the init/update does not re(create) them properly.

julianladisch:
Solved by https://github.com/folio-org/mod-inventory-storage/pull/714

PR #712, opened 2021-11-25
MODINVSTOR-831 Add Reindex support for Authority records

OleksiiKuzminov:


pfilippov-epam (approved):

MikhailFokanov (approved):
Looks good.

src/main/java/org/folio/services/reindex/ReindexJobRunner.java
```diff
@@ -47,26 +51,40 @@ public ReindexJobRunner(Context vertxContext, Map<String, String> okapiHeaders)
...
   }
 
-  public void startReindex(ReindexJob reindexJob) {
+  public void startReindex(ReindexJob reindexJob, ReindexResourceName reindexResourceName) {
     workerExecutor.executeBlocking(
-      promise -> streamInstanceIds(new ReindexContext(reindexJob))
-        .map(notUsed -> null)
-        .onComplete(promise))
+        promise -> {
+          switch (reindexResourceName) {
```
MikhailFokanov, line 76: switch without "default" and without last "break" looks weird especially in lambda.

PR #714, opened 2021-11-26
createTopics as loop

julianladisch:
Note:

This pull request is intended to be merged into the other pull request kafka-topics-insert-always.


adamdickmeiss (approved):
This looks safer than just createTopics once. Topics already existing because they really do already exist shouldn't be a problem. Topics that are marked to be deleted are problematic.. will they appear again with createTopics (despite exception) or not? It *seems* from unit tests they must be. But the looping approach  is certainly safer and ends up with no exceptions in the final iteration.

PR #723, opened 2021-12-14
[MODINVSTOR-849] New API for item checkout

OleksandrVidinieiev:
**Purpose**
Expose a new API for updating the item record in scope of a checkout process.
See [MODINVSTOR-849](https://issues.folio.org/browse/MODINVSTOR-849).

**Approach**
- make sure that item exists
- make sure that item is not already checked out
- change item status to `Checked out`
- if item contains `inTransitDestinationServicePointId` property, then remove it (this step was copied from current checkout implementation of mod-circulation)

### DIFF
--- a/src/test/java/org/folio/rest/api/ItemStorageTest.java
+++ b/src/test/java/org/folio/rest/api/ItemStorageTest.java
@@ -2826,14 +2826,12 @@ public class ItemStorageTest extends TestBaseWithInventoryUtil {
     assertThat(item.getJsonObject("status").getString("name"), is(status.value()));
   }
 
+  @SneakyThrows
   private Response checkOutItem(String itemId) {
     CompletableFuture<Response> postCompleted = new CompletableFuture<>();
     client.post(itemsStorageUrl("/" + itemId + "/check-out"), null,
       StorageTestSuite.TENANT_ID, ResponseHandler.empty(postCompleted));
-    try {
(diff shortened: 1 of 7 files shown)


src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -2779,4 +2821,19 @@ private Boolean isBefore(String dateTime, String secondDateTime) {
...
+
+  private Response checkOutItem(String itemId) {
+    CompletableFuture<Response> postCompleted = new CompletableFuture<>();
+    client.post(itemsStorageUrl("/" + itemId + "/check-out"), null,
+      StorageTestSuite.TENANT_ID, ResponseHandler.empty(postCompleted));
+    try {
+      return postCompleted.get(5, TimeUnit.SECONDS);
+    } catch (InterruptedException | ExecutionException | TimeoutException e) {
+      throw new RuntimeException(e);
+    }
```
marcjohnson-kint, line 2837: This can be replaced by the `SneakyThrows` annotation from lombok

OleksandrVidinieiev: Replaced, thanks

marcjohnson-kint:
Closing this because this experimental work has been superseded (it was decided that this approach would not continue)

PR #726, opened 2021-12-15
MODINVSTOR-851: Upgrade to RMB 33.1.3 (Juniper).

kaladay:
Resolve CVE-2021-44228 as per [MODINVSTOR-851](https://issues.folio.org/browse/MODINVSTOR-851).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 21.0.10 2021-12-15
+
+* Upgrade to RMB 33.1.3. (CVE-2021-44228) (MODINVSTOR-851)
+
 ## 21.0.9 2021-10-05
 
 * Fix memory leaks by closing Vertx Kafka producers (MODINVSTOR-794)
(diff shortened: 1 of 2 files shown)


marcjohnson-kint:
Has it been resolved that only upgrading RMB is sufficient? And we don't need to upgrade other libraries as well, e.g. the Kafka wrapper?

kaladay:
@marcjohnson-kint 

I think this module doesn't need the change.
The dependency tree looks like:
```
mvn dependency:tree -Dincludes=org.apache.logging.log4j
...
[INFO] org.folio:mod-inventory-storage:jar:21.0.10-SNAPSHOT
[INFO] \- org.folio:domain-models-runtime:jar:33.1.3:compile
[INFO]    +- org.apache.logging.log4j:log4j-core:jar:2.16.0:compile
[INFO]    \- org.apache.logging.log4j:log4j-api:jar:2.16.0:compile
```

I do not see log4j being pulled in via another other way in this module.

Also, I am going to close and recreate this PR and the other PR's so that I get the correct Jira ticket numbers on them.
There should only be a single PR for Kiwi, which should be against b22.0.
Juniper will have its own PR against b21.0.

kaladay:
Closing with the plans to re-create with the correct Ticket numbers.
Each release is using different tickets.

PR #731, opened 2022-01-12
[MODINVSTOR-637] Add disclosure file.

kaladay:
### Purpose:
Provide personal data disclosure form to comply with Folio policies and guidelines.
See: [MODINVSTOR-637](https://issues.folio.org/browse/MODINVSTOR-637).

### Approach:
Provide disclosure document.
Add checkmarks in disclosure document where appropriate.

Holdings has circulationNotes which allows for "source.personal".
The source.personal appears to allow for firstName and lastName at the very list.

The instance contributors may have middle names in them.

There are several structures with "notes" that are free form text.
These could potentially be used to store personal data.
One example is a holdings record.

The userIds are used and stored.

### DIFF
--- a/PERSONAL_DATA_DISCLOSURE.md
+++ b/PERSONAL_DATA_DISCLOSURE.md
@@ -1,0 +1,58 @@
+
+
+## Overview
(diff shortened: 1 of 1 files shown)


marcjohnson-kint (approved):
Kudos for finding the usage of first and last name in the note's source.

K-Felk (approved):

PR #733, opened 2022-01-24
MODINVSTOR-776: PUT /item-storage/items/$id performance

julianladisch:
Code changes to avoid duplicate database queries.


src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -176,4 +199,82 @@ public ItemService(Context vertxContext, Map<String, String> okapiHeaders) {
...
+  private static class PutData {
+    private Item item;
+    private HoldingsRecord holdingsRecord;
+  }
+
+  /**
+   * Fetch item with itemId and holdings record of that item from database
+   * and save them in putData.
+   */
+  private Future<Void> getItemAndHolding(String itemId, String holdingsId, PutData putData) {
```
marcjohnson-kint, line 212: This method is used in a future composition. The compose does not use the value of the future. I think this method and the chain it's used in could be easier to understand by returning an instance of PutData in the future rather than mutating the passed in parameter.

julianladisch: Changed.

K-Felk (approved):

PR #739, opened 2022-02-10
MODINVSTOR-861: Add related instances record API

K-Felk:


rladdusaw (changes requested):
A few files are missing a terminating newline, everything else looks fine.

K-Felk:
> A few files are missing a terminating newline, everything else looks fine.

Added.

ramls/related-instances.raml
```diff
@@ -0,0 +1,59 @@
...
+    description: "Delete all related instances"
+    is: [language]
+    responses:
+      204:
+        description: "All related instances deleted"
+      500:
+        description: "Internal server error, e.g. due to misconfiguration"
+        body:
+          text/plain:
+            example: "Internal server error, contact administrator"
```
marcjohnson-kint, line 53: Does the implementation of this use RMB's PgUtil? 

If so, I believe it might try to respond with 404 when the record to be deleted cannot be found, which might cause an error if that is not a described status code.

src/main/java/org/folio/rest/impl/RelatedInstanceAPI.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  @Override
+  public void deleteInstanceStorageRelatedInstancesByRelatedInstanceId(String instanceId, String lang,
+      Map<String, String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler,
+      Context vertxContext) {
+
+    PgUtil.deleteById(RELATED_INSTANCE_TABLE, instanceId, okapiHeaders, vertxContext,
+      DeleteInstanceStorageRelatedInstancesByRelatedInstanceIdResponse.class, asyncResultHandler);
+  }
```
marcjohnson-kint, line 64: Does this delete all related instance records related to a specific instance?

How would a client only delete a single relationship between two instances?

src/main/java/org/folio/rest/impl/RelatedInstanceAPI.java
```diff
@@ -0,0 +1,100 @@
...
+      Context vertxContext) {
+    if (checkInstanceRelatedToSelf(entity)) {
+        asyncResultHandler.handle(cannotBeRelatedToSelf());
+        return;
+    }
+    new RelatedInstanceService(vertxContext, okapiHeaders)
+      .updateRelatedInstance(entity)
+      .onSuccess(response -> asyncResultHandler.handle(Future.succeededFuture(response)))
+      .onFailure(handleFailure(asyncResultHandler));
+  }
```
marcjohnson-kint, line 78: This seems to take the instance ID as a parameter. Does that mean that there can only be a single related instance for each instance? If not, how does the client refer to each related instance record?

descriptors/ModuleDescriptor-template.json
```diff
@@ -2504,6 +2531,31 @@
       "displayName": "inventory view - get instances by query with their holdings and items",
       "description": "get instances by query with their holdings and items"
     },
+    {
+      "permissionName":"inventory-storage.related-instances.collection.get",
+      "displayName": "inventory storage - get related-instances collection",
+      "description": "get related-instances collection from storage"
+    },
```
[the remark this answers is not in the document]
K-Felk: There is.  I'll add.

src/main/java/org/folio/rest/impl/RelatedInstanceAPI.java
K-Felk: See above.  Relationships are modified and deleted by their own unique UUID, not the UUID of the instance records involved.

src/main/java/org/folio/rest/impl/RelatedInstanceAPI.java
marcjohnson-kint: > I think the use of "instance" is throwing you here. The instanceId being used here is the UUID assigned to the relationship itself, not the instance records involved. 

That is likely. I find that confusing and a little misleading. What controls that naming?

bq. In other words, the ID of an instance of a relationship.

Hmm, I find that terminology confusing.

bq. I worried this might be confusing, but the other crud operations and endpoints all use "instanceID" in this context to refer to the UUID of a specific data entry, so there's a consistency issue as well.

Which other CRUD endpoints are you referring to?

ramls/related-instances.raml
K-Felk: I believe I tested for this, but I'll try again to be sure.

src/main/java/org/folio/rest/impl/RelatedInstanceAPI.java
K-Felk: Hrm.  Well, on checking again, it seems that naming of the UUID in update/delete functions isn't as uniform as I thought at first.  I'll change the name of the element to "id"

marcjohnson-kint: Thank you

K-Felk: clarification:  are you asking me to change the name of the instanceId parameter, or the name of the function (deleteInstanceStorageRelatedInstancesByRelatedInstanceId)?  The latter I can change easily, the former is based on the name of the element in the RAML.  the "byinstanceId" part is auto-generated by the code that stubs out the CRUD methods.

ramls/related-instances.raml
K-Felk: Just retested, the endpoint throws a 404 when trying to delete an instance that does not exist with no problems.

K-Felk:
Given that this is already merged, how would you like me to handle changes?  Should I create a new branch from master, and raise a PR from that?

marcjohnson-kint:
> Given that this is already merged, how would you like me to handle changes? Should I create a new branch from master, and raise a PR from that?

Yes, raise a new pull request from a fresh branch.

PR #744, opened 2022-02-17
MODINVSTOR-862 Implement non-blocking Java migration for publicationPeriod as a separate process

OleksiiKuzminov:
fix subscription for async migration consumer


MikhailFokanov (approved):
Looks good.

PR #748, opened 2022-02-28
Removing of redundant full-text indexes

MikhailFokanov-ebsco:
This is a draft PR because the usage of the title full-text index in mod-orders has not been removed.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-inventory-storage</artifactId>
   <groupId>org.folio</groupId>
-  <version>23.0.0-SNAPSHOT</version>
+  <version>23.0.1-SNAPSHOT</version>
   <licenses>
     <license>
       <name>Apache License 2.0</name>
--- a/src/main/resources/templates/db_scripts/schema.json
+++ b/src/main/resources/templates/db_scripts/schema.json
@@ -375,36 +375,6 @@
         }
       ],
       "ginIndex": [
-        {
-          "fieldName": "title",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
-          "fieldName": "indexTitle",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
-          "fieldName": "contributors",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
-          "fieldName": "identifiers",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        },
-        {
-          "fieldName": "subjects",
-          "tOps": "DELETE",
-          "caseSensitive": false,
-          "removeAccents": true
-        }
       ],
       "index": [
         {
@@ -463,59 +433,6 @@
         }
       ],
       "fullTextIndex": [
-        {
-          "fieldName": "classifications",
-          "arraySubfield" : "classificationNumber",
-          "arrayModifiers": ["classificationTypeId"]
-        },
-        {
-          "fieldName": "keyword",
-          "multiFieldNames": "title,indexTitle,alternativeTitles[*].alternativeTitle,series,contributors[*].name,identifiers[*].value"
-        },
-        {
-          "fieldName": "allTitles",
-          "multiFieldNames": "title,indexTitle,alternativeTitles[*].alternativeTitle,series"
-        },
-        {
-          "fieldName": "contributorsNames",
-          "multiFieldNames": "contributors[*].name"
-        },
-        {
-          "fieldName": "title"
-        },
-        {
-          "fieldName": "languages"
-        },
-        {
-          "fieldName": "contributors",
-          "arraySubfield" : "name",
-          "arrayModifiers": ["contributorNameTypeId"]
-        },
-        {
-          "fieldName": "identifiers",
-          "arraySubfield" : "value",
-          "arrayModifiers": ["identifierTypeId"]
-        },
-        {
-          "fieldName": "invalidIsbn",
-          "sqlExpression" : "normalize_invalid_isbns(jsonb->'identifiers')",
-          "sqlExpressionQuery": "normalize_digits($)"
-        },
-        {
-          "fieldName": "isbn",
-          "sqlExpression" : "normalize_isbns(jsonb->'identifiers')",
-          "sqlExpressionQuery": "normalize_digits($)"
-        },
-        {
-          "fieldName": "dateOfPublication",
-          "sqlExpression" : "concat_array_object_values(jsonb->'publication', 'dateOfPublication')"
-        },
-        {
-          "fieldName": "subjects",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        }
       ]
     },
     {
@@ -702,12 +619,6 @@
         }
       ],
       "ginIndex": [
-        {
-          "fieldName": "electronicAccess",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        }
       ],
       "index": [
         {
@@ -738,16 +649,6 @@
         }
       ],
       "fullTextIndex": [
-        {
-          "fieldName": "callNumberAndSuffixNormalized",
-          "sqlExpression" : "normalize_holdings_call_number_and_suffix(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
-        },
-        {
-          "fieldName": "fullCallNumberNormalized",
-          "sqlExpression" : "normalize_holdings_full_call_number(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
-        }
       ]
     },
     {
@@ -864,24 +765,8 @@
         }
       ],
       "ginIndex": [
-        {
-          "fieldName": "status.name",
-          "tOps": "ADD",
-          "caseSensitive": false,
-          "removeAccents": true
-        }
       ],
       "fullTextIndex": [
-        {
-          "fieldName": "callNumberAndSuffixNormalized",
-          "sqlExpression" : "normalize_item_call_number_and_suffix(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
-        },
-        {
-          "fieldName": "fullCallNumberNormalized",
-          "sqlExpression" : "normalize_item_full_call_number(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
-        }
       ]
     },
     {


marcjohnson-kint:
@MikhailFokanov How do you know that no clients are making requests with CQL queries that would use these indexes (and with their removal would perform worse)?

MikhailFokanov:
> @MikhailFokanov How do you know that no clients are making requests with CQL queries that would use these indexes (and with their removal would perform worse)?

As we agreed in Core Platform (@skoczko, @adamdickmeiss ) I ran karate tests for all modules on rancher environment and got RMB warnings that GIN indexes are absent only for mod-orders tests. We are working on mod-orders in order to get rid of this usage.

I do agree that the this is a huge change, that should be comprehensively tested. I plan to deploy it to environment with bugfest data and ask POs to check the core functionality of their modules. Unfortunately, the whole testing of modules can be done only during Bugfest (as there are a lot of test cases).

But this change is needed because it significantly decreases the deployment time in case of data migrations.

adamdickmeiss:
I don't like that removing the entries in schema json affects how we query. I think it would preferable to ensure all generated SQL is the same -- only that indices are not created. Of course if this is never queried by anything, fine.

marcjohnson-kint:
> I don't like that removing the entries in schema json, affects how we query. I think it would preferable to ensure all generated SQL is the same -- only that indices are not created. 

Agreed. Is there logic in RMB that checks for the presence of indexes to influence the CQL -> SQL transformation?

> Of course if this is never queried by anything, fine.

There is no way to guarantee this at the moment. It is still entirely possible for a client to make requests with these queries.

marcjohnson-kint:
> > @MikhailFokanov How do you know that no clients are making requests with CQL queries that would use these indexes (and with their removal would perform worse)?
> 
> As we agreed in Core Platform (@skoczko, @adamdickmeiss ) I ran karate tests for all modules on rancher environment and got RMB warnings that GIN indexes are absent only for mod-orders tests. We are working on mod-orders in order to get rid of this usage.

I wasn't aware of that decision. Thank you for the information.

I think it is very unlikely that the Karate API tests demonstrate many of the different queries that can be / are issued to the module. They tend to only follow a fraction of the possible paths through the system.

Has there been investigation into how different parts of the system e.g. data import, data export, OAI-PMH, Z39.50, use the current CQL based search in mod-inventory-storage? 

I believe at least some of these use CQL queries with mod-inventory-storage and I doubt those are exercised via the Karate tests.

> I do agree that this is a huge change, that should be comprehensively tested. I plan to deploy it to rancher environment with bugfest data and ask POs to check the core functionality of their modules. Unfortunately, the whole testing of modules can be done only during Bugfest (as there are a lot of test cases).

I'm glad that there is a plan for manually testing this. My past experience on this project is that the POs rarely have the capacity to do this kind of testing thoroughly enough to find these kinds of issues and instead we only find them when they are in production.

One of my concerns is that the negative impact of this change will come to the Prokopovych team not the Core Platform team.

> But this change is needed because it significantly decreases the deployment time in case of data migrations.

I understand the motivation for these changes.

Though I thought that the Falcon team implemented a separate upgrade process in mod-inventory-storage to do it separately to the tenant API process.

It seems like these are two solutions to the same challenge?

adamdickmeiss:
Of course the net result of the effort is to remove indices.

I'm asked to look at the failures in this PR by @MikhailFokanov  and when I know look at them in mod-inventory-storage, my thinking is that we should instead only remove indices and not change query handling.

If the decision is now in-stone that the effort is a mod-inventory-storage then this is all we can do

(RMB would need a setting 'createindex' : true / false or something)

marcjohnson-kint:
> Of course the net result of the effort is to remove indices.
> 
> I'm asked to look at the failures in this PR by @MikhailFokanov and when I know look at them in mod-inventory-storage, my thinking is that we should instead only remove indices and not change query handling.
> 
> If the decision is now in-stone that the effort is a mod-inventory-storage then this is all we can do
> 
> (RMB would need a setting 'createindex' : true / false or something)

Thanks @adamdickmeiss  I hadn't reviewed the changes themselves. I didn't realise these changes also changed what CQL queries the module could accept.

pom.xml
```diff
@@ -2,7 +2,7 @@
   <modelVersion>4.0.0</modelVersion>
   <artifactId>mod-inventory-storage</artifactId>
   <groupId>org.folio</groupId>
-  <version>23.1.0-SNAPSHOT</version>
+  <version>23.1.1-SNAPSHOT</version>
```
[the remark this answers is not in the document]
adamdickmeiss: I think this is right.. He updated to whatever is in master. Which is currently on that next version. Later we can decide what PR should actually update the next version. May be this one.. Maybe at release time.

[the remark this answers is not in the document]
marcjohnson-kint: > I think this is right.. He updated to whatever is in master. Which is currently on that next version.

Our release docs explicitly state that the version should move to the next minor upon formal release. Given that the minor has already changed on the mainline and no further releases have been made, why increment the patch version at all?

> Later we can decide what PR should actually update the next version. May be this one.. Maybe at release time.

I also disagree with this. 

I've found that deferring the version change on the release tends to have two effects:
* the person releasing doesn't realise that the version needs to change (and so no change occurs)
* there are no snapshot builds for the intended release version

Personally, I think the version should be incremented at the point when the changes are made that justify the version change.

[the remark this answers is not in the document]
adamdickmeiss: Oops.. Forget what I said. the version in this branch should be same as in master - most certainly not with only patch changed.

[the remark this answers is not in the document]
adamdickmeiss: As for when the major/minor is changed.. It goes two ways .. If it's anybody's job, then in effect anybody needs to know if the major/minor is  *already* changed.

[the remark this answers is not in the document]
marcjohnson-kint: > It goes two ways .. If it's anybody's job, then in effect anybody needs to know if the major/minor is _already_ changed.

That's true. 

I may be being a selfish lead maintainer, in thinking that it's easier for folks to check if the version has already changed rather than the lead maintainer needing to understand the importance of every change made during the release cycle.

In theory, it should be easy to tell (although it assumes folks follow the release docs). 

If it's on a non-zero minor, then the major hasn't changed. If it's on a zero minor then it has already changed and no need to do it again.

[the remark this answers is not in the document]
adamdickmeiss: If people actually update fix-version for their stuff and also (they should) tell in issue if this is a breaking change as well in the commit message, then it's easier.. But generally, it's good that the release manager has a PO-role. If they don't know anything about what they are releasing and why, then it's a problem regardless.

adamdickmeiss:
@MikhailFokanov we discussed at the core-platcorm and thought it would be best to proceed with this.. So I think the proper solution is to remove or update tests that fail. I think you should set version to 24.0.0-SNAPSHOT to indicate "change of behavior".

pom.xml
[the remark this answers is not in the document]
marcjohnson-kint: > If people actually update fix-version for their stuff and also (they should) tell in issue if this is a breaking change as well in the commit message, then it's easier.. 

I haven't found that folks in FOLIO do most of those things. Typically commit messages don't tell folks much in the future and issues rarely include a statement about how they affect the modules compatibility.

> But generally, it's good that the release manager has a PO-role. If they don't know anything about what they are releasing and why, then it's a problem regardless.

None of the lead maintainers responsible for formal releases of modules are POs.

All of the lead maintainers are developers who may not be involved in many of the changes in that are made to a module. 

This change is a good example of that, it is being done by a team separate from the lead maintainer and it is only chance that they (me) are aware of it.

src/test/java/org/folio/rest/api/CallNumberNormalizedTestBase.java
```diff
@@ -18,103 +18,11 @@
 
   protected abstract List<String> searchByCallNumberNormalized(String callNumber) throws Exception;
 
-  @Parameters({
```
julianladisch, line 21: The three classes CallNumberNormalizedTestBase, HoldingsCallNumberNormalizedTest.java, ItemCallNumberNormalizedTest can be completely deleted. The remaining tests methods cannotSearchByNonAlphanumericsOnly and cannotSearchBySuffixOnly don't server any purpose any longer.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -1522,14 +1332,6 @@ public void cannotFindIsbnWithInnerStringAndTruncation() {
     canSort("isbn = 552*");
   }
 
-  // Interesting Times has two ISBNs: 0552167541, 978-0-552-16754-3
```
julianladisch, line 1525: The test methods cannotFindIsbnWithTailString, cannotFindIsbnWithInnerStringAndTruncation, cannotFindIsbnInInvalidIsbn, cannotFindInvalidIsbnInIsbn no longer serve any purpose and should also been removed.

src/test/java/org/folio/rest/api/AsyncMigrationTest.java
```diff
@@ -93,6 +94,7 @@ public void canGetAvailableMigrations() {
   }
 
   @Test
+  @Ignore
```
julianladisch, line 97: An `@Ignore` should have a reason: `@Ignore("Succeeds on Jenkins but fails consistently on local machines")`

julianladisch:
I've done some searches for usages of the removed indexes. All small results sets are OK because I can confirm that each one queries mod-search or doesn't query at all. Big result set are marked with ?? and I couldn't completely checked them.
?? classifications instance https://github.com/search?q=org%3Afolio-org+classifications+instance&type=Code
?? keyword instance https://github.com/search?q=org%3Afolio-org+keyword+instance&type=Code
OK alltitles https://github.com/search?q=org%3Afolio-org+allTitles&type=Code
OK contributorsNames https://github.com/search?q=org%3Afolio-org+contributorsNames&type=Code
OK contributorNameTypeId https://github.com/search?l=Java&p=3&q=org%3Afolio-org+contributorNameTypeId&type=Code
OK CONTRIBUTOR_NAME_TYPE_ID https://github.com/search?q=org%3Afolio-org+CONTRIBUTOR_NAME_TYPE_ID&type=Code
?? identifierTypeId https://github.com/search?q=org%3Afolio-org+identifierTypeId&type=code
OK invalidIsbn https://github.com/search?p=1&q=org%3Afolio-org+invalidIsbn&type=Code
?? isbn https://github.com/search?q=org%3Afolio-org+isbn&type=Code
?? dateOfPublication https://github.com/search?q=org%3Afolio-org+dateOfPublication&type=Code
OK callNumberAndSuffixNormalized https://github.com/search?q=org%3Afolio-org+callNumberAndSuffixNormalized&type=Code
OK fullCallNumberNormalized https://github.com/search?q=org%3Afolio-org+fullCallNumberNormalized&type=Code
?? status.name https://github.com/search?q=org%3Afolio-org+%22status+name%22&type=Code
OK callNumberAndSuffixNormalized https://github.com/search?q=org%3Afolio-org+callNumberAndSuffixNormalized&type=Code
OK fullCallNumberNormalized https://github.com/search?q=org%3Afolio-org+fullCallNumberNormalized&type=Code

Modules that use one of these deleted index names can easily switch to mod-search. Therefore I agree with this change. This will speed up all insert, update and delete operations because the database no longer need to keep these indexes in sync with the data.

julianladisch (approved):
Looks good, thanks!

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -738,16 +670,6 @@
...
       "fullTextIndex": [
-        {
-          "fieldName": "callNumberAndSuffixNormalized",
-          "sqlExpression" : "normalize_holdings_call_number_and_suffix(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
-        },
-        {
-          "fieldName": "fullCallNumberNormalized",
-          "sqlExpression" : "normalize_holdings_full_call_number(jsonb)",
-          "sqlExpressionQuery": "normalize_call_number_string($) || '*'"
```
julianladisch, line 749: Can createCallNumberNormalizationFunctions.sql been deleted? And a dropCallNumberNormalizationFunctions.sql been added to schema.json that removes the functions?

julianladisch (changes requested):
createCallNumberNormalizationFunctions.sql should be deleted, and dropCallNumberNormalizationFunctions.sql been added.

MikhailFokanov:
@julianladisch could you please review if I correctly fix your comments?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1125,6 +1031,11 @@
       "run": "after",
       "snippetPath": "inventory-hierarchy/correctGetUpdatedInstanceIdsView.sql",
       "fromModuleVersion": "23.1.0"
+    },
+    {
+        "run":"after",
+        "snippetPath":"dropCallNumberNormalizationFunctions.sql",
+        "fromModuleVersion":"24.0.0"
```
julianladisch, line 1038: Can you remove this entry from schema.json:

```
    {
      "run": "before",
      "snippetPath": "createCallNumberNormalizationFunctions.sql",
      "fromModuleVersion": "19.2.0"
    },
```

Can you remove this file:
https://github.com/folio-org/mod-inventory-storage/blob/860fa8e43f132d1a18f449c74318469e11eef701/src/main/resources/templates/db_scripts/createCallNumberNormalizationFunctions.sql

src/test/java/org/folio/rest/api/AsyncMigrationTest.java
julianladisch: This test succeeds locally and on Jenkins:
https://jenkins-aws.indexdata.com/job/folio-org/job/mod-inventory-storage/job/FOLIO-3388-with-canCancelMigration/
https://github.com/folio-org/mod-inventory-storage/tree/FOLIO-3388-with-canCancelMigration

Please explain why the `@Ignore` is needed.

MikhailFokanov-ebsco: Thanks for noting.

src/main/resources/templates/db_scripts/schema.json
MikhailFokanov-ebsco: I just forgot to push the changes. =

PR #749, opened 2022-03-02
MODINVSTOR-876 Store Shelving Order in item (Dewey call numbers)

pfilippov-epam:
### Overview

Populate Shelving order (effectiveShelvingOrder) on the item record when the call number type is Dewey Decimal classification.

### Approach

- Populate effective shelving location for records that have call number type set to Dewey Decimal classification. The call number type can be stored either on holdings or item record level.
- Add new unit tests to check that Dewey Decimal numbers are processed by CallNumberUtils


MikhailFokanov (approved):
Looks good.

OleksiiKuzminov (approved):

PR #750, opened 2022-03-02
MODINVSTOR-861: fixing identified issues

K-Felk:
Fixing issues identified with MODINVSTOR-861.

I added a test for the deletion operation to verify that it correctly returns a 404
when the entry cannot be found.


descriptors/ModuleDescriptor-template.json
```diff
@@ -2608,6 +2608,7 @@
       "displayName": "inventory storage module - all permissions",
       "description": "Entire set of permissions needed to use the inventory storage module",
       "subPermissions": [
+        "inventory-storage.related-instances.collection.get",
```
marcjohnson-kint, line 2611: This is only a single permission. I thought that multiple new permissions were added for the new endpoints?

K-Felk: Yeah, I for some reason thought you were asking to only have the one permission placed here.  Just added them all.  also the related-instance-type endpoint permissions didn't seem to be here either, so added those as well.

marcjohnson-kint: Thanks

marcjohnson-kint (approved):

PR #751, opened 2022-03-09
MODINVSTOR-883 Migrate to use new api-lint and api-doc CI tools

OleksiiKuzminov:


Jenkinsfile
```diff
@@ -2,7 +2,13 @@
 
 buildMvn {
   publishModDescriptor = 'yes'
-  doApiDoc = 'yes'
+
+  doApiDoc = true
+  doApiLint = true
+  apiTypes = 'RAML'
+  apiDirectories = 'ramls'
+  apiExcludes = 'instance-storage.raml holdings-storage.raml'
```
marcjohnson-kint, line 10: Why are these RAML files excluded from the documentation?

OleksiiKuzminov: Cause they have errors during the linting. Im not an owner of these APIs and not sure that can fix it without breaking changes. I believe it's better to have an additional task for it.

marcjohnson-kint: > Cause they have errors during the linting. Im not an owner of these APIs and not sure that can fix it without breaking changes. I believe it's better to have an additional task for it.

That seems like a strange decision to me. By excluding them, it effectively hides the errors, meaning they are less likely to be resolved.

What are the errors that you consider would require breaking changes to resolve?

dcrossleyau (changes requested):
As i explained a few days ago on [MODINVSTOR-883](https://issues.folio.org/browse/MODINVSTOR-883)

> Regarding that, the git history shows that those parts of the RAMLs were deliberately commented-out. I locally commented-out the "body" key directly prior to those, and got a clean api-lint run.

There is a similar entry at instance-preceding-succeeding-titles.raml (200 with no "body:")

If there is a "body" then it needs a media-type.

Do not simply hide them from processing. This also means that their API documentation will not be generated.


Also you need to follow the [api-lint](https://dev.folio.org/guides/api-lint/#jenkinsfile-for-raml) doc:

> Note: This tool replaces the deprecated “lint-raml” (runLintRamlCop) facility. Do not use both.

So remove that property from Jenkinsfile.

dcrossleyau:
The `runLintRamlCop` property still needs to be removed from Jenkinfile as instructed earlier in this PR and in the [api-lint](https://dev.folio.org/guides/api-lint/#jenkinsfile) documentation.

> Note: This tool replaces the deprecated “lint-raml” (runLintRamlCop) facility. Do not use both.

MikhailFokanov (approved):
Looks good.

dcrossleyau (approved):

PR #752, opened 2022-03-14
MODINVSTOR-885: Instances can be saved with invalid statistical codes

K-Felk:
While there is a database trigger that checks the integrity of statistical
codes in item records, there seems to be no such trigger for checking the 
integrity of statistical codes in instance records.  I added triggers that will do
so.


ramls/instance.json
```diff
@@ -386,8 +386,7 @@
       "type": "array",
       "description": "List of statistical code IDs",
       "items": {
-        "type": "string",
-        "$ref": "uuid.json"
+        "type": "string"
```
marcjohnson-kint, line 389: Why has the UUID format constraint been removed from these IDs?

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1140,7 +1140,7 @@
     },
     {
       "run": "after",
-      "snippetPath": "itemStatisticalCodeReferenceCheckTrigger.sql",
+      "snippetPath": "StatisticalCodeReferenceCheckTrigger.sql",
       "fromModuleVersion": "22.1.0"
```
marcjohnson-kint, line 1144: We shouldn't really change existing upgrade scripts (I wish we used tooling like liquibase that better described this). It is misleading and can lead to confusion when folks try to interpret these scripts in the context of versions.

I imagine that these changes aren't really intended to be applied to version 22.1.0 and later (given that 23.0.0 was the last release) which is the impression this change gives.

Rather we should add a new upgrade script that makes the changes from this point onwards.

ramls/instance.json
marcjohnson-kint: > Because it does nothing. I can see from testing the constraint is not enforced in this context. 

I don't think that necessarily means we shouldn't give the client the indication that a UUID is expected, even if it isn't validated.

Is your thinking that this would be disingenuous and it should be removed (and the minor interface version changed)?

> I also noted that statisticalcodeid fields in the item record does not have it (again, presumably because it has no practical effect in that context).

It is more likely, in this case, that the API predates FOLIO adopting that in it's JSON schema. There are likely a lot of `id` properties that don't have this on it when they should.

I believe RMB does generate code to check it for individual properties. Unfortunately, array support is much less sophisticated.

K-Felk: > Is your thinking that this would be disingenuous and it should be removed (and the minor interface version > > > > >changed)?

More that I found having the constraint there when it isn't enforced is confusing when troubleshooting.  I spent a considerable amount of time perplexed as to why this constraint wasn't being enforced, running tests and tinkering with the syntax, trying to get it to apply.  I took me a while to work out that it doesn't seem to apply in specific kinds of arrays.  

If we're going to leave it here, the it should also be added to statisticalcodeId fields in other records.  Though, again, I think its presence implies it's enforceable here, which causes confusion.  It was the fact that it *wasn't* present in the same field in item records that tipped me off to look for enforcement elsewhere in the code.

K-Felk: > How come you concerned about the UUID pattern being checked during this work?

Because the reported issue here is that the field was accepting non-UUID values.  I found through testing that is was not just null values, it would accept *any* string value.  Which, to me, signaled that enforcement of UUID pattern wasn't happening here.  Thus, my first stop was the instance description to see if a constraint had just been omitted by mistake.  But it was there, so the question immediately became "why isn't this constraint being enforced, when it's enforced elsewhere?"

K-Felk: > Ok, then let's remove it (and update the minor interface version) or alternatively, add validation for it.

To make sure I'm clear, I'd need to change the new version to 23.2?  And I'd need to do that in the pomfile as well as the news, yes?

K-Felk: Ah, yes, I get it.

descriptors/ModuleDescriptor-template.json
```diff
@@ -153,7 +153,7 @@
     },
     {
       "id": "instance-storage",
-      "version": "8.0",
+      "version": "8.1",
```
marcjohnson-kint, line 156: Is this the interface that uses this schema? No bulk edits or other APIs that use it?

K-Felk: I ran a search for the file name on the entire repo-this is the only interface/raml that directly references instance.json.  There are other instance-related endpoints, but they reference separate instance representations (like inventory-view-instance.json), which I have not changed.  Other json files reference the file (like instances.json), but I don't think their interfaces/ramls need to change?

marcjohnson-kint: > I ran a search for the file name on the entire repo-this is the only interface/raml that directly references instance.json. There are other instance-related endpoints, but they reference separate instance representations (like inventory-view-instance.json), which I have not changed.

Thank you for doing that.

> Other json files reference the file (like instances.json), but I don't think their interfaces/ramls need to change?

I believe `instances.json` references `instance.json` meaning they would need to change.

K-Felk: So anything that references a json which references instance.json would need to increment?  Because that's a lot more ramls/interfaces.

marcjohnson-kint: > So anything that references a json which references instance.json would need to increment? Because that's a lot more ramls/interfaces.

Yes, because that means the change is included in the interface definition.

NEWS.md
```diff
@@ -1,6 +1,15 @@
 ## 23.1.0 IN-PROGRESS
 
 * Add related instances record API (MODINVSTOR-861)
+* Added integrity checks to statisticalCodeIds in instance records (MODINVSTOR-885)
```
marcjohnson-kint, line 4: I think the removal of the UUID constraint on statistical codes should be included in this list

PR #763, opened 2022-05-02
[MODINVSTOR-898] refactor related instance type id

wwelling:
Resolves https://issues.folio.org/browse/MODINVSTOR-898


mkuklis (approved):
👏

marcjohnson-kint (approved):
Personally, I think this should be a major version change for the interface, irrespective of whether the module has been released or not.

Given that it hasn't, and I'm aware folks disagree with me on this (and practically it will likely make little difference, only establish what I think are poor habits), I won't hold up this change if folks decide not to do it.

marcjohnson-kint:
> @marcjohnson-kint I normally would agree with you but since this was not released yet and it's not consumed by anything it's still in this gray area where we can pull it off without a major version bump. I don't see it as a bad habit and more of being flexible.

Sure. I think this is a side effect of FOLIO not separating interfaces from implementing modules.

I agree there isn't an explicit policy on this and thus I'm not going to hold up the work.

PR #764, opened 2022-05-03
MODINVSTOR-894 optimize create instance endpoint

okolawole-ebsco:
Original premise for this ticket was to only reduce database calls when retrieving HRID value. It appears that the database calls are a smaller fraction of the entire work done. More optimizations were made.

- Single call to retrieve hrid settings and hrid sequence value via UNION ALL select statement rather than database function
- Using a single thread safe date formatter rather than recreating a new one on each call to set status dates.
- Returning response to the api client before domain event publishing is complete.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-## 23.1.0 IN-PROGRESS
+d## 23.1.0 IN-PROGRESS
 
 * Add related instances record API (MODINVSTOR-861)
 * Added integrity checks to statisticalCodeIds in instance records (MODINVSTOR-885)
@@ -10,6 +10,7 @@
 * provides `instance-storage-batch 1.1`
 * provides `instance-storage-batch-sync 1.1`
 * provides `inventory-view 1.1`
+* create instance optimization (MODINVSTOR-894)
 
 
 ## 23.0.0 2022-02-22
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,10 @@
-d## 23.1.0 IN-PROGRESS
+## 23.1.0 IN-PROGRESS
 
 * Add related instances record API (MODINVSTOR-861)
 * Added integrity checks to statisticalCodeIds in instance records (MODINVSTOR-885)
 * Removed UUID contraint on statisticalCodeIds in instance Records (MODINVSTOR-885)
+* Combined calls to retrieve HRID settings and getting sequence values (MODINVSTOR-894)
+* Allow response to be returned to the api client without waiting for domain event publishing during instance creation (MODINVSTOR-894)
 * provides `item-storage-dereferenced 0.2`
 * provides `holdings-storage 5.1`
 * provides `holdings-storage-batch-sync 1.1`
@@ -10,7 +12,6 @@ d## 23.1.0 IN-PROGRESS
 * provides `instance-storage-batch 1.1`
 * provides `instance-storage-batch-sync 1.1`
 * provides `inventory-view 1.1`
-* create instance optimization (MODINVSTOR-894)
 
 
 ## 23.0.0 2022-02-22
--- a/src/main/java/org/folio/rest/support/HridManager.java
+++ b/src/main/java/org/folio/rest/support/HridManager.java
@@ -258,8 +258,6 @@ public class HridManager {
   /**
    * Optimized version of getNextHrid that makes a single call to retrieve
    * hridsettings and the next sequence value for the specified inventory type.
-   * @param type
-   * @return
    */
   private Future<String> getNextHrid(InventoryType type) {
     final String sql = "select jsonb::TEXT from hrid_settings " +
@@ -288,8 +286,14 @@ public class HridManager {
           new ArrayTuple(1).addString(sequenceName),
           reply -> {
             var result = reply.result();
-            if (reply.failed() || result.size() != 2) {
+            if (reply.failed()) {
               fail(promise, "Failed to get hridsettings and next sequence value from the database", reply.cause());
+              return;
+            }
+            if (result.size() != 2) {
+              String errorMessage = "Result set contains " + result.size() + " items instead of 2 items";
+              fail(promise, errorMessage,
+                new IllegalStateException(errorMessage));
               return;
             }
             RowIterator<Row> iterator = result.iterator();
--- a/src/main/java/org/folio/rest/support/HridManager.java
+++ b/src/main/java/org/folio/rest/support/HridManager.java
@@ -8,11 +8,15 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Promise;
+import io.vertx.core.impl.future.FailedFuture;
 import io.vertx.core.json.Json;
 import io.vertx.core.json.JsonObject;
 import io.vertx.sqlclient.Row;
+import io.vertx.sqlclient.RowIterator;
 import io.vertx.sqlclient.RowSet;
 import java.util.List;
+
+import io.vertx.sqlclient.impl.ArrayTuple;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.folio.rest.jaxrs.model.HoldingsRecord;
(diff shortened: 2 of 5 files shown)


NEWS.md
```diff
@@ -1,4 +1,4 @@
-## 23.1.0 IN-PROGRESS
+d## 23.1.0 IN-PROGRESS
```
marcjohnson-kint, line 1: I think this might be a typo

NEWS.md
```diff
@@ -10,6 +10,7 @@
 * provides `instance-storage-batch 1.1`
 * provides `instance-storage-batch-sync 1.1`
 * provides `inventory-view 1.1`
+* create instance optimization (MODINVSTOR-894)
```
marcjohnson-kint, line 13: Please put this above the `provides` entries (which go last) and capitalize similarly to other new entries.

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -64,7 +64,11 @@ public Future<Response> createInstance(Instance entity) {
           InstanceStorage.PostInstanceStorageInstancesResponse.class, postResponse);
 
         return postResponse.future()
-          .compose(domainEventPublisher.publishCreated());
+          .compose(response -> {
+            // return response without waiting for event publish
+            domainEventPublisher.publishCreated().apply(response);
+            return Future.succeededFuture(response);
+          });
```
marcjohnson-kint, line 71: I thought the scope of this work was to change the database operations relating to HRIDs.

Why is this change also included in this pull request?

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
```diff
@@ -1,11 +1,11 @@
...
+import java.time.format.DateTimeFormatter;
 
 public final class StatusUpdatedDateGenerator {
+    private final static DateTimeFormatter df = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
     public static String generateStatusUpdatedDate() {
-      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
-      return sdf.format(new Date());
-    } 
+      return df.format(ZonedDateTime.now());
+    }
```
marcjohnson-kint, line 10: Why has this been changed?

Is there a significant computation saving from the static field (I would think the compiler would have already optimised this away)?

src/main/java/org/folio/rest/support/HridManager.java
```diff
@@ -246,6 +255,68 @@ private Future<String> getNextHrid(final HridSettings hridSettings, InventoryTyp
         Objects.toString(hridPrefix, ""), sequence.getLong(0)));
   }
 
+  /**
+   * Optimized version of getNextHrid that makes a single call to retrieve
+   * hridsettings and the next sequence value for the specified inventory type.
+   */
+  private Future<String> getNextHrid(InventoryType type) {
```
marcjohnson-kint, line 262: Is the old getNextHrid method still used now this one has been introduced?

src/main/java/org/folio/services/instance/InstanceService.java
okolawole-ebsco: After initial changes were made relating to the db and profiling the application, I saw that there was still room for improvement. The reason for the story is to optimize this hot spot for data import. I can push this out to another story if you feel strongly.

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
okolawole-ebsco: During my analysis of memory allocations during my test runs, large percentage of allocations were from  SimpleDateFormat instantiated by this method. This change is meant to reduce these allocations by using a single object that format dates. This reduces GC clean up over time(during multi hour data import jobs), I did not notice substantial change in response times of a single request.

src/main/java/org/folio/rest/support/HridManager.java
okolawole-ebsco: the old getNextHrid method is not used anymore. I am very open to taking them out.

NEWS.md
okolawole-ebsco: updated.

NEWS.md
okolawole-ebsco: thanks for catching this.

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
okolawole-ebsco: Most computation in code written by us is used by regex pattern matching for publication dates and spawning the new dates for status update field on the instance.

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/rest/support/HridManager.java
marcjohnson-kint: Yes, please remove a now unused method.

src/main/java/org/folio/services/instance/InstanceService.java
marcjohnson-kint: > After initial changes were made relating to the db and profiling the application, I saw that there was still room for improvement. The reason for the story is to optimize this hot spot for data import.

Thank you for explaining that.

> I can push this out to another story if you feel strongly.

There is no need to split it into another issue. I only wanted to understand why this change was made as well.

I think it could be worth expanding the comment to explain why the message is published asynchronously.

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
marcjohnson-kint: > During my analysis of memory allocations during my test runs, large percentage of allocations were from SimpleDateFormat instantiated by this method.

That is useful to know, thank you.

> This change is meant to reduce these allocations by using a single object that format dates. This reduces GC clean up over time(during multi hour data import jobs), I did not notice substantial change in response times of a single request.

Thanks, that makes sense.

src/main/java/org/folio/services/instance/InstanceService.java
okolawole-ebsco: expand my comment.

src/main/java/org/folio/rest/support/HridManager.java
okolawole-ebsco: methods removed

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -64,7 +64,16 @@ public Future<Response> createInstance(Instance entity) {
...
+          .compose(response -> {
+            // Return the response without waiting for a domain event publish
+            // to complete. Units of work performed by this service is the same
+            // but the ordering of the units of work provides a benefit to the
+            // api client invoking this endpoint. The response is returned
+            // a little earlier so the api client can continue its processing
+            // while the domain event publish is satisfied.
+            domainEventPublisher.publishCreated().apply(response);
+            return Future.succeededFuture(response);
+          });
```
julianladisch, line 76: A more idiomatic way is using

    .onSuccess(response -> {

so that no `return` is needed.

okolawole-ebsco: Thank you. I will keep this in mind next time!

src/main/java/org/folio/rest/support/StatusUpdatedDateGenerator.java
```diff
@@ -1,11 +1,12 @@
...
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 
 public final class StatusUpdatedDateGenerator {
+    private final static DateTimeFormatter df = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
     public static String generateStatusUpdatedDate() {
-      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSSZ");
-      return sdf.format(new Date());
-    } 
+      return df.format(ZonedDateTime.now(ZoneOffset.UTC));
```
okolawole-ebsco, line 10: @marcjohnson-kint I had to add the UTC offset due to issues I was seeing in another module. I ran some tests prior to verify that the dates generated by the most recent change was the same as the previous version. With this change, I am no longer leaving the timezone to change.

marcjohnson-kint: > @marcjohnson-kint I had to add the UTC offset due to issues I was seeing in another module. I ran some tests prior to verify that the dates generated by the most recent change was the same as the previous version. 

Ok. Whilst I think it _should_ be unnecessary to do this, it is the polite thing to do (and some parts of FOLIO expect it)

> With this change, I am no longer leaving the timezone to change.

Do you mean chance?

PR #766, opened 2022-05-05
MODINVSTOR-896 add GET for /inventory-storage/migrations/jobs

OleksiiKuzminov:


pfilippov-epam (approved):

ramls/migration/migration-job-collection-response.json
```diff
@@ -0,0 +1,15 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "description": "Get collection of async jobs",
+  "type": "object",
+  "properties": {
+    "jobs": {
+      "description": "List of async jobs records",
+      "javaType": "java.util.List<org.folio.rest.jaxrs.model.AsyncMigrationJob>"
```
hjiebsco, line 8: Curious why it is not a reference to the individual `job` JSON schema?

descriptors/ModuleDescriptor-template.json
```diff
@@ -1367,6 +1367,11 @@
           "pathPattern": "/inventory-storage/migrations/jobs/{id}",
           "permissionsRequired": ["inventory-storage.migration.job.item.get"]
         },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/inventory-storage/migrations/jobs",
+          "permissionsRequired": ["inventory-storage.migration.job.item.get"]
```
hjiebsco, line 1373: Is it a general practice to share permission between `collection` and `item`?

src/main/java/org/folio/services/migration/async/AsyncMigrationJobService.java
```diff
@@ -44,6 +46,13 @@ public AsyncMigrationJobService(Context vertxContext, Map<String, String> okapiH
     this.migrationContext = new AsyncMigrationContext(vertxContext, okapiHeaders, postgresClient);
   }
 
+  public Future<AsyncMigrationJobCollection> getAllAsyncJobs() {
+    return migrationJobRepository.get(new Criterion())
+      .map(list -> new AsyncMigrationJobCollection()
```
hjiebsco, line 51: Since this returns all jobs, curious if there is some process to clean up old jobs periodically or it replies on the hosting to manage that?

hjiebsco (commented):
Just some minor questions. Thanks.

ramls/migration/migration-job-collection-response.json
OleksiiKuzminov: Unfortunately, Rmb creates a separate class for it

descriptors/ModuleDescriptor-template.json
OleksiiKuzminov: No, but for admin API it makes sense. Too many unnecessary permissions can slow down the current login approach on the folio. It already affects karate tests, so I decided to avoid creating new permission for the same retrieve operations in the admin API

src/main/java/org/folio/services/migration/async/AsyncMigrationJobService.java
OleksiiKuzminov: For now, there are no mechanisms to clean up old jobs. If it's necessary we can create a ticket for it.

PR #772, opened 2022-05-19
MODINVSTOR-911: Send Kafka message after returning API response

julianladisch:
When changing data domain events get published via Kafka.

After the domain event has been published mod-inventory-storage sends the response to its client.

This delay is not needed.

mod-inventory-storage should respond to its client and then publish the domain event to Kafka.

Performance gains:

From 25 ms to 18 ms for POST /item-storage/items API response time. (Caused by fetching item and holding to calculate the domain event.)

From 19 ms to 18 ms for POST /holdings-storage/holdings response time.


adamdickmeiss (approved):
1: Eventual consistency for this.. I guess, anyway. as ElasticSearch takes time anyway .. 
2: no error reporting (if sending messages goes wrong, that is not reported).

okolawole-ebsco (approved):

PR #773, opened 2022-05-20
MODINVSTOR-908: Single HRID SQL query for batch (instance/holding/item)

julianladisch:
A batch of 10 instances currently requires 10 SQL queries, each fetching
a single hrid.

Combining them into one SQL is faster:

Tested with the POST /instance-storage/batch/synchronous API and
a batch size of 10 instances.

The API response time on a system without load reduces
from 88 ms to 75 ms, this is -15 %.

The throughput on a system under load (many parallel requests to
this API) increases from 46 requests/second to 64 request/second
(= from 460 instance/second to 640 instances/second),
this is +39 %.

### DIFF
--- a/src/main/java/org/folio/rest/support/HridManager.java
+++ b/src/main/java/org/folio/rest/support/HridManager.java
@@ -1,7 +1,5 @@
 package org.folio.rest.support;
 
-import static io.vertx.core.CompositeFuture.all;
-import static java.util.stream.Collectors.toList;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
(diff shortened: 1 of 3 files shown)


adamdickmeiss (approved):
OK. The best optimization would be to get rid of HRIDs altogether. They are evil and much worse than simply UUIDs.

okolawole-ebsco (approved):

julianladisch:
Smart library unions like GBV provide an hrid when creating the record. Then mod-inventory-storage skips the hrid creation step.

PR #775, opened 2022-05-24
MODINVSTOR-576, MODINVSTOR-901: Correctly implement delete by CQL

julianladisch:
* Correctly implement delete by CQL for instance/holding/item (MODINVSTOR-576)
* Reject DELETE instance/holding/item without CQL query (MODINVSTOR-901)

instanceDeleteCascade.sql automatically deletes connected marc source records
to avoid additional database queries.

For the domain event (Kafka) messages the deleted record is needed.
To avoid additional queries the delete query returns the deleted record and
the instance id it belongs to.

To avoid deserializing and serializing from/to JsonObject and/or POJO the
database returns a Json String (= raw) that is directly used for the domain
event (Kafka) message. DomainEventRaw takes raw Json Strings whereas
existing DomainEvent only takes POJOs.


descriptors/ModuleDescriptor-template.json
```diff
@@ -70,7 +70,7 @@
     },
     {
       "id": "item-storage",
-      "version": "9.0",
+      "version": "9.1",
```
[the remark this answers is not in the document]
julianladisch: The old code allows to delete all records. It was a mistake that the old code allows to omit the query parameter. This bug causes serious issues: https://issues.folio.org/browse/MODINVSTOR-901
Therefore it is no longer allowed to omit the query parameter. This is a bug fix.

New feature: Also allow to delete only selected records using query with CQL: https://issues.folio.org/browse/MODINVSTOR-576

src/main/java/org/folio/rest/support/CqlUtil.java
```diff
@@ -0,0 +1,22 @@
+package org.folio.rest.support;
+
+import java.util.regex.Pattern;
+
+public final class CqlUtil {
+  private static final Pattern CQL_MATCHES_ALL = Pattern.compile(
+      "^ *id *==? *\\* *$|^ *id *==? *\"\\*\" *$|^ *cql.allRecords *= *1 *$");
```
adamdickmeiss, line 7: Quite ugly to parse CQL that way. It would be better to run the parser and see if returns TRUE (the CQL parser of folio-vertx-lib simply returns null (to indicate no WHERE clause). But since the result of this is for optimization , I guess it doesn't matter.

adamdickmeiss (commented):
I don't know enough about this module anymore, so I hope others will chip in.

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -110,27 +113,53 @@ public Future<Response> updateInstance(String id, Instance newInstance) {
...
-    return instanceRepository.getById(id)
-      .compose(CommonValidators::refuseIfNotFound)
-      // before deleting the instance record delete its source marc record (foreign key!)
-      .compose(instance -> marcRepository.deleteById(id).map(notUsed -> instance))
-      .compose(instance -> {
-        final Promise<Response> deleteResult = promise();
-
-        deleteById(INSTANCE_TABLE, id, okapiHeaders, vertxContext,
-          InstanceStorage.DeleteInstanceStorageInstancesByInstanceIdResponse.class, deleteResult);
+    return instanceRepository.delete("id==" + id)
```
kaladay, line 130: Should this also check that id is safe?
That is, is it possible to inject CQL here via the `id` variable?
`String id = "c1726a23-2e7c-47c6-a278-0b177d8db9b5) or (id==*)";`

src/test/java/org/folio/rest/support/CqlUtilTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import org.folio.rest.testing.UtilityClassTester;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+@RunWith(JUnitParamsRunner.class)
+
+public class CqlUtilTest {
```
kaladay, line 12: I'm used to seeing the Annotation next to the Class rather than next to the Import.
I recommend swapping lines 10 and 11.

src/main/java/org/folio/services/domainevent/ItemDomainEventPublisher.java
```diff
@@ -52,6 +52,12 @@ public Future<Void> publishUpdated(HoldingsRecord hr, List<Item> oldItems) {
       .compose(domainEventService::publishRecordsUpdated);
   }
 
+  @Override
+  public void publishRemoved(String instanceId, String itemRaw) {
+    String instanceIdAndItemRaw = "{\"instanceId\":\"" + instanceId + "\"," + itemRaw.substring(1);
```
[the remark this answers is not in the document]
julianladisch: If itemRaw is

    {"id": "123","barcode": "987"}

then instanceIdAndItemRaw is

    {"instanceId":"567","id": "123","barcode": "987"}

The closing brace of itemRaw is copied and becomes the closing brace of instanceidAndItemRaw.

src/main/java/org/folio/services/instance/InstanceService.java
julianladisch: Good catch.
CQL encoding the id string will prevent this.

src/test/java/org/folio/rest/support/CqlUtilTest.java
julianladisch: Fixed.

descriptors/ModuleDescriptor-template.json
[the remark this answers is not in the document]
julianladisch: Agreed, major version bumped.

kaladay (approved):

src/main/resources/templates/db_scripts/instanceDeleteCascade.sql
```diff
@@ -0,0 +1,10 @@
+-- When deleting an instance automatically delete the connected marc record.
+-- Approach: foreign key "ON DELETE CASCADE"
+
+ALTER TABLE ${myuniversity}_${mymodule}.instance_source_marc
+  DROP CONSTRAINT IF EXISTS instance_source_marc_id_fkey;
+
+ALTER TABLE ${myuniversity}_${mymodule}.instance_source_marc
+  ADD CONSTRAINT instance_source_marc_id_fkey
+  FOREIGN KEY (id) REFERENCES instance(id) ON DELETE CASCADE;
```
marcjohnson-kint, line 9: I think it's interesting that active changes are being made in this area.

No one should be using MARC records stored directly in inventory and this API should've been removed a long time ago.

How has it been decided to cascade the deletion rather than fail deletion when a MARC record is present?

src/main/java/org/folio/services/kafka/InventoryProducerRecordBuilder.java
```diff
@@ -51,7 +51,7 @@ public InventoryProducerRecordBuilder propagateOkapiHeaders(Map<String, String>
 
   public KafkaProducerRecord<String, String> build() {
     try {
-      var valueAsString = getMapper().writeValueAsString(this.value);
+      var valueAsString = value instanceof String ? (String) value : getMapper().writeValueAsString(value);
```
marcjohnson-kint, line 54: Is there a better way to handle this?

I'm surprised that there isn't a mapper that writes a string as a string that can avoid this special case workaround.

marcjohnson-kint (commented):
I think the timing of this is change is unfortunate, coming only 3 days before the feature freeze for 2022 R2.

I would rather significant (and especially breaking changes) weren't made this close to the release period. 

It increases the chances of delaying the module release and that issues will be discovered in the BugFest period rather than in testing activities throughout the development cycle.

If this change goes ahead please:
* make sure there are changes in all dependent modules for supporting these new interfaces before merging
* (given the changes to message publishing) contact all of the teams that maintain modules that use the Kafka messages to test that their modules still work with those messages.

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -166,7 +166,7 @@ static void deleteAll(URL rootUrl) {
     CompletableFuture<Response> deleteAllFinished = new CompletableFuture<>();
 
     try {
-      client.delete(rootUrl, TENANT_ID,
+      client.delete(rootUrl + "?query=id==*", TENANT_ID,
```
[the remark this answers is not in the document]
julianladisch: This is to test all possibilities.

Changed to cql.allRecords=1 as requested.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1930,6 +1931,47 @@ public void canDeleteAllItems()
...
+    var response = client.delete(itemsStorageUrl(query), StorageTestSuite.TENANT_ID).get(5, SECONDS);
+    assertThat(response.getBody(), is("Expected CQL but query parameter is empty"));
+    assertThat(response.getStatusCode(), is(400));
+  }
+
+  @Test
+  public void cannotDeleteItemsWithInvalidCql() throws Exception {
+    var response = client.delete(itemsStorageUrl("?query=\""), StorageTestSuite.TENANT_ID).get(5, SECONDS);
+    assertThat(response.getBody(), containsStringIgnoringCase("parse"));
+    assertThat(response.getStatusCode(), is(400));
```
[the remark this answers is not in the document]
julianladisch: Fixed.

src/test/java/org/folio/rest/api/InstanceStorageTest.java
```diff
@@ -440,6 +440,49 @@ public void canDeleteAnInstance() throws InterruptedException, TimeoutException,
     assertRemoveEventForInstance(createdInstance.getJson());
   }
 
+  @Test
+  public void cannotDeleteNonexistingInstance() throws Exception {
```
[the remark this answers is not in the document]
julianladisch: Changed.

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -559,6 +559,51 @@ public void canDeleteAllHoldings() {
...
+    assertNotExists(h1);
+    assertNotExists(h3);
+    assertNotExists(h5);
+    assertRemoveEventForHolding(h1);
+    assertRemoveEventForHolding(h3);
+    assertRemoveEventForHolding(h5);
+  }
+
+  @Test
+  public void cannotDeleteHoldingsWithEmptyCql() throws Exception {
```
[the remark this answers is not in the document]
julianladisch: Replaced by SneakyThrows

src/test/java/org/folio/rest/api/HoldingsStorageTest.java
```diff
@@ -559,6 +559,51 @@ public void canDeleteAllHoldings() {
...
+  @Test
+  public void canDeleteHoldingsByCql() throws Exception {
+    UUID instanceId1 = UUID.randomUUID();
+    UUID instanceId2 = UUID.randomUUID();
+    instancesClient.create(smallAngryPlanet(instanceId1));
+    instancesClient.create(nod(instanceId2));
+
+    var h1 = holdingsClient.create(new HoldingRequestBuilder()
+        .forInstance(instanceId1).withPermanentLocation(mainLibraryLocationId).withHrid("1234")
+        ).getJson();
```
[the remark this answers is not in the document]
julianladisch: I disagree, but changed it to get this done.

src/main/java/org/folio/services/kafka/InventoryProducerRecordBuilder.java
julianladisch: I couldn't find such a mapper.

src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -159,10 +161,30 @@ public Future<Response> deleteItem(String itemId) {
       });
   }
 
-  public Future<Void> deleteAllItems() {
+  // suppress "Remove useless curly braces around lambda containing only one statement"
+  @SuppressWarnings("java:S1602")
```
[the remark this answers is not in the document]
julianladisch: Because multiple nested lambdas and expressions are more readable when curly braces show code stand on its own.

Changed to get this done.

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -110,27 +114,53 @@ public Future<Response> updateInstance(String id, Instance newInstance) {
...
+          return Response.noContent().build();
+        });
+  }
 
-        return deleteResult.future()
-          .onSuccess(domainEventPublisher.publishRemoved(instance));
-      });
+  /**
+   * Delete instances and connected marc records.
+   */
```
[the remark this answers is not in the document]
julianladisch: The code also deletes the MARC record itself. This is the information it is intended to provide.

NEWS.md
```diff
@@ -1,14 +1,17 @@
-## 23.1.0 IN-PROGRESS
+## 24.0.0 IN-PROGRESS
```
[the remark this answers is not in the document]
julianladisch: Done.

src/main/resources/templates/db_scripts/instanceDeleteCascade.sql
julianladisch: You should have asked this question before you merged the pull request four years ago:
https://github.com/folio-org/mod-inventory-storage/pull/126/files#diff-dbf6259d56a78ce52a3e91870b73432e7ac83f8b06f58a22135feb02fb17a9bdR349

src/test/java/org/folio/rest/api/InstanceDomainEventTest.java
```diff
@@ -84,6 +84,6 @@ public void eventIsNotSentWhenRemoveAllFailed() {
 
     final Response removeResponse = instancesClient.attemptDeleteAll();
 
-    assertThat(removeResponse.getStatusCode(), is(500));
+    assertThat(removeResponse.getStatusCode(), is(400));
```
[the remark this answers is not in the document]
julianladisch: All four test methods in this class check a failure caused by a foreign key violation:
eventIsNotSentWhenUpdateFailed()
eventIsNotSentWhenCreateFailed()
eventIsNotSentWhenRemoveFailed()
eventIsNotSentWhenRemoveAllFailed()
They all expect a 400 HTTP status code.
Can you give an example how "Failed" could be reworded to better reflect the expected failure?

src/main/java/org/folio/rest/support/CqlQuery.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.regex.Pattern;
+
+public final class CqlQuery {
+  private static final Pattern CQL_MATCHES_ALL = Pattern.compile(
+      "^ *id *==? *\\* *$"              // id=*
+      + "|"
+      + "^ *id *==? *\"\\*\" *$"        // id="*"
+      + "|"
+      + "^ *cql.allRecords *= *1 *$");  // cql.allRecords=1
```
adamdickmeiss, line 11: Any value for cql.allRecords should match.

src/main/java/org/folio/rest/support/CqlQuery.java
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * True if {@code cqlQuery} is a CQL expression that is known to always match all records, for
+   * example {@code id==*} or {@code cql.allRecords=1}.
+   * If false is returned the expression may still match all records because this
+   * method covers only a few cases and doesn't consider the existing data.
+   */
+  public boolean isMatchingAll() {
+    return CQL_MATCHES_ALL.matcher(cql).find();
+  }
+}
```
adamdickmeiss, line 28: I don't like this. Parsing CQL (and SQL) by regular expressions is just a bad idea. I would have preferring letting the CQL2PgJSON parser do the work and if it evaluates to `true` or `TRUE`, it would match. Admitted, that would not work with id= searches.. But that's a very lame way of getting all records anyway.

For example dealing with `(cql.allRecords = ok)` as well.

adamdickmeiss (commented):
I don't have enough knowledge to judge either way about this.

src/main/java/org/folio/rest/support/CqlQuery.java
marcjohnson-kint: > I don't like this. Parsing CQL (and SQL) by regular expressions is just a bad idea. I would have preferring letting the CQL2PgJSON parser do the work and if it evaluates to `true` or `TRUE`, it would match. 

I agree that would be better.

> Admitted, that would not work with id= searches.. But that's a very lame way of getting all records anyway.

I completely agree with this. I don't think we should try to be clever about figuring out if a query matches all records.

julianladisch: The regex pattern matching has been replaced with CQLParser.

If you want the single DELETE_ALL Kafka event been replaced by potentially thousands of single record DELETE Kafka events please create a Jira and I will be handle it in a separate pull request. Replacing DELETE_ALL by multiple DELETEs will be a performance regression that requires approval by the sysops.

src/main/resources/templates/db_scripts/instanceDeleteCascade.sql
julianladisch: I couldn't find a Jira that approves the removal of the MARC records APIs.

src/test/java/org/folio/rest/api/InstanceDomainEventTest.java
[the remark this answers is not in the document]
julianladisch: Resolving because the intention of this test is to demonstrate a foreign key violation, and not an internal failure.

src/main/java/org/folio/rest/support/CqlQuery.java
marcjohnson-kint: > Replacing DELETE_ALL by multiple DELETEs will be a performance regression that requires approval by the sysops.

How would this be a regression? As I understand it, the semantically explicit `cql.allRecords` publishes the single message.

Have I misunderstood?

src/main/resources/templates/db_scripts/instanceDeleteCascade.sql
marcjohnson-kint: There may well not be one. 

I would hope it was obvious that it has been superseded by mod-source-record-storage.

I thought those APIs were informally deprecated a long time ago.

src/main/java/org/folio/rest/support/CqlQuery.java
julianladisch: You wrote:

> I don't think we should try to be clever about figuring out if a query matches all records.

If we don't figure out whether the query matches all records (for example `cql.allRecords`) we cannot call the fast deleteAllInstances() and always need to call the slow delete(String cql) method. This is a performance regression.

marcjohnson-kint: > You wrote:
> 
> > I don't think we should try to be clever about figuring out if a query matches all records.
> 
> If we don't figure out whether the query matches all records (for example `cql.allRecords`) we cannot call the fast deleteAllInstances() and always need to call the slow delete(String cql) method. This is a performance regression.

I believe in my other [comment|https://github.com/folio-org/mod-inventory-storage/pull/775#discussion_r886650100] I suggested that I thought that a dedicated message was acceptable for the case where the intent was obvious and explicit (cql.allRecords).

My objection was not to that optimisation when the intention was obvious, it was with the inference of intent with other queries that happened to delete all records.

marcjohnson-kint: To be very explicit, I think treating cql.allRecords as an explicit way to deliberately delete all records is a good idea and in that case only a specific message could be published.

I don't think that any special provision should be taken for other queries that only happen to delete all records. 

It seems that code is still attempting to do that.

marcjohnson-kint: I wouldn't consider accommodating only the explicit intent of `cql.allRecords` a regression. This is the replacement for the current API.

I do think the code attempting to infer intent (listed below) about other queries should be removed. I would not consider that a regression, because that functionality did not exist previously.

```
    return "id".equalsIgnoreCase(node.getIndex())
        && StringUtils.equalsAny(base, "=", "==")
        && "*".equals(node.getTerm());
```

julianladisch: Since RMB v30.2.3 the section ["CQL: Matching all records" of the RMB documentation](https://github.com/folio-org/raml-module-builder#cql-matching-all-records) declares `id=*` to be officially supported:

> A search matching all records in the target index can be executed with a cql.allRecords=1 (CQL standard) or a id=* (RMB specific) query.

Should we add a restriction to the RMB documentation and the mod-inventory-storage documentation and the Nolana release notes saying that mod-inventory-storage no longer has full support for `id=*` in all cases?

marcjohnson-kint: > Should we add a restriction to the RMB documentation and the mod-inventory-storage documentation and the Nolana release notes saying that mod-inventory-storage no longer has full support for id=* in all cases?

I think @adamdickmeiss summed that up pretty well:

> Admitted, that would not work with id= searches.. But that's a very lame way of getting all records anyway.

I think it's poor guidance to recommend that folks use a query like `id=*` for matching all records, especially when there is a specific provision for an intent revealing clause in the standard.

I ambivalent about whether mod-inventory-storage allows that CQL or not. My concern here is with implementing specific additional behaviour based upon it. I already think that the special message is a compromise.

I don't think this decision needs to be made for all cases, rather the case that we are exploring here, which is a specific behaviour for deletion only.

There are lots of CQL queries that folks can write that don't make much sense. That is one of the major trade offs with using an open query language rather than defined criteria.

If you want to include documentation to describe how folks should avoid using `id=*` etc for matching all records and for deletion specifically, I would support that endeavour.

steveellis (approved):

PR #786, opened 2022-06-10
MODINVSTOR-915: Upgrade RMB to 34.0.0

K-Felk:


src/test/java/org/folio/rest/api/AsyncMigrationTest.java
```diff
@@ -31,6 +31,7 @@
 
 import static io.vertx.core.Future.succeededFuture;
 import static java.util.concurrent.TimeUnit.SECONDS;
+import org.apache.commons.collections4.map.CaseInsensitiveMap;
```
marcjohnson-kint, line 34: This is a strange location for a new non-static import.

Please use these rules:
* No use of wildcard import statements
* Static imports in alphabetic order first
* Then imports in defined then alphabetic order (java, javax, org, com, everything else) with a line break between defined groups

More info can be found in https://issues.folio.org/browse/FOLIO-1667

kaladay (approved):

PR #787, opened 2022-06-13
MODINVSTOR-923: Reject batch upsert changing HRID

julianladisch:
Changing the HRID is rejected in the single record
(instance, holding, item) PUT APIs.

However, the synchronous batch API allow it. This is a bug.

The batch API Java code doesn't fetch the records before running
the update query for performance reasons. Therefore we need to
check the HRID in the database.

The new trigger method hrid_lock() fails the update on HRID change.
It replaces HridValidators.java that is removed.

A custom PostgreSQL error code 23578 indicates the hrid lock violation.

Currently no other module relies on the error message:
https://github.com/search?q=org%3Afolio-org+%22hrid+field+cannot%22&type=code


src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1033,9 +1033,14 @@
       "fromModuleVersion": "23.1.0"
     },
     {
-        "run":"after",
-        "snippetPath":"dropCallNumberNormalizationFunctions.sql",
-        "fromModuleVersion":"24.0.0"
+      "run": "after",
+      "snippetPath": "hridLock.sql",
+      "fromModuleVersion": "23.1.0"
```
marcjohnson-kint, line 1038: This should be 24.0.0 now following other recent changes.
```suggestion
      "fromModuleVersion": "24.0.0"
```

src/main/resources/templates/db_scripts/hridLock.sql
```diff
@@ -0,0 +1,26 @@
...
+CREATE TRIGGER lock_hrid BEFORE UPDATE ON instance
+  FOR EACH ROW EXECUTE FUNCTION lock_hrid();
+
+DROP TRIGGER IF EXISTS lock_hrid ON holdings_record CASCADE;
+CREATE TRIGGER lock_hrid BEFORE UPDATE ON holdings_record
+  FOR EACH ROW EXECUTE FUNCTION lock_hrid();
+
+DROP TRIGGER IF EXISTS lock_hrid ON item CASCADE;
+CREATE TRIGGER lock_hrid BEFORE UPDATE ON item
+  FOR EACH ROW EXECUTE FUNCTION lock_hrid();
```
marcjohnson-kint, line 26: I think this is an unfortunate design decision, it means that business logic is either shifted or duplicated in the database, which increases the maintenance and testing burden of logic like this.

julianladisch: There is no duplication.

HridValidators.java is deleted and moved to hridLock.sql.

If HridValidators.java should be used these 6 APIs need to call it:

/instance-storage/instances/{instanceId}
/holdings-storage/holdings/{holdingsRecordId}
/item-storage/items/{itemId}
/instance-storage/batch/synchronous
/holdings-storage/batch/synchronous
/item-storage/batch/synchronous

If hridLock.sql is used only 3 triggers need to be enabled, they are automatically called.

If hridLock.sql is used the batch APIs can avoid a SELECT before running the upgrade. This gives a performance boost.

Please explain how this increases the maintenance and testing burden. Isn't it similar to foreign key and uniqueness checks that already exist for these tables?
Please explain why the maintenance and testing burden overrules the performance needs that libraries have requested for the 3 batch APIs.

src/main/resources/templates/db_scripts/schema.json
julianladisch: Changed.

adamdickmeiss (commented):
I'm not competent to review this. Moreover, the HRIDs is the single biggest mistake in FOLIO. A global single point of uniqueness that is much worse than UUIDs. Not only are UUIDs unique and can be created across the board, they make it impossible to move objects from one system to another without having to "re-calculate" HRIDs .. And finally, as soon as you move to 6 digits, a UUID prefix is easier to "remember" than a HRIDs as it contains letters too, and thus can represent more items with fewer characters. This issue along with many others shows the burden HRIDs have on the system (and why UUIDs are much better).

src/main/resources/templates/db_scripts/hridLock.sql
marcjohnson-kint: > If hridLock.sql is used the batch APIs can avoid a SELECT before running the upgrade. This gives a performance boost.

I understand the motivation.

> Please explain how this increases the maintenance and testing burden. 

This increases the maintenance burden because it increases the need for a developer to refer to both the java and database code to understand the behaviour of the system.

It increases the testing burden because this behaviour can now _only_ be tested using the database. Especially the business logic for denying the change of a HRID which relies on the coordination between the java and database code which makes testing in isolation harder.

This becomes more relevant when we consider the increasingly likely future of the inventory modules being merged.

> Isn't it similar to foreign key and uniqueness checks that already exist for these tables? 

I consider constraints in the database as a back stop and should be used sparingly. My line for this is crossed when business logic (denying the changing of a HRID) is implemented in a trigger. I think SQL and triggers are a poor way of expressing that kind of logic.

> Please explain why the maintenance and testing burden overrules the performance needs that libraries have requested for the 3 batch APIs.

I didn't make that assertion. I said I considered this change to be an unfortunate trade off.

I think this is another example of how FOLIO tends to make convenient small design decisions that end up not fitting in with the broader view. My understanding is that what folks want from these batch APIs is consistent and sustainable performance, which I think is different to outright speed (which I think is the argument made for this work).

marcjohnson-kint:
This is a significant change to the implementation of multiple APIs. I think it's worth considering if a more intermediary fix could be put in place for the Morning Glory release, given those are intended to make minimal changes.

julianladisch:
This pull request if for master = Nolana.

skoczko:
@marcjohnson-kint @jeremythuff @K-Felk Any chance you could review the PR? Thanks!

marcjohnson-kint:
@skoczko 

> @marcjohnson-kint Any chance you could review the PR? Thanks!

What kind of review / feedback are you asking for? I have given my feedback on the design choices for this work already.

marcjohnson-kint:
I'm closing this as the underlying [JIRA issue](https://issues.folio.org/browse/MODINVSTOR-923) is closed as won't do

PR #790, opened 2022-06-27
MODINVSTOR-927: InstallUpgradeIT

julianladisch:
Add an InstallUpgradeIT.java integration test. It should check the shaded fat uber jar and Dockerfile:

* Test /admin/health.
* Test that logging works.
* Test installation and migration with smoke test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,8 +18,7 @@ buildMvn {
     buildJavaDocker {
       publishPreview = false
       publishMaster = 'yes'
-      healthChk = 'yes'
-      healthChkCmd = 'curl -sS --fail -o /dev/null  http://localhost:8081/apidocs/ || exit 1'
+      // healthChk has been moved into InstallUpgradeIT.java
     }
   }
 }
--- a/pom.xml
+++ b/pom.xml
@@ -167,6 +167,18 @@
       <artifactId>kafka</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.rest-assured</groupId>
+      <artifactId>rest-assured</artifactId>
+      <version>5.1.1</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>  <!-- for testcontainers -->
+      <groupId>org.apache.logging.log4j</groupId>
+      <artifactId>log4j-slf4j-impl</artifactId>
+      <version>2.17.2</version>
+      <scope>test</scope>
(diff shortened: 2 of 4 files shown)


Jenkinsfile
```diff
@@ -18,8 +18,7 @@ buildMvn {
     buildJavaDocker {
       publishPreview = false
       publishMaster = 'yes'
-      healthChk = 'yes'
-      healthChkCmd = 'curl -sS --fail -o /dev/null  http://localhost:8081/apidocs/ || exit 1'
+      // healthChk has been moved into InstallUpgradeIT.java
```
marcjohnson-kint, line 21: What does this mean?

As I understand it, this configuration is about how the docker container build is configured, how is an integration test relevant to that configuration?

steveellis: @julianladisch I share Mark's question, but I think that removing this from the Jenkinsfile means is that docker is no longer responsible for testing the health check endpoint. Instead you have put it into the new test class which calls it using testcontainers. It also does two other things (logging test, and data migration test) which seem like great things to test this way.

How many other modules use this pattern for smoke type tests? Is this something we should consider implementing more widely?

steveellis (approved):

Jenkinsfile
julianladisch: healthChkCmd is a workaround for modules that doesn't check whether the Docker image actually works in the maven integration test. healthChkCmd is run in the Jenkins build pipeline. It checks whether the the Dockerfile created in the previous step of the build pipeline actually works. This check is performed by running healthChkCmd.

This pull request adds the integration test InstallUpgradeIT that exactly does this check: It creates the Docker image and executes it. This is clearly documented in the javadoc of InstallUpgradeIT:

> Check the shaded fat uber jar and Dockerfile

Therefore the healthChkCmd is no longer needed, it should be removed to save build pipeline resources.

marcjohnson-kint: > healthChkCmd is a workaround for modules that doesn't check whether the Docker image actually works in the maven integration test. healthChkCmd is run in the Jenkins build pipeline. It checks whether the the Dockerfile created in the previous step of the build pipeline actually works. This check is performed by running healthChkCmd.

I wasn't aware that this is a workaround.

> Therefore the healthChkCmd is no longer needed, it should be removed to save build pipeline resources.

Should all modules move to this approach?

Is this documented somewhere?

PR #799, opened 2022-08-01
MODINVSTOR-935 - Schema differences between MG Bugfest and clean MG deployment

rmathew1011:
[MODINVSTOR-935](https://issues.folio.org/browse/MODINVSTOR-935)


marcjohnson-kint (commented):
I don't think it's sufficient to only change the existing SQL files.

One of the reasons why there is a gap between the clean deployment and an upgrade is due to when these scripts are triggered.

I would expect that we would need to either change when (which version) these scripts were configured to run or to add additional scripts to add / remove these elements.

How were these changes tested? Was an upgrade from the Lotus version of mod-inventory-storage to the Morning Glory version of mod-inventory-storage simulated?

src/main/resources/templates/db_scripts/relatedInstanceType.sql
```diff
@@ -0,0 +1,53 @@
...
+    NEW.jsonb = jsonb_set(NEW.jsonb, '[metadata, createdDate]', to_jsonb(NEW.creation_date));
+    IF NEW.created_by IS NULL THEN
+      NEW.jsonb = NEW.jsonb #- '{metadata, createdByUserId}';
+    ELSE
+      NEW.jsonb = jsonb_set(NEW.jsonb,'{metadata, createdByUserId}' to_jsonb(NEW.created_by));
+    END IF;
+    RETURN NEW;
+  END;
+$$ LANGUAGE plpgsql COST 100
+GO
```
julianladisch, line 19: `GO` is a psql command, it is not available when running these db_scripts during migration.

src/main/resources/templates/db_scripts/relatedInstanceType.sql
```diff
@@ -0,0 +1,53 @@
+DROP FUNCTION IF EXISTS set_related_instance_type_md_json;
+
+CREATE FUNCTION set_related_instance_type_md_json ()
```
julianladisch, line 3: It's RMB's task to create and upgrade these functions. If there is a bug in RMB it should be fixed in RMB (because it affects other modules as well).

marcjohnson-kint: @julianladisch please can you assist with investigation into the report that those functions are not included in a clean installation of the module.

K-Felk (commented):
[Removed]

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1015,7 +1015,7 @@
     {
       "run": "after",
       "snippetPath": "statisticalCodeTypesReferenceCheckTrigger.sql",
-      "fromModuleVersion": "23.0.0"
+      "fromModuleVersion": "25.0.0"
```
marcjohnson-kint, line 1018: Yes, I think this is the change needed on the mainline (assuming the next version is 25.0.0 rather than 24.1.0).

A different version will be needed for the back port.

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1041,6 +1041,11 @@
       "run": "after",
       "snippetPath": "createInstanceSetView.sql",
       "fromModuleVersion": "25.0.0"
+    },
+    {
+      "run": "after",
+      "snippetPath": "removeRelatedInstanceTypeOperations.sql",
+      "fromModuleVersion": "25.0.0"
```
marcjohnson-kint, line 1048: Oh, is there another difference that I missed in my write up or is this the `set_md` differences that are the responsibility of RMB?

jeremythuff: No, I did this before I saw your write-up. Rincy will be removing these.

K-Felk (approved):

NEWS.md
```diff
@@ -1,6 +1,7 @@
 ## 25.0.0 IN-PROGRESS
 
 * Fixed effective location migration script for holdings records (MODINVSTOR-940)
+* Fixing mg bugfest and clean deployment schema difference ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
```
marcjohnson-kint, line 4: I would probably change this in a similar way to the feedback I gave on the back port pull request

PR #800, opened 2022-08-05
MODINVSTOR-918: GET Instance Set by CQL

julianladisch:
The implementation should be as general as possible and easy to extend
so that it can be used for many use cases.


julianladisch:
Replace by a new pull request.

PR #805, opened 2022-08-11
MODINVSTOR-940: back port to 24.0.2

K-Felk:
I asked for review on this because I've never released an operator script before.  My understanding from talking
to others is that we put the script in the wiki, and refer to it in the release notes.  I've put information about it in the newsfile here to be used in the release notes.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 ## 24.0.2 2022-08-10
 
 * Fixing when migration script to populate holdings effective locations runs ([MODINVSTOR-940] (https://issues.folio.org/browse/MODINVSTOR-940))
+* There is an operator script (on the wiki in inventory->scripts) that can be used to manually populate effectiveLocationId fields in holdings records for libraries who still have difficulties getting this field populated for existing records
 
 ## 24.0.1 2022-07-18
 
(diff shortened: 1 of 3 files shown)


K-Felk:
I've actually been trying to track down an example of what mentions in release notes look like for operator scripts, but cannot find any examples.  MODSOURMAN-627, just to pick an example, has an associated operator script.  But it's not mentioned in the release notes for the release the issue is associated with (3.3.0)

NEWS.md
```diff
@@ -1,3 +1,8 @@
+## 24.0.2 2022-08-10
+
+* Fixing when migration script to populate holdings effective locations runs ([MODINVSTOR-940] (https://issues.folio.org/browse/MODINVSTOR-940))
+* There is an operator script (on the wiki in inventory->scripts) that can be used to manually populate effectiveLocationId fields in holdings records for libraries who still have difficulties getting this field populated for existing records
```
marcjohnson-kint, line 4: I would not include any operator scripts in the release notes. By definition, they are not part of the module itself.

marcjohnson-kint:
> I've actually been trying to track down an example of what mentions in release notes look like for operator scripts, but cannot find any. MODSOURMAN-627, just to pick an example, has an associated operator script. But it's not mentioned in the release notes for the release the issue is associated with (3.3.0)

I would not include any operator scripts in the release notes. By definition, they are not part of the module itself.

They might be included in the release notes for a flower release on the wiki, as that is what operators tend to read.

marcjohnson-kint (approved):

NEWS.md
K-Felk: Ah, of course.  When she said release notes, I assumed she meant the ones that go out with a release version-this makes much more sense.  I'll amend, merge, and release.

PR #809, opened 2022-08-11
Fixing mb bugfest and clean deployment schema difference and version …

rmathew1011:
[MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
-## 24.0.4 2022-08-11
+## 24.0.4 IN-PROGRESS
 
-* Fixing mb bugfest and clean deployment schema difference version modified ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
+* Adds integrity checks for statistical code types during upgrade ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
 
 ## 24.0.2 2022-08-10
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 24.0.4 2022-08-11
+
+* Fixing mb bugfest and clean deployment schema difference version modified ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
+
 ## 24.0.2 2022-08-10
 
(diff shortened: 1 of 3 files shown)


NEWS.md
```diff
@@ -1,3 +1,7 @@
+## 24.0.4 2022-08-11
```
marcjohnson-kint, line 1: This version hasn't been released yet, so it should not have a release date

```suggestion
## 24.0.4 IN-PROGRESS
```

NEWS.md
```diff
@@ -1,3 +1,7 @@
...
+
+* Fixing mb bugfest and clean deployment schema difference version modified ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
```
marcjohnson-kint, line 3: Given that this is documentation for an operator or user of the module, I think it is better to use language that states what has happened.

```suggestion
* Adds integrity checks for statistical code types during upgrade ([MODINVSTOR-935] (https://issues.folio.org/browse/MODINVSTOR-935))
```

src/main/resources/templates/db_scripts/schema.json
```diff
@@ -1015,7 +1015,7 @@
     {
       "run": "after",
       "snippetPath": "statisticalCodeTypesReferenceCheckTrigger.sql",
-      "fromModuleVersion": "23.0.0"
+      "fromModuleVersion": "24.0.4"
```
marcjohnson-kint, line 1018: The next targeted release on the 24.0.x branch is 24.0.3. Why was 24.0.4 chosen for the version where the script should run?

NEWS.md
rmathew1011: Text modified

NEWS.md
rmathew1011: Description updated

src/main/resources/templates/db_scripts/schema.json
rmathew1011: updated to 24.0.3

marcjohnson-kint (approved):
This seems ok to me. It would be ideal if the upgrade (from the lotus version to this branch) could be tested locally to check it works as expected.

jeremythuff (approved):

PR #810, opened 2022-08-12
MODINVSTOR-389: Add Former and Alternative Alternative Title Types

K-Felk:
After looking at how this was done for MODINVSTOR-350 (Uniform Title) it looks like all that's
necessary on the back end is adding the two new title types to the reference data, which I have done.


marcjohnson-kint:
> After looking at how this was done for MODINVSTOR-350 (Uniform Title) it looks like all that's necessary on the back end is adding the two new title types to the reference data, which I have done.

This is sufficient for loading these records on a fresh install when the reference records flag is set. Most operators do not set the reference records flag on an upgrade, meaning they won't be loaded then.

I think we need to establish what behaviour is expected under those circumstances, especially given the issue has been determined urgent due to these records being needed in existing systems.

That doesn't need to stop the merging of this change.

marcjohnson-kint (commented):
Are there any tests for loading reference records? If so, could these be included in those tests?

marcjohnson-kint (approved):

K-Felk:
> Are there any tests for loading reference records? If so, could these be included in those tests?

There are tests for loading reference records, but none of them check to see if a specific record is loaded.  Rather, they check to see that the number of records loaded under a specific area falls within a min/max threshold.  Check ReferenceTablesTest.java to see this.

rmathew1011 (changes requested):
Changes requested for : reference-

1. data/alternative-title-types/AlternativeTitle.json
2.  reference-data/alternative-title-types/FormerTitle.json 

newline could be added at the end of these files

PR #811, opened 2022-08-16
MODINVSTOR-949: Speed up setEffectiveHoldingsLocation.sql migration

julianladisch:
The migration script setEffectiveHoldingsLocation.sql sets the
effective location of all holdings. It should skip all holdings
that already have the effective location been set.

The effective location has a database index because it is a
foreign key to the location table:
https://github.com/folio-org/mod-inventory-storage/blob/v24.0.2/src/main/resources/templates/db_scripts/schema.json#L585-L611

Note that permanentLocationId is a required field:
https://github.com/folio-org/mod-inventory-storage/blob/v24.0.2/ramls/holdingsrecord.json#L371

Unit tests already exist in
src/test/java/org/folio/rest/api/EffectiveLocationMigrationTest.java

### DIFF
--- a/src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
+++ b/src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
@@ -1,11 +1,4 @@
-START TRANSACTION;
-
 UPDATE ${myuniversity}_${mymodule}.holdings_record
-  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}', TO_JSONB(holdings_record.jsonb->>'temporaryLocationId'))
-WHERE jsonb->>'temporaryLocationId' IS NOT NULL;
-
-UPDATE ${myuniversity}_${mymodule}.holdings_record
-  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}', TO_JSONB(holdings_record.jsonb->>'permanentLocationId'))
-WHERE jsonb->>'temporaryLocationId' IS NULL;
-
-END TRANSACTION;
+  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}',
+                        TO_JSONB(COALESCE(temporarylocationid, permanentlocationid)))
+WHERE effectivelocationid IS NULL;


marcjohnson-kint:
@julianladisch Has there been any performance testing to determine if the changes significantly improve the performance over the original script?

src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
```diff
@@ -1,11 +1,4 @@
...
-
-UPDATE ${myuniversity}_${mymodule}.holdings_record
-  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}', TO_JSONB(holdings_record.jsonb->>'permanentLocationId'))
-WHERE jsonb->>'temporaryLocationId' IS NULL;
-
-END TRANSACTION;
\ No newline at end of file
+  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}',
+                        TO_JSONB(COALESCE(temporarylocationid, permanentlocationid)))
+WHERE effectivelocationid IS NULL;
```
marcjohnson-kint, line 4: Where does the value of this come from?

There does not seem to be any assignment (the case is different to the case of the only other usage).

src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
```diff
@@ -1,11 +1,4 @@
...
-WHERE jsonb->>'temporaryLocationId' IS NOT NULL;
-
-UPDATE ${myuniversity}_${mymodule}.holdings_record
-  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}', TO_JSONB(holdings_record.jsonb->>'permanentLocationId'))
-WHERE jsonb->>'temporaryLocationId' IS NULL;
-
-END TRANSACTION;
\ No newline at end of file
+  SET jsonb = JSONB_SET(holdings_record.jsonb, '{effectiveLocationId}',
+                        TO_JSONB(COALESCE(temporarylocationid, permanentlocationid)))
```
marcjohnson-kint, line 3: Where do these two values come from? They don't seem to have been retrieved from the JSONB (and don't share the case of these properties).

src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
julianladisch: https://github.com/folio-org/raml-module-builder#foreign-keys-constraint

For each foreign key RMB creates a trigger that automatically copies the foreign key value from the jsonb field into a dedicated field. This is needed to create the database index and to create the foreign key relationship constraint.

This applies to the three fields temporarylocationid, permanentlocationid, effectivelocationid.

julianladisch:
> Has there been any performance testing to determine if the changes significantly improve the performance over the original script?

The old code always runs a full table scan and always makes a full table write.

The new code skips all records that already have the effective location.

I've made a manual test that shows that the new `WHERE effectivelocationid IS NULL` uses the index.

julianladisch:
A manual test shows that all records are skipped when they have an effective location:

```
 Update on holdings_record  (cost=0.14..5.35 rows=1 width=222) (actual time=0.021..0.023 rows=0 loops=1)
   ->  Index Scan using holdings_record_effectivelocationid_idx on holdings_record  (cost=0.14..5.35 rows=1 width=222) (actual time=0.019..0.019 rows=0 loops=1)
         Index Cond: (effectivelocationid IS NULL)
 Planning Time: 0.203 ms
 Execution Time: 0.137 ms
```

src/main/resources/templates/db_scripts/setEffectiveHoldingsLocation.sql
marcjohnson-kint: Thank you, that was context I was missing.

K-Felk (approved):
I'll also copy these changes to the operator script.

marcjohnson-kint:
> I'll also copy these changes to the operator script.

@julianladisch Is the intention of this change to improve the performance of the upgrade when this script is triggered?

If so, and this is intended to address the concerns around Morning Glory, as I understand it, that would mean this change needs back porting and the version that the script is associated with needs changing.

How wrong is my understanding?

julianladisch:
https://issues.folio.org/browse/MODINVSTOR-949 "Speed up setEffectiveHoldingsLocation.sql migration" has the intention to improve the performance of the upgrade when this script is triggered. The intended mod-inventory-storage versions noted in the "Fix Version/s" field are 25.0.0 and 24.0.3, the intended release noted in the "Release" field is "Morning Glory R2 2022 Bug Fix".

marcjohnson-kint:
> https://issues.folio.org/browse/MODINVSTOR-949 "Speed up setEffectiveHoldingsLocation.sql migration" has the intention to improve the performance of the upgrade when this script is triggered. The intended mod-inventory-storage versions noted in the "Fix Version/s" field are 25.0.0 and 24.0.3, the intended release noted in the "Release" field is "Morning Glory R2 2022 Bug Fix".

@julianladisch  Will you be back porting these improvements to 24.0.4 (as 24.0.3 is already released without this change)?

marcjohnson-kint:
> https://issues.folio.org/browse/MODINVSTOR-949 "Speed up setEffectiveHoldingsLocation.sql migration" has the intention to improve the performance of the upgrade when this script is triggered. The intended mod-inventory-storage versions noted in the "Fix Version/s" field are 25.0.0 and 24.0.3, the intended release noted in the "Release" field is "Morning Glory R2 2022 Bug Fix".

Ok, thank you. You are correct, I should've checked the JIRA issue for this information.

PR #812, opened 2022-08-17
Release v24.0.3

Shans-Kaluhin:
Update NEWS

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+## 24.0.3 2022-08-17
+
+* Restrict cancellation of already finished reindex job ([MODINVSTOR-936] (https://issues.folio.org/browse/MODINVSTOR-936))
+* Remove shelfKey pattern check for call-number ([MODINVSTOR-942] (https://issues.folio.org/browse/MODINVSTOR-942))
+
 ## 24.0.2 2022-08-10
 
 * Fixing when migration script to populate holdings effective locations runs ([MODINVSTOR-940] (https://issues.folio.org/browse/MODINVSTOR-940))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 25.0.0 IN-PROGRESS
+
+* POST /item-storage/items error message on non-UUID statistical code ID (MODINVSTOR-755)
+
 ## 24.0.0 2022-07-01
 
 * Upgraded RMB to 34.0.0 ([MODINVSTOR-915](https://issues.folio.org/browse/MODINVSTOR-915))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+## 24.0.2 2022-08-10
+
+* Fixing when migration script to populate holdings effective locations runs ([MODINVSTOR-940] (https://issues.folio.org/browse/MODINVSTOR-940))
+
 ## 24.0.1 2022-07-18
 
 * Improve populating shelfKey from callNumber ([MODINVSTOR-932](https://issues.folio.org/browse/MODINVSTOR-932))
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
-## 25.0.0 IN-PROGRESS
+## 24.0.1 2022-07-18
 
-* POST /item-storage/items error message on non-UUID statistical code ID (MODINVSTOR-755)
+* Improve populating shelfKey from callNumber ([MODINVSTOR-932](https://issues.folio.org/browse/MODINVSTOR-932))
+* POST /item-storage/items error message on non-UUID statistical code ID ([MODINVSTOR-755](https://issues.folio.org/browse/MODINVSTOR-755))
 
 ## 24.0.0 2022-07-01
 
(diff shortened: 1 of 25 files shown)


skoczko:
@marc-johnson why has not one reviewed/approved this? CC @marc-johnson

marcjohnson-kint:
> @marc-johnson why has no one reviewed/approved this?

I don't know. I suggest you ask the Spitfire team, they made the release.

The Prokopovych team did not perform a review because no one asked for a review. We do not the capacity to monitor the repositories or to review changes made by other teams (and IIRC it is still policy that teams can work without owning review).

It isn't unusual for other teams to release the inventory modules, as multiple teams actively work on them.

adamdickmeiss:
In any case the tag v24.0.3 is not belonging to any branch  https://github.com/folio-org/mod-inventory-storage/tree/v24.0.3 . Normally a released work happn on branch b24.0 which does exist. But the PR suggests that master was used (and master is 25.0.0-SNAPSHOT).

marcjohnson-kint:
> In any case the tag v24.0.3 is not belonging to any branch

That is the nature of the maven release process we use. None of the tags belong to a branch.

> Normally a released work happn on branch b24.0 which does exist. 

This is what I would expect to. I'm awaiting a response from the Spitfire team.

> the PR suggests that master was used (and master is 25.0.0-SNAPSHOT).

I'm assuming this was a mistake. Master is no longer 25.0.0 I believe, this merge regressed that.

marcjohnson-kint:
It seems I was wrong, this merge did not regress the version in the mainline, which surprises me. However that is at least a good thing.

adamdickmeiss:
The release branch is not in GitHub at least perhaps on someone's machine is not visible to us. So the only way to inspect what is there is to download the tar.

marcjohnson-kint:
> The release branch is not in GitHub at least perhaps on someone's machine is not visible to us. So the only way to inspect what is there is to download the tar.

I'm not sure I follow this. Wouldn't the tag be what was included in the release itself?

As I understand it the issue is that the 24.0 branch is now not reflective of that release and cannot be used as the basis for further releases.

adamdickmeiss:
> > The release branch is not in GitHub at least perhaps on someone's machine is not visible to us. So the only way to inspect what is there is to download the tar.
> 
> I'm not sure I follow this. Wouldn't the tag be what was included in the release itself?
Not the git tag.
> 
> As I understand it the issue is that the 24.0 branch is now not reflective of that release and cannot be used as the basis for further releases.
I'd still continue from b24.0. Looking at the tar from v24.0.3 I *think* it's based b24.0, but since it's really git managed anymore being a tar one would have to diff things manually.

marcjohnson-kint:
> I'd still continue from b24.0. Looking at the tar from v24.0.3 I _think_ it's based b24.0, but since it's really not git managed anymore being a tar one would have to diff things manually.

If we release 24.0.4 from the 24.0 branch and that does not include the changes from 24.0.3, then those will be removed in 24.0.4

adamdickmeiss:
> > I'd still continue from b24.0. Looking at the tar from v24.0.3 I _think_ it's based b24.0, but since it's really not git managed anymore being a tar one would have to diff things manually.
> 
> If we release 24.0.4 from the 24.0 branch and that does not include the changes from 24.0.3, then those will be removed in 24.0.4

Yep. One would have to ensure those are cherry-picked to b24.0 .. Which already has 24.0.4-SNAPSHOT and changes.. The news file is messed up as it of course does not mention 24.0.3 but also messed up because it *may* not include 24.0.3 changes.

adamdickmeiss:
I was a bit wrong b24.0 was using 24.0.4-SNAPSHOT..  https://github.com/folio-org/mod-inventory-storage/commit/4f9b288caf08722d2bb19c05a4dee21b2793015a


Then went back to 24.0.3-SNAPSHOT  https://github.com/folio-org/mod-inventory-storage/commit/98915bc5e921d21ea7eac05fa82d837b905bd715
..!!!

24.0.3 now released.

Shans-Kaluhin:
> 24.0.3 now released.

Hi @adamdickmeiss, v24.0.3 was released 5 days ago [Jenkins build](https://jenkins-aws.indexdata.com/job/folio-org/job/mod-inventory-storage/view/tags/job/v24.0.3/) And now it's running on the Morning-Glory bugfest environment.
From comparing releases v24.0.2 and v24.0.3. we can see changes that we merged to master [Git comparing](https://github.com/folio-org/mod-inventory-storage/compare/v24.0.2...v24.0.3).
In this PR I just update NEWS and tag to the current version.

Shans-Kaluhin:
It looks like the main problem is that I did not merged changes to b24.0.

marcjohnson-kint:
> It looks like the main problem is that I did not merged changes to b24.0.

Yes, this is the primary challenge right now, because that means we cannot be confident of using that as the basis for further patch releases.

Beyond that, I think it could be worth reviewing how this release was made (I already updated the JIRA information that was 
out of date), as a patch release should not involve any merge into the mainline branch.

Shans-Kaluhin:
@marcjohnson-kint what if I create pull-request for b24.0. with changes from v24.0.3

marcjohnson-kint:
> @marcjohnson-kint what if I create pull-request for b24.0. with changes from v24.0.3

Yes, please do that.

Shans-Kaluhin:
Sorry for troubles, here is it: https://github.com/folio-org/mod-inventory-storage/pull/815

PR #813, opened 2022-08-18
MODINVSTOR-892 Create a pre-defined Authority Source file list

oliinyko:
**Purpose** 
Create a pre-defined Authority Source file list.

Context: Catalogers want a way to know the source of their authority records. They want a facet/filter that provides them the ability to only view records of the selected source or sources. This knowledge may prove important to our team in supporting the link between bibliographic records and authority records.

JIRA: https://issues.folio.org/browse/MODINVSTOR-892

**Approach**
- CRUD created and reference data populated
- Authority schema is updated with new 'sourceFileId' field


ramls/authorities/authority.json
```diff
@@ -245,6 +245,11 @@
         ]
       }
     },
+    "sourceFileId": {
```
psmagin, line 248: authority API interface version should be increased

psmagin (approved):

maryna-steshenko (approved):

PR #818, opened 2022-08-25
MODINVSTOR-952 Extend Pre-defined Authority Source file list to include a base URL

oliinyko:
**Purpose**
We need to populate a linked bib field $0 with a valid URI so that it's easier to support updating this information with rollout of additional capabilities. 

JIRA: https://issues.folio.org/browse/MODINVSTOR-952

**Approach**
- AuthoritySourceFile schema is updated with 'baseUrl' field
- AuthoritySourceFile schema is updated with 'source' field to distinguish pre-defined and custom records.

### DIFF
--- a/ramls/authoritysourcefile.json
+++ b/ramls/authoritysourcefile.json
@@ -23,6 +23,14 @@
       "type": "string",
       "description": "Type of authority records stored in source file"
     },
+    "baseUrl": {
+      "type": "string",
+      "description": "Base URL of the source file origin"
+    },
+    "source": {
+      "type": "string",
+      "description": "label indicating where the authority source file entry originates from, i.e. 'folio' or 'local'"
+    },
     "metadata": {
       "type": "object",
       "$ref": "raml-util/schemas/metadata.schema",
--- a/ramls/authoritysourcefile.json
+++ b/ramls/authoritysourcefile.json
@@ -29,7 +29,11 @@
     },
     "source": {
       "type": "string",
-      "description": "label indicating where the authority source file entry originates from, i.e. 'folio' or 'local'"
+      "description": "label indicating where the authority source file entry originates from, i.e. 'folio' or 'local'",
+      "enum": [
+        "folio",
+        "local"
+      ]
     },
     "metadata": {
       "type": "object",
@@ -41,7 +45,8 @@
   "required": [
     "name",
     "codes",
-    "type"
+    "type",
+    "source"
   ]
 }
 
--- a/src/main/java/org/folio/rest/impl/AuthoritySourceFileAPI.java
+++ b/src/main/java/org/folio/rest/impl/AuthoritySourceFileAPI.java
@@ -13,9 +13,9 @@ import org.folio.rest.persist.PgUtil;
 
 public class AuthoritySourceFileAPI implements org.folio.rest.jaxrs.resource.AuthoritySourceFiles {
 
-  public static final String REFERENCE_TABLE = "authority_source_file";
+  private static final String REFERENCE_TABLE = "authority_source_file";
 
-  public static final String URL_PROTOCOL_PATTERN = "^(https?://www\\.|https?://|www\\.)";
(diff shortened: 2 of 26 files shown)


ramls/authoritysourcefile.json
```diff
@@ -23,6 +23,14 @@
...
       "description": "Type of authority records stored in source file"
     },
+    "baseUrl": {
+      "type": "string",
+      "description": "Base URL of the source file origin"
+    },
+    "source": {
+      "type": "string",
+      "description": "label indicating where the authority source file entry originates from, i.e. 'folio' or 'local'"
+    },
```
psmagin, line 33: Let's introduce the enum. may be: predefined, local

src/main/java/org/folio/rest/impl/AuthoritySourceFileAPI.java
```diff
@@ -5,14 +5,18 @@
...
 import org.folio.rest.annotations.Validate;
 import org.folio.rest.jaxrs.model.AuthoritySourceFile;
 import org.folio.rest.jaxrs.model.AuthoritySourceFiles;
 import org.folio.rest.persist.PgUtil;
 
 public class AuthoritySourceFileAPI implements org.folio.rest.jaxrs.resource.AuthoritySourceFiles {
+
   public static final String REFERENCE_TABLE = "authority_source_file";
 
+  public static final String URL_PROTOCOL_PATTERN = "^(https?://www\\.|https?://|www\\.)";
```
psmagin, line 18: Can they be private?

src/main/java/org/folio/rest/impl/AuthoritySourceFileAPI.java
```diff
@@ -65,8 +71,22 @@ public void putAuthoritySourceFilesById(String id, String lang,
                                           Map<String, String> okapiHeaders,
                                           Handler<AsyncResult<Response>> asyncResultHandler,
                                           Context vertxContext) {
+    normalizeBaseUrl(entity);
+
     PgUtil.put(REFERENCE_TABLE, entity, id, okapiHeaders, vertxContext,
       PutAuthoritySourceFilesByIdResponse.class,
       asyncResultHandler);
   }
```
psmagin, line 79: What about this requirement? 
"Predefined Base URL can be overridden by a user with permission to do so AND that override can be set to null "

Maybe PATCH endpoint?

psmagin: And also it should be restricted to edit other fields or delete predefined entities. I think we can use your `source` field to identify what can be edited or deleted.

psmagin (changes requested):

src/main/java/org/folio/rest/impl/AuthoritySourceFileAPI.java
oliinyko: > "Predefined Base URL can be overridden by a user with permission to do so AND that override can be set to null "
Maybe PATCH endpoint?

> And also it should be restricted to edit other fields or delete predefined entities. I think we can use your `source` field to identify what can be edited or deleted.


@psmagin Anyway we should keep PUT with possibility to modify the whole entity as it is used by [TenantLoading](https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/main/java/org/folio/rest/tools/utils/TenantLoading.java) service when reference data needs to be updated. I've discussed this with Khalila, seems that she is ok with restricting modification on UI side (as it's done for some other resources i.e. [classificationTypes](https://folio-snapshot.dev.folio.org/settings/inventory/classificationTypes)). The discussion is here [MODINVSTOR-892](https://issues.folio.org/browse/MODINVSTOR-892)

ramls/authoritysourcefile.json
oliinyko: folio and local seems to be a common naming for this purpose
Example of settings page with Classification identifier types: image

Shans-Kaluhin (approved):

viacheslavkol (approved):

PR #823, opened 2022-09-27
MODINVSTOR-924: Batch update with optimistic locking disabled

julianladisch:
* provides `instance-storage-batch-sync-unsafe 1.0`: /instance-storage/batch/synchronous-unsafe
* provides `holdings-storage-batch-sync-unsafe 1.0`: /holdings-storage/batch/synchronous-unsafe
* provides `item-storage-batch-sync-unsafe 1.0`: /item-storage/batch/synchronous-unsafe

Implement using RMB methods from `PgUtil` and `OptimisticLockingUtil`.

This requires https://github.com/folio-org/raml-module-builder/pull/1083 that hasn't been released yet and therefore temporarily needs a snapshot dependency on RMB; must be replaced by a non-snapshot release of RMB before releasing mod-inventory-storage.


ramls/holdings-sync.raml
```diff
@@ -48,3 +48,29 @@ types:
         body:
           text/plain:
             example: "Internal server error"
+/holdings-storage/batch/synchronous-unsafe:
```
marcjohnson-kint, line 51: As I understand it, RAML files are roughly equivalent to interface definitions.

Given that the unsafe variations of the batch operations are defined in separate interfaces, I think they should also be in separate RAML files.

marcjohnson-kint:
My interpretation of this change is that it provided two separate ways to allow for operations to avoid optimistic concurrency control:
* the suppression of optimistic concurrency control (based upon an environment variable) for _all_ operations
* the _unsafe_ batch endpoints that always disable optimistic concurrency control

How wrong is my understanding of these changes?

ramls/holdings-sync.raml
julianladisch: Good point, split.

julianladisch:
The existing APIs always have optimistic locking - they always require the valid _version property. It is not possible to disable optimistic locking for the existing APIs (unless the optimistic locking trigger is disabled in the database).

When using the unsafe API all three requirements must be met:
* invoking user has the batch-unsafe.post permission for the API (checked by Okapi)
* mod-inventory-storage has been started with DB_ALLOW_SUPPRESS_OPTIMISTIC_LOCKING environment variable
* not expired = DB_ALLOW_SUPPRESS_OPTIMISTIC_LOCKING timestamp is in the future, not in the past

NEWS.md
```diff
@@ -10,11 +10,14 @@
...
 * Extend instance contributors schema with Authority ID ([MODINVSTOR-950](https://issues.folio.org/browse/MODINVSTOR-950))
-* provides `instance-storage 9.0`
 * Require CQL query for bulk delete of instances/holdings/items ([MODINVSTOR-576](https://issues.folio.org/browse/MODINVSTOR-576), [MODINVSTOR-901](https://issues.folio.org/browse/MODINVSTOR-901))
 * provides `instance-storage 9.0`
 * provides `holdings-storage 6.0`
 * provides `item-storage 10.0`
+* Batch update with optimistic locking disabled ([MODINVSTOR-924](https://issues.folio.org/browse/MODINVSTOR-924))
+* provides `instance-storage-batch-sync-unsafe 1.0`
+* provides `holdings-storage-batch-sync-unsafe 1.0`
+* provides `item-storage-batch-sync-unsafe 1.0`
```
marcjohnson-kint, line 20: Please organise all of the `provides` and `requires` statements to be at the end of the news entries for a release (as per the convention established for these modules)

pom.xml
```diff
@@ -34,8 +34,8 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     <ramlfiles_path>${basedir}/ramls/</ramlfiles_path>
-    <raml-module-builder-version>34.0.0</raml-module-builder-version> <!-- also update vertx.version -->
-    <vertx.version>4.3.1</vertx.version>
+    <raml-module-builder-version>35.0.0-SNAPSHOT</raml-module-builder-version> <!-- also update vertx.version -->
```
marcjohnson-kint, line 37: Now that RMB 35.0.0 has been released, please update this dependency

marcjohnson-kint:
> The existing APIs always have optimistic locking - they always require the valid _version property. It is not possible to disable optimistic locking for the existing APIs (unless the optimistic locking trigger is disabled in the database).
> 
> When using the unsafe API all three requirements must be met:
> 
> * invoking user has the batch-unsafe.post permission for the API (checked by Okapi)
> * mod-inventory-storage has been started with DB_ALLOW_SUPPRESS_OPTIMISTIC_LOCKING environment variable
> * not expired = DB_ALLOW_SUPPRESS_OPTIMISTIC_LOCKING timestamp is in the future, not in the past

What is the behaviour of the unsafe batch APIs when that environment variable is not set?

Please add documentation to the readme explaining how the unsafe batch APIs work under these different conditions.

Also, please include the new supported environment variables in the `envs` section of the module descriptor (and in the launch descriptor if relevant).

marcjohnson-kint:
> * invoking user has the batch-unsafe.post permission for the API (checked by Okapi)

Is this a general permission used across all of the unsafe batch APIs or is this used to represent the regular individual required permissions for each endpoint?

NEWS.md
julianladisch: Done.

julianladisch:
> What is the behaviour of the unsafe batch APIs when that environment variable is not set?

This is explained in the API documentation:
* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/instance-sync-unsafe.raml
* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/holdings-sync-unsafe.raml
* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/item-sync-unsafe.raml

julianladisch:
> >  * invoking user has the batch-unsafe.post permission for the API (checked by Okapi)
> 
> Is this a general permission used across all of the unsafe batch APIs or is this used to represent the regular individual required permissions for each endpoint?

This is the API permission declared in ModuleDescriptor:

```
    {
      "id": "item-storage-batch-sync-unsafe",
      "version": "1.0",
      "handlers": [
        {
          "methods": ["POST"],
          "pathPattern": "/item-storage/batch/synchronous-unsafe",
          "permissionsRequired": ["inventory-storage.items.batch-unsafe.post"]
        }
      ]
    },
```

```
    {
      "id": "holdings-storage-batch-sync-unsafe",
      "version": "1.0",
      "handlers": [
        {
          "methods": ["POST"],
          "pathPattern": "/holdings-storage/batch/synchronous-unsafe",
          "permissionsRequired": ["inventory-storage.holdings.batch-unsafe.post"]
        }
      ]
    },
```

```
    {
      "id": "instance-storage-batch-sync-unsafe",
      "version": "1.0",
      "handlers": [
        {
          "methods": ["POST"],
          "pathPattern": "/instance-storage/batch/synchronous-unsafe",
          "permissionsRequired": ["inventory-storage.instances.batch-unsafe.post"]
        }
      ]
    },
```

julianladisch:
> Please add documentation to the readme explaining how the unsafe batch APIs work under these different conditions.

Which parts of the API documentation

* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/instance-sync-unsafe.raml
* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/holdings-sync-unsafe.raml
* https://github.com/folio-org/mod-inventory-storage/blob/fa91ae96ff216d04138ca10169d9ca8571f0aaaa/ramls/item-sync-unsafe.raml

should be copied to the README.md? Why isn't the API documentation sufficient? Should I add unit tests that compare the copy in the README.md with the original in the API documentation in the .raml files and fails if they differ?

marcjohnson-kint:
> should be copied to the README.md? Why isn't the API documentation sufficient? 

Personally, I prefer the explanation that was provided in a previous comment, where the behaviour was described together.

I find that easier to follow and more comprehensive than the API docs (which, for example, does not mention all of the environment variables).

I think that's especially important for APIs that behaviour in an unusual way and may not be intuitive to folks who regularly consume FOLIO APIs.

It's entirely up to you whether you accept this opinion.

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1058,6 +1078,26 @@ public void canPostSynchronousBatch() {
       .forEach(DomainEventAssertions::assertCreateEventForItem);
   }
 
+  @Test
+  public void cannotPostSynchronousBatchUnsafeIfNotAllowed() {
+    JsonArray itemsArray = threeItems();
+    assertThat(postSynchronousBatchUnsafe(itemsArray), statusCodeIs(413));
```
marcjohnson-kint, line 1084: There are multiple ways for this circumstance to occur. Could it be worth having separate tests for each circumstance (e.g. at least two environment variables are involved in this decision)?

src/test/java/org/folio/rest/api/ItemStorageTest.java
```diff
@@ -1058,6 +1078,26 @@ public void canPostSynchronousBatch() {
...
+  @Test
+  public void canPostSynchronousBatchUnsafe() {
+    OptimisticLockingUtil.configureAllowSuppressOptimisticLocking(
+        Map.of(OptimisticLockingUtil.DB_ALLOW_SUPPRESS_OPTIMISTIC_LOCKING, "9999-12-31T23:59:59Z"));
+
+    JsonArray itemsArray = threeItems();
+    // insert
+    assertThat(postSynchronousBatchUnsafe(itemsArray), statusCodeIs(HttpStatus.HTTP_CREATED));
+    // unsafe update
+    assertThat(postSynchronousBatchUnsafe(itemsArray), statusCodeIs(HttpStatus.HTTP_CREATED));
```
marcjohnson-kint, line 1096: What is this update intended to achieve? 

It seems to be a resubmission of the same record representations of from the first request.

This tests doesn't seem to check that the expected state is achieved beyond the status code of the response.

julianladisch: If optimistic locking is enabled an insert always succeeds. Therefore we need to do an update to investigate whether optimistic locking is enabled or disabled.

When mod-inventory-storage uses a software library we don't test the software library again but we test whether the software library is correctly used by the mod-inventory-storage code. Here we only need to test the HTTP response code to find out whether the update is rejected with an optimistic locking failure or it succeeded without 200 status code.

If you suspect that the HTTP response code and the record state (unchanged or changed) do not align then this should be tested in the RMB unit tests. We don't want to repeat all these tests for each mod-inventory-storage API (instance, holdings, item) that call the RMB library code for unsafe batch updates.

src/test/java/org/folio/rest/api/ItemStorageTest.java
marcjohnson-kint: > The multiple ways (undefined, expired, not expired) are already tested in RMB: https://github.com/folio-org/raml-module-builder/blob/master/domain-models-runtime/src/test/java/org/folio/rest/tools/utils/OptimisticLockingUtilTest.java#L137-L154
> 
> Which other ways are missing from the tests in RMB? I will add them at RMB.

I'm not reviewing the changes to RMB, only to mod-inventory-storage.

> When mod-inventory-storage uses a software library we don't test the software library again but we test whether the software library is correctly used. Here we only need to test whether the env variable actually can switch on and switch off the unsafe API.

I think we likely have a very different framing of the value of automated testing. I value that the tests demonstrate the behaviour of the system in a way that folks can use to understand the system. I want to know that the system behaves in the intended way, irrespective of the underlying libraries / frameworks being used.

I think this is especially important in this situation where a framework like RMB is so dominant in the implementation of the module.

Some of those differences might be exacerbated by the levels of trust we have in the tooling. I imagine you have a much higher degree of trust in RMBs functionality given you implemented some of it (especially in this case).

Using your framing, how can I know that the library is being used correctly from only a single example?

If we were to apply that reasoning to all of the tests in the module, then we should only have a single test for each endpoint that predominantly use RMB methods e.g. only a single test for creating a record or submitting a batch, irrespective of the context of the batch, as that would be sufficient to demonstrate the usage.

At that point, I'd be tempted to switch to an interaction based testing style, mock the RMB components and check that they were invoked as expected (and save the overhead of the system based tests we currently have).

I think these differences in perspective on automated testing might mean that any feedback I provide on tests won't be valuable to you because it will be based upon different desirable characteristics.

src/test/java/org/folio/rest/api/ItemStorageTest.java
julianladisch: > Why implement this test at all, instead of trusting the existing RMB tests (and the existing tests for the safe batch endpoints)?
I trust that RMB works correctly if called correctly.
But I don't know whether the code in mod-inventory-storage correctly calls the RMB code and whether the response from the RMB code is correctly propagated.
Therefore tests in mod-inventory-storage are needed to show
* that the RMB method is actually called
* that the correct parameters are passed to the RMB method
* that the result from the RMB method is correctly handled, for example returned as JSON body of the mod-inventory-storage API.

adamdickmeiss (approved):

PR #825, opened 2022-10-10
MODINVSTOR-965: Upgrade to RAML Module Builder 35.x.x.

kaladay:
https://issues.folio.org/browse/MODINVSTOR-965


K-Felk (approved):
I also have spent many hours trying to understand why tests in this module time out so much, without success.  :(

marcjohnson-kint (approved):

PR #830, opened 2022-10-20
MODINVSTOR-973: Revert "Merge pull request #828 from MODINVSTOR-972"

julianladisch:
This reverts commit 13127d8ed0f5bc07c4441327301354bb8839d9a1, reversing changes made to 411cc47de64a08ed0394c81426c993210f5a2f2f.

MODINVSTOR-972 reverted
* MODINVSTOR-965: Upgrade to RAML Module Builder 35.x.x
* MODINVSTOR-924: Batch update with optimistic locking disabled

This (MODINVSTOR-973) is a revert of the revert MODINVSTOR-972.

After the revert of the revert MODINVSTOR-965 and MODINVSTOR-924 are available again.

The pull request also includes "MODINVSTOR-974: Remove DB_*_READER env var from ModuleDescriptor" = PR #831


jakub-id:
@marcjohnson-kint I also understood this was the plan. But since your other PR was failing and you closed it can we go ahead and approve this? I would really like for us to get going on this so we can see the snapshot build succeed (or fail). Thanks. cc @julianladisch

marcjohnson-kint:
@jakub-id 

> I also understood this was the plan. But since your other PR was failing and you closed it can we go ahead and approve this? I would really like for us to get going on this so we can see the snapshot build succeed (or fail). Thanks.

I have no issue with this work progressing. Given the confusion about how this work is organised and who is doing it, I would prefer you or another member of Core Platform who understands this approach approve this change, for example, I don't understand how this fits with #831 

I'm also pretty uncomfortable approving this change because it gives the impression that I endorse the changes included in this work.

If that isn't possible, and the only way way folks are willing to move forward is for me to approve this work, then I will reluctantly do so.

marcjohnson-kint:
I think I now understand better how this work was organised. I think #831 is a pull request into the branch for this pull request.

I believe that allows me to approve that change separately to this work, which I appreciate, thank you, even though it's made me understanding this work more complicated.

I have approved that change.

Due to the way those changes are currently separate, I cannot approve this change as it stands because it would cause a recurrence of the same outage.

I don't think there is anything else I need to do in order to move this work forward, due to how the changes have been split into separate linked PRs.

adamdickmeiss (approved):

julianladisch:
@marcjohnson-kint MODINVSTOR-974-remove-DB_x_READER has been successfully merged into this PR.

marcjohnson-kint:
@julianladisch Sure and @adamdickmeiss has approved this pull request. I don't believe there is anything I need to do.

PR #832, opened 2022-10-20
MODINVSTOR-975: ItemStorageTest.canMoveItemToNewInstance Timeout

julianladisch:
5% of all builds fail with a ConditionTimeout in the await() in ItemStorageTest.canMoveItemToNewInstance after migrating to RMB 35.0 and Vert.x 4.3.4.

The current code in assertUpdateEventForItem
* waits for a minimum number of events for the item
* takes the last UPDATE event
* and asserts that the event content matches the expectations:
  * old item content, new item content, event type is UPDATE

The new approach is less brittle:
* Wait until there exists an event that matches the expectations.

This better ignores any other events that might appear in the event stream because of the async nature of the event implementation.


marcjohnson-kint:
> Wait until there exists an event that matches the expectations.

I agree that this is a preferable approach to this assertion.

marcjohnson-kint (approved):
Thank you for investigating the test failures.

I think this has raised my awareness of how unstable these tests are (I don't really understand why they have got significantly more unstable recently).

With that in mind, whilst I think there could be a need for a more general overhaul of these aspects of the tests (e.g. to clear out any messages before the assertion / check part of the test and to remove branching in the checks, which demonstrates false generalisation), I'm in favour of merging this in case it improves upon the current situation.

adamdickmeiss (approved):

PR #833, opened 2022-10-21
MODINVSTOR-976: AsyncMigrationTest:92 ConditionTimeout

julianladisch:
80% of all builds fail with a ConditionTimeout in AsyncMigrationTest.canMigrateItemsInstances:92

Replacing
`kafkaProducer.close()`
by
`kafkaProducer.flush(x -> kafkaProducer.close())`
in CommonDomainEventPublisher reduces this to 50%.

### DIFF
--- a/src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
+++ b/src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
@@ -149,7 +149,7 @@ public class CommonDomainEventPublisher<T> {
       });
 
     return promise.future()
-      .onComplete(e -> kafkaProducer.close())
+      .onComplete(e -> kafkaProducer.flush(x -> kafkaProducer.close()))
       .onSuccess(records -> log.info("Total records published from stream {}", records));
   }
 
@@ -165,7 +165,7 @@ public class CommonDomainEventPublisher<T> {
     return producer.send(producerRecord)
       .<Void>map(notUsed -> null)
       .onComplete(result -> {
-        producer.close();
+        producer.flush(x -> producer.close());
 
         if (result.failed()) {
           log.error("Unable to send domain event [{}], payload - [{}]",


src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
```diff
@@ -149,7 +149,7 @@ public <R> Future<Long> publishStream(ReadStream<R> readStream,
       });
 
     return promise.future()
-      .onComplete(e -> kafkaProducer.close())
+      .onComplete(e -> kafkaProducer.flush(x -> kafkaProducer.close()))
```
adamdickmeiss, line 152:
```suggestion
      .eventually(e -> kafkaProducer.flush())
      .eventually(e -> kafkaProducer.close())
```

adamdickmeiss: According to my experiments, it is sufficient to wait for the close to happen.

```suggestion
      .eventually(x -> kafkaProducer.close())
```

There is a similar problem in publish where onComplete is used. That can be simplified too and actually wait for close to succeed or fail.

marcjohnson-kint, line 152: This change is in production code, does that mean that failures could happen in production usage?

If so, what is the impact of those?

adamdickmeiss: Good question.

The original close has the impact of "close". The close returns a Future which is ignored. So it may not be closed when the method completes.

With the new code it flushes and closes, but also that is not guaranteed to be complete until completion which is why I think it's more proper to use eventually here.

A more fundamental question is whether close should be used at all. If the Kafka Client works like everything else in Vert.x, it's a pool and you better reuse it and not close at all. Using close+flush will only make things even slower.

adamdickmeiss:
Closing as it's replaced by #838

PR #835, opened 2022-10-21
MODINVSTOR-977: hasRemoveEvent fixing sporadic unit test failures

julianladisch:
About 30% of all builds fail because of at least one of these failures:

```
InstanceStorageTest.canDeleteInstancesByCql:487
  Expected: is "DELETE"
       but: was "CREATE"
```

```
  HoldingsStorageTest.canDeleteHoldingsByCql:647
Expected: is "DELETE"
     but: was "CREATE"
```

```
ItemStorageTest.canDeleteItemsByCql:2045
  Expected: is "DELETE"
       but: was "CREATE"
```

The current tests assert that the last event is the expected DELETE event. Due to the async nature of Kafka messages there might be a CREATE event.

Rewrite the test so that it succeeds if the expected DELETE event exists at any position in the event stream.

Write a hasRemoveEvent method that can be used with await().until(() -> hasRemoveEvent(...));

### DIFF
--- a/src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
+++ b/src/test/java/org/folio/rest/support/matchers/DomainEventAssertions.java
@@ -38,6 +38,7 @@ import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.kafka.client.consumer.KafkaConsumerRecord;
 import io.vertx.kafka.client.producer.KafkaHeader;
+import java.util.Collection;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


marcjohnson-kint:
Thank you for investigating the test failures.

I think this has raised my awareness of how unstable these tests are (I don't really understand why they have got significantly more unstable recently).

With that in mind, whilst I think there could be a need for a more general overhaul of these aspects of the tests (e.g. to clear out any messages before the assertion / check part of the test and to use matchers rather than avoiding assertion exceptions), I'm in favour of merging this in case it improves upon the current situation.

marcjohnson-kint (approved):

adamdickmeiss (commented):
I prefer to use FakeKafkaConsumer.removeAllEvents() before the test begins, so that only events  from *this* tests are seen.

Other than that, before MODINVSTOR-976 should be fixed first.

marcjohnson-kint:
> I prefer to use FakeKafkaConsumer.removeAllEvents() before the test begins, so that only events from _this_ tests are seen.

I agree. Sometimes I also want to remove the events before the important act in the test. Some of the steps to prepare the initial state for a test can also publish events and those can get in the way as well.

adamdickmeiss:
Calling FakeKafkaConsumer.removeAllEvents() does not fix the ItemStorageTest case, so I think we should go with the solution here.

adamdickmeiss (approved):

PR #838, opened 2022-10-25
Fix AsyncMigrationTest:92 ConditionTimeout MODINVSTOR-976

adamdickmeiss:
Streams are reused and closing a stream will make exceptions like java.lang.IllegalStateException:
 Cannot perform operation after producer has been closed


marcjohnson-kint (approved):
I don't know enough about the producer implementation to know what the right changes are.

I'm happy to try this change out in case it makes the builds (and the system) more stable.

PR #841, opened 2022-10-25
Use JUnit 5 to run tests MODINVSTOR-978

marcjohnson-kint:


adamdickmeiss (approved):
For this to take effect, wouldn't tests actually have to be rewritten?

marcjohnson-kint:
> For this to take effect, wouldn't tests actually have to be rewritten?

To get the benefit of new features, the tests need to be rewritten.

To help make that transition incrementally, JUnit 5 provides a [runner](https://junit.org/junit5/docs/current/user-guide/#migrating-from-junit4-running) for running JUnit 4 tests alongside JUnit 5 tests.

K-Felk:
It looks like as long as we make sure the vintage test runner is imported, (and it is) Junit4 tests will continue to be supported.

K-Felk (approved):

PR #843, opened 2022-10-26
MODINVSTOR-982: Find and reduce intermittent test failures that still happen

kaladay:
https://issues.folio.org/browse/MODINVSTOR-982

The tests pass when run individually but have intermittent failures when run as a whole.
This makes testing and debugging rather inconsistent as well.
Several of these changes are in theory given the inconsistent test results.
Each change set has its own isolated commit.

These changes do not solve all of the test failures that inconsistently happen but they do appear to solve some of them.

### Kafka events are not consistently cleared between tests.
Commit b7038749c176ca6acd21e57ed702a10364e79def.
The states are not always cleaned before tests.
Many, but not all, of the tests clear data before starting and very few clear data after completing.
This results in tests not actually running in isolation.
This can cause inconsistent and incorrect test results.
    
This takes a compatible approach by implementing a custom function `removeAllEvents()`.
Ideally, this should instead be named `beforeEach()` but then any class that implements this would need to be updated to call `super.removeAllEvents()`.

### Incorrect UUID used for "Invalid ISMN" in test.
Commit b64d06b88cd3239af399a10b90da3a0d9025e99e.

The same UUID is used for both "ISMN" and "Invalid ISMN".
This can cause inconsistent and incorrect test behavior.
    
The test database data (`src/test/resources/lotus-23.0.0.sql`) has a UUID of `4f07ea37-6c7f-4836-add2-14249e628ed1` so use that instead.

### The FakeKafkaConsumer should return first and last events by date.
Commit 594ff313b855a0228cbc17579f45470ad0395ae4

This could be an incorrect interpretation but it seems to me that the events are assumed to be ordered in the list by date. The last received event is expected to the last occurred event and this is expected to therefore be the last event in the list.

The events arrive in an asynchronous manner.
This results in the possibility of the first and last events being ordered differently than the date those events were generated.
If this list is not supposed to be ordered by date such that fetching the last item is the last item by date and the first item is the first item by date, then this commit should be reverted.

These changes are independent of but seem to focus on the same type of problem as addressed by "MODINVSTOR-977: hasRemoveEvent fixing sporadic unit test failures (#835)" (8a3327103dce753720d5b026a01b715d44b0191f).

### Real time is not guaranteed, add 1 second offset.
commit 6102f28ebf082b820552423b922a32c475406e1e

I've seen a test using `StorageTestSuite.tenantOp()` fail at this spot and I am hypothesising that this is due to the non-real-time nature of most time operations in computers (this generally requires a real time OS of which most OS' are not).

A request with a wait of `wait=60000` is being used (I believe that represents 60 seconds in milliseconds).
The code following this is waiting exactly 60 seconds.
Add 1 second to account for how the request does not guarantee real time accuracy and one timer may be off by some arbitrarily small fraction.


marcjohnson-kint:
> This takes a compatible approach by implementing a custom function `removeAllEvents()`.
> Ideally, this should instead be named `beforeEach()` but then any class that implements this would need to be updated to call `super.removeAllEvents()`.

Why is the name of the method important in this case? Given it is the `before` annotation that is the trigger for execution?

src/test/java/org/folio/rest/api/TestBaseWithInventoryUtil.java
```diff
@@ -118,6 +120,11 @@ public static void beforeAny() {
     LocationsTest.createLocation(fourthFloorLocationId,  FOURTH_FLOOR_LOCATION,  "TestBaseWI/FF");
   }
 
+  @Before
+  public void removeAllEvents() {
+    FakeKafkaConsumer.removeAllEvents();
```
marcjohnson-kint, line 125: I agree it is a good idea to remove all of the captured messages before each test

src/test/java/org/folio/rest/api/UpcIsmnMigrationScriptTest.java
```diff
@@ -21,7 +21,7 @@ public class UpcIsmnMigrationScriptTest extends MigrationTestBase {
     ids.put("UPC","1795ea23-6856-48a5-a772-f356e16a8a6c");
     ids.put("Invalid UPC", "b3ea81fb-3324-4c64-9efc-7c0c93d5943c");
     ids.put("ISMN", "ebfd00b6-61d3-4d87-a6d8-810c941176d5");
-    ids.put("Invalid ISMN", "ebfd00b6-61d3-4d87-a6d8-810c941176d5");
+    ids.put("Invalid ISMN", "4f07ea37-6c7f-4836-add2-14249e628ed1");
```
marcjohnson-kint, line 24: Thanks for finding this. 

This kinda leads me to wonder about whether this tests actually tests the intended behaviour if they both had the same ID.

marcjohnson-kint:
> This could be an incorrect interpretation but it seems to me that the events are assumed to be ordered in the list by date. The last received event is expected to the last occurred event and this is expected to therefore be the last event in the list.
> 
> The events arrive in an asynchronous manner. This results in the possibility of the first and last events being ordered differently than the date those events were generated. If this list is not supposed to be ordered by date such that fetching the last item is the last item by date and the first item is the first item by date, then this commit should be reverted.

I agree that this assumption in the tests could cause false failures. I'm ok with changing the implementation to impose a defined order.

Generally, I think the tests likely need reworking to no rely on the ordering at all.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
```diff
@@ -127,13 +128,35 @@ public static Collection<KafkaConsumerRecord<String, JsonObject> > getHoldingsEv
...
+
+    while (iter.hasNext()) {
+      KafkaConsumerRecord<String, JsonObject> record = iter.next();
+
+      if (last == null || record.timestamp() > last.timestamp()) {
+        last = record;
+      }
+    }
+
+    return last;
```
marcjohnson-kint, line 142: I agree that imposing a defined order on this is a reasonable activity (though it is a somewhat false assumption that messages are published in timestamp order).

I would prefer a more intent revealing implementation if possible, especially given this is test code that won't have it's own tests. 

What led to the use of an iterator instead of something like `stream.sorted` or `collections.sort`?

kaladay: > What led to the use of an iterator instead of something like stream.sorted or collections.sort?
I wasn't focusing as much on the details as I probably should have.
The existing code was using events.stream() and I just continued using it without thinking much about it as my focus was more on the test failures.

I'm all for improving this and I also am concerned about the assumption in timestamp order as well.

kaladay:
> > This takes a compatible approach by implementing a custom function `removeAllEvents()`.
> > Ideally, this should instead be named `beforeEach()` but then any class that implements this would need to be updated to call `super.removeAllEvents()`.
> 
> Why is the name of the method important in this case? Given it is the `before` annotation that is the trigger for execution?

It seems to me to be good practice to use names like `beforeEach()` for the functions using `@BeforeEach` annotation.
And ideally there would be only a single `@BeforeEach`, for that matter.
This change is causing there to be two functions using `@BeforeEach`.
It is unclear to me the order in which multiple functions execute.
In this specific case, I think the order is inmaterial and not a problem.

src/test/java/org/folio/rest/api/StorageTestSuite.java
```diff
@@ -296,7 +296,7 @@ public static void tenantOp(String tenantId, JsonObject job) throws InterruptedE
       tenantPrepared = new CompletableFuture<>();
       client.get(storageUrl("/_/tenant/" + id + "?wait=60000"), tenantId,
           ResponseHandler.any(tenantPrepared));
-      response = tenantPrepared.get(60, TimeUnit.SECONDS);
+      response = tenantPrepared.get(61, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
kaladay: Should I remove/revert this change then?

kaladay:
> Generally, I think the tests likely need reworking to no rely on the ordering at all.

Should I revert the ordering commit https://github.com/folio-org/mod-inventory-storage/commit/594ff313b855a0228cbc17579f45470ad0395ae4 and raise a new ticket specifically focusing on redesigning tests to not rely on Kafka messaging ordering?

src/test/java/org/folio/rest/api/StorageTestSuite.java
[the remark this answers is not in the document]
marcjohnson-kint: No need to revert these changes. I only think they are unlikely to resolve the underlying issue (I think it's unlikely that a tenant activation taking 60 seconds is likely to be successful).

Please add a comment explaining why the timeout is 61 seconds.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
marcjohnson-kint: I suggest adding a comment making the assumption about the timestamp order being reliable.

And experiment with using a more expressive way of achieving the sorted list.

marcjohnson-kint:
> > Generally, I think the tests likely need reworking to no rely on the ordering at all.
> 
> Should I revert the ordering commit [594ff31](https://github.com/folio-org/mod-inventory-storage/commit/594ff313b855a0228cbc17579f45470ad0395ae4) and raise a new ticket specifically focusing on redesigning tests to not rely on Kafka messaging ordering?

No need, I'm stating my belief about a general direction for these tests. This is separate to these improvements.

marcjohnson-kint:
> It seems to me to be good practice to use names like `beforeEach()` for the functions using `@BeforeEach` annotation. 

Agreed.

> And ideally there would be only a single `@BeforeEach`, for that matter. This change is causing there to be two functions using `@BeforeEach`. It is unclear to me the order in which multiple functions execute. In this specific case, I think the order is inmaterial and not a problem.

I think I misunderstood the context of this. My understanding now is that you wanted to use `beforeEach` for the name of that method, however because some of the sub-classes already have it defined that would cause a method name class.

kaladay:
> I think I misunderstood the context of this. My understanding now is that you wanted to use `beforeEach` for the name of that method, however because some of the sub-classes already have it defined that would cause a method name class.

Yes that is correct. I wanted to use `beforeEach()` but a sub-class is implementing it and I want to avoid naming conflicts and any possible resulting inheritance problems.

PR #847, opened 2022-11-10
MODINVSTOR-969: Add new sample data for bound-withs

julianladisch:
https://issues.folio.org/browse/MODINVSTOR-969


src/main/java/org/folio/rest/impl/TenantRefAPI.java
```diff
@@ -122,19 +129,30 @@ Future<Integer> loadData(TenantAttributes attributes, String tenantId,
...
+      tl.add("bound-with/bound-with-parts", BOUND_WITH_PARTS);
       tl.withFilter(service -> servicePointUserFilter(service, servicePoints))
         .withPostOnly()
         .withAcceptStatus(422)
-        .add("users", "service-points-users");
+        .add("users", SERVICE_POINTS_USERS);
+      future = future.compose(n -> tl.perform(attributes, headers, vertxContext, n));
+    }
+
+    if (isNew(attributes, "25.1.0")) {
```
adamdickmeiss, line 148: What happens if this code is executed twice and database content is really 25.1.0 or later? The isNew will return true if module is being enabled after being disabled. Yet the database content with sample data already holds these entries if it was 25.1.0 or later at some point.

adamdickmeiss (approved):

PR #849, opened 2022-11-14
Follow up changes regarding MODINVSTOR-982

kaladay:
This is the result of a more in depth investigation regarding https://issues.folio.org/browse/MODINVSTOR-982..

The mod_inventory_storage tests have structurally questionable designs that I believe are hindering the maintainability, reliability, and debugging of the tests.
In one particular case, a parent class (TestBaseWithInventoryUtil) calls a child class (LocationAPI) that also calls unit test class (LocationUnitTest) to perform non-unit-test LocationAPI testing.

This is a proposal of changes that are not necesserily the ideal solution but are instead a step towards an improved situation that ideally should have fewer random failures and should have easier to read and follow code.

### Summarized Highlights
1) StorageTestSuite has several problems.
  a) Due to the pom file excluding everything in `org/folio/rest/api/*Test.java`, there are several tests classes not being run at all (dead test code).
  b) The tests are clearly not isolated, changing the TestSuiteOrder (such as alphabetic sorting) causes tests to fail (fragile test design).
2) Some tests fail due to state related problems.
  a) Enable all disabled test classes like UpdateItemStatusDateFunctionMigrationTest.
  b) ReferenceTablesTest, ShelfLocationsTest, and BoundWithStorageTest test classes now fail.
  c) This is likely a problem due to lack of clean states between different test classes.
  d) Keeping all tests, but removing only ReferenceTablesTest class from suite results in failures.
3) If ReferenceTablesTest class is disabled or fails, then ShelfLocationsTest class might fail.
4) Some of the tests that were disabled fail when enabled (UpdateItemStatusDateFunctionMigrationTest and ReferenceTablesTest).
5) Test order execution problems.
  a) By design, JUnit does not specify the execution order of test method invocations. Originally the methods were simply invoked in the order returned by the reflection API (see https://github.com/junit-team/junit4/wiki/Test-execution-order).
  b) Some of the tests are run outside of the StorageTestSuite but depend on Kafka and other services managed by StorageTestSuite. The order tests are run has an impact on this.
6) InstallUpgradeIt.java, is this even run?
7) The "commit()" functionality of Kafka appears directly related to several of the random failres (such as getting "CREATE" when expecting "UPDATE").

### Detailed Explanation

Double-down on the Junit4 Suite usage to help organize and control test order.
This has a downside of explicitly naming each class to use in the test suites.
Multiple test suites now exist and are used to directly control order.

The ordering problems in StorageTestSuite are not addressed.

[For migration to Junit5, the suites can still be supported.](https://junit.org/junit5/docs/snapshot/user-guide/#running-tests-junit-platform-runner-test-suite)

Significant amounts of code are relocated to newly created utility classes to better keep tests isolated and help make the code easier to debug.
I did not have a particular place in mind so I arbitrarily chose a top level location in the directory structure to create a utility directory at.
The utility classes are grouped by purpose.

I explore exposed the ability to explicitly perform a Kafka commit.
On one hand this seemed to help make success more consistent.
I explicitly added commits in between parts of individual tests and found that the explicit commit is causing more consistent failures.
This is the exact opposite of what I expected.
I turned off auto commit and remove the explicit commit code and found that the random failures happen less often.

There are still occasional timeout failures but these are not as often after disabling auto commit.

Additional code is added to help catch problems and debug problems.
A timestamp is saved and compared against when Kafka messages are received.
This timestamp gets updated between many tests.
The log can be enabled by appending `-Dlog4j.debug` to the `mvn test` command.

Adding the additional Suites required changing the surefire logic in the `pom.xml`.
This revealed that some tests were never being run.
Two of these tests that were never being run consistently fail and have been added and left commented out so that they are not lost again.

There are different ways in which the "test_tenant" is being and in some places is even being hardcoded.
Move all of the hard coded and inconsistent uses to a single location in the utility file.
Now the "test_tenant" defined in the utility file as `TENANT_ID` is used.

The added utility files include significant changes to the Vertx and HttpClient usage.
The circular logic represented by the `invokeStorageTestSuiteAfter` is removed.
This should help make following the logic in the code easier.

Additional logic is added to perform database clean up before tests.
There are still several tests that this is not done to.
I only performed an as-needed case.
The logic could be better fine-tuned to only clean what is needed but more investigation and work is needed for this.

Some of the naming structures are changed to make the eventual transition into Junit5 slightly more straight-forward.
Renaming functions like `setUp` to `beforeEach` to directly match the annotation Junit5 uses `@BeforeEach`.


pom.xml
```diff
@@ -521,12 +521,8 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.19.1</version>
         <configuration>
-          <excludes>
-            <exclude>org/folio/rest/api/*Test.java</exclude>
```
marcjohnson-kint, line 525: This exclusion was in place due to tests being run as part of the suite were also being run outside of the suite.

How is that repeated execution avoided in the new approach?

src/test/java/org/folio/rest/RestTestSuite.java
```diff
@@ -0,0 +1,26 @@
...
+
+@RunWith(Suite.class)
+@Suite.SuiteClasses({
+  CollectionUtilTest.class,
+  CompletableFutureUtilTest.class,
+  CqlQueryTest.class,
+  ItemDamagedStatusAPIUnitTest.class,
+  StorageHelperTest.class,
+  TenantRefAPITest.class,
+})
```
marcjohnson-kint, line 20: Given that this suite has no shared logic, is the only purpose to enforce an execution order on these tests?

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
```diff
@@ -1,39 +1,42 @@
...
+import org.junit.runner.RunWith;
 
 @RunWith(JUnitParamsRunner.class)
 public class AuthorityStorageTest extends TestBase {
 
+  @SneakyThrows
   @Before
   public void beforeEach() {
+    clearData();
     StorageTestSuite.deleteAll(authoritiesStorageUrl(""));
```
marcjohnson-kint, line 37: Why are authorities deleted separately, are they not included in the `clear data` operation?

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
```diff
@@ -106,6 +109,10 @@ public void deleteByIdNotFound() {
   public void putById() {
     assertEquals(0, authoritiesClient.getAll().size());
     createAuthRecords(1);
+
+    // Clear events to help avoid delayed message problems.
+    removeAllEvents();
```
marcjohnson-kint, line 114: Why is this being done part way through a write operation? Is it to remove all captured events published during the write because they aren't relevant?

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
```diff
@@ -13,33 +23,30 @@
...
 @RunWith(JUnitParamsRunner.class)
 public class BoundWithStorageTest extends TestBaseWithInventoryUtil {
-  static ResourceClient boundWithPartsClient  = ResourceClient.forBoundWithParts(client);
+  static ResourceClient boundWithPartsClient  = ResourceClient.forBoundWithParts(getClient());
 
+  @SneakyThrows
+  @After
+  public void beforeEach() {
+    deleteAllById(boundWithPartsClient);
+    clearData();
```
marcjohnson-kint, line 34: Why isn't removing all bound with parts done in the `clear data` operation?

src/test/java/org/folio/rest/api/HridSettingsIncreaseMaxValueMigrationTest.java
```diff
@@ -1,27 +1,35 @@
...
   private static final String ITEMS_SEQ = "hrid_items_seq";
 
+  @SneakyThrows
+  @Before
+  public void beforeEach() {
+    clearData();
+    setupMaterialTypes();
+    setupLoanTypes();
+    setupLocations();
+    removeAllEvents();
```
marcjohnson-kint, line 30: Thank you for using intent revealing names that describe the high level sequence nicely

pom.xml
kaladay: My understanding is that the Inclusion what brought in the *Test.java.
By only including the Suites, the explicit exclusion is not needed.
Or so goes my thinking.

src/test/java/org/folio/rest/RestTestSuite.java
kaladay: The fixed order is only one of the reasons for using a test suite like this.

A purpose here is to use the Suites to bring in the tests rather than via an Inclusion in the pom.xml.
There is an AllTestSuite that is designed to pull in all the other test Suite classes and to control the order in which the individual test suites run.

I don't have any particular idea how to group them so I grouped them based on directory structure.
The "rest" tests are in the rest test suite.
The "api" tests are in the storage test suite.
The "service" tests are in the service test suite.

There is a bonus of being able to run the Suites as a set:
```
mvn test -Dtest=RestTestSuite
```

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
kaladay: > Why are authorities deleted separately, are they not included in the clear data operation?

This is one of the problems I lost a lot of time fighting with.
The `clearData()` does not clear everything.
I've attempted to delete all data before all tests to maintain clean state.
Tests across several test classes do not like this.
They are too interdependent on other tests and data or the delete process needs to be far more complex than it currently is.

I essentially gave up and resigned to just have the clearAll() delete all data in such a way that tests don't suddenly fail because the data was deleted.

There is a huge problem with isolation of the data and I essentially found it too unweildly to try and solve.
I only eased the isolation pains with some of the changes rather than solving them.

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
kaladay: The `createAuthRecords(1);` results in a CREATE message in the Kafka logs.
For whatever reason that CREATE sometimes appears after the UPDATE.

This is an attempt to help avoid these delayed message problems.
Emphasis on attempt.

The real solution is to not rely on the messaging order as we currently do but that is a far bigger task and out of scope of my goals here.

Originally, the `removeAllEvents()` called `commit()` here to help force Kafka to impose some influence over the order. This actually revealed the problem with commit/autocommit in Kafka as a major factor in the random out of order messages.

I removed the commit() and disabled autocommit but left this particular call in places.
The areas in code that have the most consistent random failures based on CREATE appearing after UPDATE retain this function call in the middle of tests and usually after a create operation. I believe I also added that short comment very briefly describing the purpose of the call.

src/test/java/org/folio/rest/api/BoundWithStorageTest.java
kaladay: > Why isn't removing all bound with parts done in the `clear data` operation?

This is one of those cases where adding to the `clearData()` causes other tests (in other test classes) to fail.
There is an isolation problem.
I suspect a complete redesign (that would require far more changes than what I made in this PR) is necessary to properly address this isolation problem.

I gave up and added the clearing of bound with data here

pom.xml
marcjohnson-kint: Cool, that's much neater.

src/test/java/org/folio/rest/RestTestSuite.java
marcjohnson-kint: > I don't have any particular idea how to group them so I grouped them based on directory structure.

That's likely because many of the tests are integration (in the infrastructure sense) tests and thus they aren't really organised in a particular way.

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
marcjohnson-kint: > The `createAuthRecords(1);` results in a CREATE message in the Kafka logs. For whatever reason that CREATE sometimes appears after the UPDATE.

By Kafka logs, do you mean the logs produced by Kafka, the topic itself or the fake consumer that collects in-memory lists?

> This is an attempt to help avoid these delayed message problems. Emphasis on attempt. The real solution is to not rely on the messaging order as we currently do but that is a far bigger task and out of scope of my goals here.

Agreed.

> The areas in code that have the most consistent random failures based on CREATE appearing after UPDATE retain this function call in the middle of tests and usually after a create operation. I believe I also added that short comment very briefly describing the purpose of the call.

Cool. The reason I asked for more info is because the comments often say what removeAllEvents would do and would refer to the `delayed messages problem` however not what that was in the particular context.

kaladay: > By Kafka logs, do you mean the logs produced by Kafka, the topic itself or the fake consumer that collects in-memory lists?

Sorry, I mis-typed.
I was thinking of the messages kafka manages/sends/receives rather than a "log".

> Cool. The reason I asked for more info is because the comments often say what removeAllEvents would do and would refer to the delayed messages problem however not what that was in the particular context.

I can add more verbose comments in these cases then.

src/test/java/org/folio/rest/api/AuthorityStorageTest.java
kaladay: I've added additional documentation as requested.

src/test/java/org/folio/rest/api/ModesOfIssuanceMigrationScriptTest.java
```diff
@@ -2,24 +2,38 @@
...
+
+    StorageTestSuite.deleteAll(modesOfIssuanceUrl(""));
+
+    removeAllEvents();
+  }
+
+  @After
+  public void afterEach() {
     StorageTestSuite.deleteAll(modesOfIssuanceUrl(""));
   }
```
marcjohnson-kint, line 38: Why is this now being done before and after each test?

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
```diff
@@ -1,43 +1,66 @@
...
+  }
+
+  @After
+  public void afterEach() {
+    StorageTestSuite.deleteAll(TENANT_ID, "preceding_succeeding_title");
+    StorageTestSuite.deleteAll(TENANT_ID, "instance_relationship");
+    StorageTestSuite.deleteAll(TENANT_ID, "bound_with_part");
+
+    deleteAllById(precedingSucceedingTitleClient);
+  }
```
marcjohnson-kint, line 62: Why is this being done before and after each test?

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
```diff
@@ -497,7 +508,7 @@ public void instanceStatusesBasicCrud()
     assertThat(postResponse.getStatusCode(), is(HttpURLConnection.HTTP_CREATED));
 
     String entityUUID = postResponse.getJson().getString("id");
-    URL entityUrl = instanceStatusesUrl("/" + entityUUID);
+    instanceStatusesUrl("/" + entityUUID);
```
marcjohnson-kint, line 511: Is this call needed anymore, given the assignment is no longer done? Does it have any side effects? 

If not, I think it can be removed

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -89,108 +184,14 @@ public static <T> T get(Future<T> future) {
...
-    .onSuccess(handler)
-    .onFailure(error -> logger.error(error.getMessage(), error));
-  }
-
   /**
    * Assert that a GET at the url returns 404 status code (= not found).
    * @param url  endpoint where to execute a GET request
    */
-  void assertGetNotFound(URL url) {
+  public void assertGetNotFound(URL url) {
```
marcjohnson-kint, line 191: Why does this need to be public? Is it being used by something outside of TestBase?

src/test/java/org/folio/rest/api/TestBase.java
```diff
@@ -65,6 +74,92 @@ public abstract class TestBase {
   static AuthorityReindexFixture authorityReindex;
   static AsyncMigrationFixture asyncMigration;
 
+  @BeforeClass
+  public static void testBaseBeforeClass() {
```
marcjohnson-kint, line 78: Given that you've gone to active effort to rename the before and after methods in many of these. I think it's worth choosing either `all` or `class` as the postfix for once per class methods.

src/test/java/org/folio/utility/VertxUtility.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private VertxUtility() {
+    throw new UnsupportedOperationException("Cannot instantiate utility class.");
+  }
+
+  public static HttpClient getClient() {
+    return client;
+  }
+
+  public static void startVertx(String tenantId)
```
marcjohnson-kint, line 38: This doesn't really only start vertx, it also deploys the RMB verticle and enable the module for the tenant.

I think it could be easier to follow if the name of this describes these responsibilities more explicitly.

src/test/java/org/folio/utility/VertxUtility.java
```diff
@@ -0,0 +1,171 @@
...
+import java.util.concurrent.TimeoutException;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.rest.RestVerticle;
+import org.folio.rest.support.HttpClient;
+import org.folio.rest.support.Response;
+import org.folio.rest.support.ResponseHandler;
+import org.folio.rest.tools.utils.NetworkUtils;
+
+public class VertxUtility {
```
marcjohnson-kint, line 24: This class seems to be more related to a RMB module rather than vert.x.

Please rename it to describe this scope more specifically e.g. ModuleUtility. Alternatively, split these methods into two classes, one for vert.x and another for the RMB module specific stuff e.g. enabling the module for a tenant

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
```diff
@@ -50,6 +57,15 @@ public FakeKafkaConsumer consume(Vertx vertx) {
...
       final List<KafkaConsumerRecord<String, JsonObject>> storageList;
 
+      // Messages that are earlier than the timestamp are stale and should be ignored.
+      if (message.timestamp() < timestamp.toInstant().getMillis()) {
+        logger.debug("Timestamp is " + message.timestamp()
+          + " less than " + timestamp.toInstant().getMillis()
+          + " for message: " + message.topic() + ", " + message.key() + ", "
+          + message.value().encodePrettily());
+        return;
+      }
```
marcjohnson-kint, line 67: I think this drops messages that have been received out of order based upon the timestamp of the messages.

Have I understood that correctly?

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
```diff
@@ -216,14 +244,18 @@ private static String instanceAndIdKey(KafkaConsumerRecord<String, JsonObject> m
     return instanceAndIdKey(message.key(), id);
   }
 
+  private void setConsumer(KafkaConsumer<String, JsonObject> consumer) {
+    this.consumer = consumer;
+  }
```
marcjohnson-kint, line 249: What benefit does this provide over interacting with the field directly?

src/test/java/org/folio/services/kafka/topic/KafkaTopicsExistsTest.java
```diff
@@ -1,38 +1,79 @@
...
+      ExecutionException,
+      TimeoutException {
 
-  KafkaAdminClient kafkaAdminClient;
+    // tests expect English error messages only, no Danish/German/...
+    Locale.setDefault(Locale.US);
+
+    PostgresClient.setPostgresTester(new PostgresTesterContainer());
+    startKafka();
+    startVertx(TENANT_ID);
```
marcjohnson-kint, line 48: Why does this test need to deploy and enable the module?

src/test/java/org/folio/utility/LocationUtility.java
```diff
@@ -0,0 +1,242 @@
...
+        createLibrary(libraryID, "Main Library", "ML", campusID);
+
+        UUID spID = UUID.randomUUID();
+        servicePointIDs.add(spID);
+        String spName = "Service Point " + spID.toString();
+        createServicePoint(spID, spName, "SP" + spID.toString(), spName, "SP Description", 0, false, null);
+      }
+    } catch (Exception e) { // should not happen
+      throw new AssertionError("CreateLocUnits failed:", e);
+    }
```
marcjohnson-kint, line 60: Could sneaky throws be used to avoid this?

src/test/java/org/folio/utility/RestUtility.java
```diff
@@ -0,0 +1,68 @@
...
+import io.vertx.core.MultiMap;
+import io.vertx.core.buffer.Buffer;
+import io.vertx.core.http.HttpMethod;
+import io.vertx.ext.web.client.HttpResponse;
+import java.net.URL;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class RestUtility {
+  public static final String TENANT_ID = "test_tenant";
```
marcjohnson-kint, line 16: What led you to choose putting this here rather than one of the other utility classes?

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
kaladay: I remember this one giving me a lot of trouble with affecting tests from other classes.

I think I can try and AfterClass rather than an After each.
I am pretty sure I cannot remove the After/AfterClass in this particular case.

src/test/java/org/folio/rest/api/TestBase.java
kaladay: One of my quirks where I strive for consistency.
It just seems that the many other methods in the test classes are generally `public`.

I can remove the `public`.

src/test/java/org/folio/rest/api/ModesOfIssuanceMigrationScriptTest.java
marcjohnson-kint: Cool, thanks. FWIW I prefer before rather than after, because the responsibility for a clean slate lives with the tests that want it

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
marcjohnson-kint: What leads to a `before each` being insufficient to achieve this?

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: > It just seems that the many other methods in the test classes are generally public.

I suspect that's because folks were using the suites / bases as places to put stuff for reuse even when not inheriting.

pom.xml
```diff
@@ -521,12 +521,8 @@
...
-          </excludes>
           <includes>
-            <include>org/folio/rest/api/StorageTestSuite.java</include>
-            <include>**/*Test.java</include>
+            <include>org/folio/AllTestSuite.java</include>
           </includes>
```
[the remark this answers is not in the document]
kaladay: The renames will make the reviews even harder, so I will take the additional PR approach.
One of my reasons for moving into the Suite is to make the transition from junit4 to junit5 more explicit.
Anything removed from the suite would be removed because it is migrated to junit5.

The file rename you suggest also addresses this goal so I have no objections to doing this.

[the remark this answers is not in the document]
marcjohnson-kint: I agree with splitting this into a separate PR.

I think it's worth discussing what we do more after this is merged.

If the goal is to isolate the tests more because they use shared infrastructure and state and that reduces their stability, then we may want to move all of the tests that require this style of integration to maven style integration tests.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
kaladay: Yes this is the idea.
My understanding is that this is using Kafkas timestamp rather than any Folio crafted one.
However, the fact that logs do not appear when debug is enabled has shown that this is not where the problem happens.

I left this debug logger because it still may come in handy should this problem ever actually occur.
Ideally we should never see this log message appear when debug is enabled.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
kaladay: The original code for `public FakeKafkaConsumer consume(Vertx vertx)` creates the  consumer within the method.
There is no access to the `consumer` after that function call and there is the potential for the consumer being a weak reference.

Make the consumer a strong reference.
This is to help avoid the possibility of the `consumer` being closed before everything is done with it.
This appears to help solve several of the connection closed warnings/errors.

This also allows for the following:
```
  public void unsubscribe() {
    consumer.unsubscribe();
  }
```

The consumer is on the class and may now be explicitly unsubscribed.

The `TestBase` then can do this:
```
  @AfterClass
  public static void afterAll() {
    kafkaConsumer.unsubscribe();
  }
```

The `consumer` should be unsubscribed after each class if we are always subscribing to a new one before each class.
This appears to help solve several of the connection closed warnings/errors.

src/test/java/org/folio/services/kafka/topic/KafkaTopicsExistsTest.java
kaladay: I wish I knew.

This test was previously doing this before by its extending of `TestBase`.
When this gets called and is not called via the `StorageTestSuite`, then the `invokeStorageTestSuiteAfter` boolean gets set appropriately and the `StorageTestSuite` methods are called outside of the Suite tests.

This is a major cause of some of the random, inconsistent, or confusing test failures.
I decided not to address what I would consider a design problem with the test due to the scope and focus of my changes. The PR is big enough as it is.

In every place where we have to call `startKafka();` that looks more like a unit test or is not inside the `StorageTestSuite` should be considered for redesign. I do not know what the redesign would look like. I only know that the fact that I have to do this to preserve previous behavior (without having to call StorageTestSuite methods) is a problem.

src/test/java/org/folio/utility/LocationUtility.java
kaladay: I left a lot of the original code or logic alone as much as possible when moving things into the utility.
The problem is that there is no obvious change set to compare against.

I have no problems changing this.
The try..catch.. isn't even needed here.
I can get away without even using sneakythrows.

I'm going to remove the try..catch.. but I am not add the `@SneakyThrows` (unless you explicitly want `@SneakyThrows` added).

src/test/java/org/folio/utility/RestUtility.java
kaladay: The `RestUtility` needs the `TENANT_ID` when adding it to the **REST** requests.
When I noticed that there were different (seemingly random) classes also using hardcoded tenant strings I decided to just have a single location for this. The `RestUtility` already had this and so I exposed the constant.

The `StorageTestSuite` is focused on everything in the suite.
There are tests that are outside of the suite that directly use `TENANT_ID`.
This makes `StorageTestSuite` not a good candidate because it is specialized in regards to certain tests.
I believe that the `RestUtility` is more generalized than `StorageTestSuite`.

The `VertxUtility` (now `ModuleUtility`) is where I moved all of the vertx related code that is shared between many different tests. There is no `TENANT_ID` use in any of those methods and so it does not belong there.

Like `VertxUtility`, the `KafkaUtility` is for the Kafka related shared code and is only providing start and stop methods. The `TENANT_ID` doesn't belong there as well.

A class for providing constants rather than doing double-duty with a utility is likely to be even better.
Doing this would introduce even more changes and deviations that what is already done.
I tried to keep such changes to a minimum to the degree that I could where I was able to.

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
kaladay: The state After the class is done with is different.
Tests from other classes may fail as a result.
The existing of preceding_succeeding_title, instance_relationship, bound_with_part can cause deletes to fail due to constraint violations.
The lack of proper deletion results in an inconsistent state and then tests sometimes have the wrong number of rows.
Then test assertions fail.

Switching to an `@AfterClass` appears to still fix the state for everything outside of this class without running between every test.
We should be able to remove such logic if and once we get data isolation between test classes.

kaladay:
> Thank you for fixing some of the build failures caused by Kafka messages and caused by a state that one test class creates and another test class requires!

The only problem is the "...some of the build failures..." part.
The  `InstanceStorageTest` and `HoldingsStorageTest` test classes still have random errors.
I simply have run out of time to address these remaining random failures.
I have a day or two left to look into this and after that I will likely have other duties to attend to.

Overall I think the occurance of problems is now far less.

src/test/java/org/folio/rest/api/ReferenceTablesTest.java
kaladay: Nope. Removed.

src/test/java/org/folio/rest/api/TestBase.java
kaladay: Done. I've chosen `all` over `class` to make it more consistent with junit5 (which uses `all` instead of `class`).

src/test/java/org/folio/rest/RestTestSuite.java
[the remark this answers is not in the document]
kaladay: I was being a bit excessively paranoid about controlling the order to help avoid the random failures rather early on in the process of writing the changes reflected in this PR.

Given that a new strategy is now proposed to avoid excessive Suite usage, I partially reverted the suite changes to avoid adding code that we will immediately remove in a subsequent PR.

src/test/java/org/folio/utility/RestUtility.java
marcjohnson-kint: Sure, this can go down as something to sort out later

src/test/java/org/folio/utility/LocationUtility.java
marcjohnson-kint: I only suggested the SneakyThrows because I assumed that the catch related to an expected exception. 

If there isn't, then I have no need for SneakyThrows

src/test/java/org/folio/services/kafka/topic/KafkaTopicsExistsTest.java
marcjohnson-kint: Fair enough, this one can wait.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
marcjohnson-kint: > The consumer should be unsubscribed after each class if we are always subscribing to a new one before each class.
This appears to help solve several of the connection closed warnings/errors.

Hmm, that should only really be necessary when a different Kafka instance is used. That suggests that we don't have alignment between Kafka deployments and tests.

marcjohnson-kint: Maybe a comment could be added to explain the use of the strong reference and explicit disconnection (though I do prefer could being explicit, so I'm good with this change).

Maybe in the future, this class becomes instantiated within the tests and not shared.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
marcjohnson-kint: I think we need to be careful with this. 

I don't know the source of the timestamp, however if it is in the message, it probably comes from the publisher.

I also don't think we can rely on only wanting messages received in order.

src/test/java/org/folio/rest/api/TestBase.java
marcjohnson-kint: I very much prefer `all` (and `each`). I only wondered if there was a specific reason this one was different. I know it's hard to track stuff like this when there are so many changes.

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
marcjohnson-kint: Thanks for explaining the reasoning further.

Please add a comment to explain that reasoning.

I reckon this is partly because of the mixture of suite and non-suite tests and inconsistent `before` methods. The original intent was that the `before` would clean the state of the system (without enabling a new tenant because that is the most time consuming preparatory step).

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
kaladay: > I think we need to be careful with this.

I can remove the return statement and leave the log message just in case.

src/test/java/org/folio/rest/support/kafka/FakeKafkaConsumer.java
kaladay: Comments added.

src/test/java/org/folio/rest/api/PrecedingSucceedingTitleTest.java
kaladay: Comments added.

marcjohnson-kint (approved):
I suggest we get this merged and further changes to move away from the suites (inc. how to address the likely significant increase in test execution time) should be made in separate changes

PR #853, opened 2022-11-23
Modinvstor 599

Arghya22:
[MODINVSTOR-599] ( https://issues.folio.org/browse/MODINVSTOR-599 )

## Purpose

- Create enum for persisting settings of Closed Library Date Management in Service Point object 

## Approach

- updated raml of HoldShelfExpiryPeriod 
- test case updated where HoldShelfExpiryPeriod is accepted in api

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1091,7 +1091,7 @@
     },
     {
       "id": "service-points",
-      "version": "3.2",
+      "version": "3.3",
       "handlers": [
         {
           "methods": ["GET"],
--- a/ramls/time-period.json
+++ b/ramls/time-period.json
@@ -18,25 +18,11 @@
         "Months"
       ],
       "default" : "Days"
-    },
-    "closeLibraryDateManagement": {
-      "type": "string",
-      "description": "enum for closeLibraryDateManagement",
-      "enum":[
-        "Keep_the_current_due_date",
-        "Move_to_the_end_of_the_previous_open_day",
-        "Move_to_the_end_of_the_next_open_day",
-        "Keep_the_current_due_date_time",
-        "Move_to_end_of_current_service_point_hours",
-        "Move_to_beginning_of_next_open_service_point_hours"
-      ],
-      "default" : "Keep_the_current_due_date"
     }
   },
   "required": [
     "duration",
-    "intervalId",
-    "closeLibraryDateManagement"
+    "intervalId"
   ],
   "additionalProperties": false
 }
--- a/ramls/time-period.json
+++ b/ramls/time-period.json
@@ -18,11 +18,25 @@
         "Months"
       ],
       "default" : "Days"
+    },
+    "closeLibraryDateManagement": {
+      "type": "string",
+      "description": "enum for closeLibraryDateManagement",
+      "enum":[
+        "Keep_original_date",
+        "Move_to_end_of_previous_open_day",
+        "Move_to_end_of_next_open_day",
+        "Keep_original_date_time",
+        "Move_to_current_service_pt_hours",
+        "Move_to_beginning_of_next_open_service_pt_hours"
+      ],
+      "default" : "Keep_original_date"
     }
   },
   "required": [
     "duration",
-    "intervalId"
+    "intervalId",
+    "closeLibraryDateManagement"
   ],
   "additionalProperties": false
 }
--- a/ramls/time-period.json
+++ b/ramls/time-period.json
@@ -23,14 +23,14 @@
       "type": "string",
       "description": "enum for closeLibraryDateManagement",
       "enum":[
-        "Keep_original_date",
-        "Move_to_end_of_previous_open_day",
-        "Move_to_end_of_next_open_day",
-        "Keep_original_date_time",
-        "Move_to_current_service_pt_hours",
-        "Move_to_beginning_of_next_open_service_pt_hours"
+        "Keep_the_current_due_date",
+        "Move_to_the_end_of_the_previous_open_day",
+        "Move_to_the_end_of_the_next_open_day",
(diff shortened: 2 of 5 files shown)


nirmalr-shah (approved):

ramls/time-period.json
```diff
@@ -18,11 +18,25 @@
...
+      "enum":[
+        "Keep_the_current_due_date",
+        "Move_to_the_end_of_the_previous_open_day",
+        "Move_to_the_end_of_the_next_open_day",
+        "Keep_the_current_due_date_time",
+        "Move_to_end_of_current_service_point_hours",
+        "Move_to_beginning_of_next_open_service_point_hours"
+      ],
+      "default" : "Keep_the_current_due_date"
     }
```
marcjohnson-kint, line 34: I don't think this property should be added to the time period structure. 

The time period structure is intended to be a general representation of time periods and closed library due date management is not an inherent property of a time period. Rather it is a separate concept that might be used to change operations involving a time period.

The description of this change suggests the intention is to associate the closed due date management to a service period. That suggests to me that it should be included in the service point representation not the time period.

Arghya22: Hi @marcjohnson-kint , the intention of putting closeLibraryDateManagement in timeperiod structure , as to include it in holdShelfExpiryPeriod object ( I see the mention raml is time-period.json)  , which eevntually will be part of Service point object .

ramls/time-period.json
```diff
@@ -18,11 +18,25 @@
...
   },
   "required": [
     "duration",
-    "intervalId"
+    "intervalId",
+    "closeLibraryDateManagement"
```
marcjohnson-kint, line 39: A new required property is a compatibility breaking change. 

That should be represented by a change to the major version of the interface(s) affected by the change (in both the module descriptor and the RAML documentation).

And a major version change in the module implementation version (in the pom.xml)

ramls/time-period.json
marcjohnson-kint: I understand that intention. I am suggesting that it could be better to include the property in the service point rather than the time period.

From a domain modelling perspective, this structure is intended to represent a period of time. A period of time is completely unaware of the context in which might be used. 

In this case, part of that context is, that a closed library due date management policy is to be applied to hold shelf expiry at this service point. That is a characteristic of the hold shelf expiry beyond the basic period of time.

Does that make sense?

marcjohnson-kint (dismissed):
Changes to an API should include the relevant changes to the interface version(s) affected.

Please change those versions accordingly (in this case, the major version needs to change because it is a compatibility breaking change).

This also means that any module that depends upon this interface needs to be updated.

Given that a required property is being introduced, I think it's might also be worth applying a step during the module upgrade to set that value in existing records. Otherwise, I'm not sure they can be successfully read.

ramls/time-period.json
marcjohnson-kint: We already have examples of [similar modelling|https://github.com/folio-org/mod-circulation-storage/blob/174ccdd8b5c684e4774d2ffc1c8a4ac1f8f38a2f/ramls/loan-policy.json#L25-L33] in loan policies.

marcjohnson-kint:
Having had a call with @Arghya22, I thought I expand upon my previous comments. I think there are broadly two approaches to adding the new closed due date management property.

The first is an additive approach that is non-breaking. This assumes that the new property is not required. If it is required, then it is also a breaking change.

The second restructures the properties related to hold shelf expiry to reduce duplication and make the structure more explicit.

The second will likely require more code changes to some dependent modules, although those will mostly be small compared to the changes need to use the new property in either form and the overhead of breaking changes.

If it is decided that the new property is not required (and thus no breaking change is required), I would NOT take the restructuring approach, instead preferring the additive approach. Even if the property is required, I would still consider it acceptable to use the additive approach to reduce the changes in downstream modules.

### Non-breaking / Additive Change
```
{ // service point object
  ...
  "holdShelfExpiryPeriod": {
    "type": "object",
    "$ref": "time-period.json",
    "description": "expiration period for items on the hold shelf at the service point"
  },
  "holdShelfClosedLibraryDateManagement": {
    "type": "string",
    "description": "policy for how to change the hold shelf expiration when the library is closed",
    "enum": [
      "Keep_the_current_due_date",
      "Move_to_the_end_of_the_previous_open_day",
      "Move_to_the_end_of_the_next_open_day",
      "Keep_the_current_due_date_time",
      "Move_to_end_of_current_service_point_hours",
      "Move_to_beginning_of_next_open_service_point_hours"
    ],
    "default": "Keep_the_current_due_date"
  }
}
```

### Breaking / Re-structuring Change
```
{ // service point object
  ...
  "holdShelfExpiry": {
    "period": {
      "type": "object",
      "$ref": "time-period.json",
      "description": "expiration period for items on the hold shelf at the service point"
    },
    "closedLibraryDateManagement": {
      "type": "string",
      "description": "policy for how to change the hold shelf expiration when the library is closed",
      "enum": [
        "Keep_the_current_due_date",
        "Move_to_the_end_of_the_previous_open_day",
        "Move_to_the_end_of_the_next_open_day",
        "Keep_the_current_due_date_time",
        "Move_to_end_of_current_service_point_hours",
        "Move_to_beginning_of_next_open_service_point_hours"
      ],
      "default": "Keep_the_current_due_date"
    }
  }
}
```

descriptors/ModuleDescriptor-template.json
```diff
@@ -1091,7 +1091,7 @@
     },
     {
       "id": "service-points",
-      "version": "3.2",
+      "version": "3.3",
```
marcjohnson-kint, line 1094: This value should also be changed in the interface documentation - https://github.com/folio-org/mod-inventory-storage/blob/master/ramls/service-point.raml

arin-epm (approved):

PR #855, opened 2022-11-24
MODINVSTOR-998: Fix Kafka test failure in HoldingsStorageTest:348

julianladisch:
When upgrading to Vert.x 4.3.5 and RMB 35.0.4 we get build failures:

30% of all GitHub Actions builds fail with this message:

```
  HoldingsStorageTest.canReplaceAHoldingAtSpecificLocation:348
    Expected: is "UPDATE"
         but: was "CREATE"
```

Fix this by matching the expected Kafka message at any position in the Kafka message stream, not only at the last position.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1091,7 +1091,7 @@
     },
     {
       "id": "service-points",
-      "version": "3.2",
+      "version": "3.3",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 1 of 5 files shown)


adamdickmeiss (approved):

marcjohnson-kint (approved):
Thank you for making the check not depend upon the last message received. 

(eventually, I'd like to move that to using a collection matcher rather than the exception catching loop, however I understand that is out of scope of this improvement)

PR #857, opened 2022-11-24
MODINVSTOR-1000: ItemStorageTest.canMoveItemToNewInstance Timeout/NPE

julianladisch:
https://issues.folio.org/browse/MODINVSTOR-1000

Use async coordination so that publishUpdated is only called if oldHoldings has been fetched.

Then oldHoldings cannot be null, doesn't throw NPE and the ConditionTimeout won't happen any longer.

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -1091,7 +1091,7 @@
     },
     {
       "id": "service-points",
-      "version": "3.2",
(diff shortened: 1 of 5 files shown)


marcjohnson-kint (approved):
I agree, always fetching the old holdings because it is required for the message publishing leads to this code being more reliable.

marcjohnson-kint:
@julianladisch Thank you for finding this

PR #862, opened 2022-12-01
Use Awaitility Matcher when Creating and Deleting Instances MODINVSTOR-1001

marcjohnson-kint:
In order to remove the need for catching assertion exceptions and improve the stability of checking for published messages, replace the current use of assertions during awaitility with matcher composition.

The use of a `hasItem` matcher for checking the collection of received messages means that the test will pass when any of the messages matches the expectations, irrespective of ordering. This relies on the matching being sufficiently specific to not produce false matches.

As this is a new approach, the changes have been limited to only messages for creating and deleting instances.

### Approach Taken
* Inline current approach for checking for deleted message (for reference during replacement)
* introduce a type to represent an event message (to ease writing matchers rather than manipulating JSON)
* introduce new approach for deletion message inline in a single test, using matchers each assertion at a time
* move matchers into a separate class
* move method for finding messages related to an instance
* apply same approach for checking creation messages (to demonstrate broader applicability)

### DIFF
--- a/src/test/java/org/folio/rest/api/InstanceStorageTest.java
+++ b/src/test/java/org/folio/rest/api/InstanceStorageTest.java
@@ -2,7 +2,6 @@ package org.folio.rest.api;
 
 import static java.net.HttpURLConnection.HTTP_OK;
 import static java.util.concurrent.TimeUnit.SECONDS;
-import static org.awaitility.Awaitility.await;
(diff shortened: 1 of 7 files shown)


kaladay (approved):
Makes sense to me.

It would be great to see some of the work-arounds that I just recently added (namely the timestamp sorting ones) removed. This is a good step in that direction.

PR #877, opened 2023-01-10
MODINVSTOR-1015: LCNAF translation overwritten on upgrade to 25.0.3

julianladisch:
When adding "ns" code to LCNAF authority_source_file entry do not alter any other JSON field.

This avoids overwriting the translation in the name field.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -6,13 +6,13 @@ jobs:
     runs-on: macos-latest
     strategy:
       matrix:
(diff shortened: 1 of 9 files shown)


psmagin:
@julianladisch, after merging this PR, the new release should be created? But how it helps if somebody already upgraded to 25.0.3? It's related to the main release, so I assume that the module will be upgraded to the latest version 25.0.3, and will not have a problem as you described.

psmagin (approved):

PR #888, opened 2023-03-10
MODINVSTOR-1038: Use instanceRepository.getById(id)

julianladisch:
The API

`GET /instance-storage/instances/{id}`

uses a CQL query to get the single instance.

For a primary key lookup no CQL is needed.

Instead we switch to the existing method InstanceInternalRepository#getById.


okolawole-ebsco (approved):
Looks good to me!

PR #901, opened 2023-05-04
MODINVSTOR-1053: Fix of metadata update on call number changes via the holdings record

hasanepam:
## Purpose
Fix metadata update on call number changed via holdings record

## Approach
Added metadata update function

## Learning
[MODINVSTOR-1053](https://issues.folio.org/browse/MODINVSTOR-1053)

### DIFF
--- a/src/main/java/org/folio/services/item/ItemService.java
+++ b/src/main/java/org/folio/services/item/ItemService.java
@@ -162,7 +162,6 @@ public class ItemService {
       .compose(oldHoldings -> {
         putData.oldHoldings = oldHoldings;
         effectiveValuesService.populateEffectiveValues(newItem, putData.newHoldings);
-        populateMetadata(newItem);
         return doUpdateItem(newItem);
       })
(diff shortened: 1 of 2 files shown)


src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -60,6 +61,7 @@ public class ItemService {
     ": Key \\(([^=]+)\\)=\\((.*)\\) already exists.$");
   private static final Pattern KEY_NOT_PRESENT_PATTERN = Pattern.compile(
     ": Key \\(([^=]+)\\)=\\((.*)\\) is not present in table \"(.*)\".$");
+  private static final String OKAPI_USERID_HEADER = "X-Okapi-User-Id";
```
psmagin, line 64: Please use the already defined constant `XOkapiHeaders.USER_ID`

viacheslavkol (approved):

psmagin (approved):

PR #904, opened 2023-05-18
MODINVSTOR-1053: Update item metadata on holdings callNumber changes

hasanepam:
## Purpose
Update item metadata when on holdings callNumber and/or location changes

## Approach
Additional check added before item metadata update

## Learning
[MODINVSTOR-1053](https://issues.folio.org/browse/MODINVSTOR-1053)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -58,6 +58,11 @@
         },
         {
           "methods": ["GET"],
+          "pathPattern": "/authority-storage/reindex",
+          "permissionsRequired": ["authority-storage.authority.reindex.collection.get"]
+        },
+        {
+          "methods": ["GET"],
           "pathPattern": "/authority-storage/reindex/{id}",
           "permissionsRequired": ["authority-storage.authority.reindex.item.get"]
         },
@@ -1312,6 +1317,11 @@
           "methods": ["POST"],
           "pathPattern": "/instance-storage/reindex",
           "permissionsRequired": ["inventory-storage.instance.reindex.post"]
+        },
+        {
+          "methods": ["GET"],
+          "pathPattern": "/instance-storage/reindex",
+          "permissionsRequired": ["inventory-storage.instance.reindex.collection.get"]
         },
         {
           "methods": ["GET"],
--- a/ramls/authorities-reindex.raml
+++ b/ramls/authorities-reindex.raml
@@ -10,11 +10,18 @@ documentation:
 
 types:
   reindexJob: !include reindex/get-job-by-id-response.json
+  reindexJobs: !include reindex/reindex-jobs.json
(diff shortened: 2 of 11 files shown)


src/main/java/org/folio/services/item/ItemService.java
```diff
@@ -228,14 +230,27 @@ public Future<List<Item>> updateItemsOnHoldingChanged(AsyncResult<SQLConnection>
...
+    return isBlank(item.getItemLevelCallNumber()) && isNotBlank(holdingsRecord.getCallNumber())
+      || isBlank(item.getItemLevelCallNumberPrefix()) && isNotBlank(holdingsRecord.getCallNumberPrefix())
+      || isBlank(item.getItemLevelCallNumberSuffix()) && isNotBlank(holdingsRecord.getCallNumberSuffix())
+      || isBlank(item.getItemLevelCallNumberTypeId()) && isNotBlank(holdingsRecord.getCallNumberTypeId())
+      || isNull(item.getPermanentLocationId())
+          && isNull(item.getTemporaryLocationId())
+          && (!isNull(holdingsRecord.getTemporaryLocationId())
+          || !isNull(holdingsRecord.getPermanentLocationId()));
+  }
+
```
psmagin, line 243: I don't this will work properly because if an item already had some call number, and in holdings it was changed then this condition `isBlank(item.getItemLevelCallNumber()) && isNotBlank(holdingsRecord.getCallNumber())` will be `false`. But should be true because item callnumber will be changed.

psmagin (changes requested):

src/main/java/org/folio/services/item/ItemService.java
hasanepam: in [EffectiveCallNumberComponentsUtil.java](https://github.com/folio-org/mod-inventory-storage/blob/master/src/main/java/org/folio/rest/support/EffectiveCallNumberComponentsUtil.java) file there is a method called buildComponents and there it seems to me that holdings call number will be placed in item only when item does not have call number value.

psmagin (approved):

src/main/java/org/folio/services/item/ItemService.java
mreno-EBSCO: Suppose holdings record "hr" has a call number set. "hr" also has item "i" set, but with no item level call number and the metadata for "i" is correct. Suppose "hr" is updated but none of the holdings record fields mentioned in this method are updated. Will the metadata for "i" be updated? Any time "hr" is updated (unless "i" is updated to have the item level call number set), "i".getItemLevelCallNumber() is blank and "hr".getCallNumber() is not blank. It appears to me that we don't actually know, in this method, whether or not any of those holdings record fields have been updated.

I just want to make sure we are not updating item dates when not necessary.

PR #908, opened 2023-05-25
MODINVSTOR-921: API to get the current maximum assigned HRID

julianladisch:
https://issues.folio.org/browse/MODINVSTOR-921

Extend GET /hrid-settings-storage/hrid-settings API to also
return "currentNumber". Currently it only returns "startNumber".

Create hrid_settings_view view that reads hrid_settings record and
merges in last_value from the sequences.

In updateSequence switch to transactional ALTER SEQUENCE RESTART
to allow for rollback.

All changes in non-HRID tests are not regressions caused by the
new code but by the test setup.

### DIFF
--- a/ramls/reindex/get-job-by-id-response.json
+++ b/ramls/reindex/get-job-by-id-response.json
@@ -25,6 +25,15 @@
         "Id publishing cancelled"
       ]
     },
+    "resourceName": {
+      "description": "Reindex resource name",
+      "type": "string",
+      "enum": [
+        "Instance",
+        "Authority",
+        "Unknown"
+      ]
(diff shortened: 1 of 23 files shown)


src/main/resources/templates/db_scripts/hridSettingsView.sql
```diff
@@ -0,0 +1,22 @@
+SELECT 'hridSettingsView.sql';
```
psmagin, line 1: @julianladisch, for what this select is used?

psmagin (approved):
As new field is added to response I believe that interface version should be bumped

PR #910, opened 2023-06-01
MODINVSTOR-990 Create endpoints to retrieve all reindex jobs

ArtHov:
feat(reindex): Create endpoints to retrieve all reindex jobs

- Add resourceName field to ReindexJob
- Add filtration by resourceName for new endpoints

Closes: MODINVSTOR-990

### DIFF
--- a/src/main/java/org/folio/rest/impl/ReindexAuthoritiesApi.java
+++ b/src/main/java/org/folio/rest/impl/ReindexAuthoritiesApi.java
@@ -1,5 +1,6 @@
 package org.folio.rest.impl;
 
+import static org.folio.persist.ReindexJobRepository.AUTHORITY_REINDEX_JOBS_QUERY;
 import static org.folio.rest.persist.PgUtil.get;
 import static org.folio.rest.persist.PgUtil.getById;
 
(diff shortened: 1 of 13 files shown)


hasanepam (approved):

src/main/java/org/folio/rest/impl/ReindexAuthoritiesApi.java
```diff
@@ -41,7 +42,7 @@ public void getAuthorityStorageReindex(String query, int offset, int limit,
                                          Context vertxContext) {
 
     get(ReindexJobRepository.TABLE_NAME, ReindexJob.class, ReindexJobs.class,
-      query, offset, limit, okapiHeaders, vertxContext,
+      AUTHORITY_REINDEX_JOBS_QUERY, offset, limit, okapiHeaders, vertxContext,
```
viacheslavkol, line 45: This way filtration query will be ignored.
I'd be better to remove "query" parameter from api, or compose query and your "type" filter in case "query" should be present by requirements

src/main/java/org/folio/rest/impl/ReindexAuthoritiesApi.java
```diff
@@ -41,8 +42,11 @@ public void getAuthorityStorageReindex(String query, int offset, int limit,
                                          Handler<AsyncResult<Response>> asyncResultHandler,
                                          Context vertxContext) {
 
-    get(ReindexJobRepository.TABLE_NAME, ReindexJob.class, ReindexJobs.class,
-      AUTHORITY_REINDEX_JOBS_QUERY, offset, limit, okapiHeaders, vertxContext,
+    var searchQuery = Objects.isNull(query) ? AUTHORITY_REINDEX_JOBS_QUERY :
+      AUTHORITY_REINDEX_JOBS_QUERY + " and " + query;
```
viacheslavkol, line 46: `AUTHORITY_REINDEX_JOBS_QUERY + " and (" + query + ")"`

PR #912, opened 2023-06-22
MODINVSTOR-1080 Modify inventory dtos to include source field

bekhzodkurbonboev:
Resolves [MODINVSTOR-1080](https://issues.folio.org/browse/MODINVSTOR-1080) Modify inventory dtos to include source field

This table describes what setting we are going to implement in Consortium Manager:
https://wiki.folio.org/display/FOLIJET/Consortium+manager+settings
For _mod-inventory-storage_ they are _loan type_ and _electronic access relationship_. (All other already have 'source' field.)
Setting created in Consortium Manager will have value _'Consortium'_ and we will add validation to prevent updating them

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -312,7 +312,7 @@
     },
     {
       "id": "loan-types",
-      "version": "2.2",
+      "version": "2.3",
       "handlers": [
         {
           "methods": ["GET"],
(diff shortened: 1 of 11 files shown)


bekhzodkurbonboev:
Following is the current status of endpoints, _'source'_ field is not defined for _'loan type'_, but for _'electronic access relationship'_ this field is populated on UI and stored in DB but not defined in the schema: 
image

psmagin (approved):

PR #916, opened 2023-08-02
MODINVSTOR-941: Batch holdings update updates items

julianladisch:
Closes https://issues.folio.org/browse/MODINVSTOR-941 "Location changes in holdings records via batch-synchronous APIs not triggering update to item effective location" Partly addresses https://issues.folio.org/browse/MODINVSTOR-870 "POST/PUT instance/holding/item performance"

* Update items when doing a bulk holdings updates. Item properties to update:
  * effective shelving order
  * effective call number components
  * effective location id
* Database trigger fills holdings HRID if needed
* No performance regression
* Performance improvements

Move all Java code needed to update affected items into SQL functions. Otherwise Java code must fetch and check item records that requires round trips from Java to DB and back to Java and database locks for consistency resulting in an unacceptable performance regression. Performance of this bulk holdings update API is a critical requirement for several institutions.

Move Java code that fetches the next holdings HRID from database into a database trigger. This avoids a holdings record query and an HRID query in Java.

The code changes are minimal so that is can easily been back-ported to Orchid and released as a critical service patch (CSP).

Separate pull requests will remove all Java code that is duplicated in SQL functions; they will not be back-ported as CSPs: https://issues.folio.org/browse/MODINVSTOR-1090
https://issues.folio.org/browse/MODINVSTOR-1091

The algorithm used to calculates the effective value in SQL is exactly the same as the algorithm used in Java.
Many unit tests show that they actually yield the same:

* For effective shelving order this is tested in the new WriteSqlTest#setEffectiveShelvingOrder unit tests, and in the existing ItemStorageTest#canCreateItemEffectiveShelvingOrder tests.
* For effective call number components this is tested in the existing ItemEffectiveCallNumberComponentsTest and ItemEffectiveCallNumberDataUpgradeTest tests.
* For NlmCallNumber this is tested in the new WriteSqlTest#lcNlmCallNumber tests.
* For LCCallNumber this is tested in the new WriteSqlTest#lcNlmCallNumber tests.
* For DeweyCallNumber this is tested in the new WriteSqlTest#deweyCallNumber tests.
* For SuDocCallNumber this is tested in the existing CallNumberUtilsTest#checkSuDocShelvingKey and CallNumberUtilsTest#testSuDocSortingOrder tests.
* For effective location order this is tested in the existing ItemEffectiveLocationTest, ItemStorageTest and HoldingsStorageTest tests.

Please note that the tests for NlmCallNumber, LCCallNumber, DeweyCallNumber and SuDocCallNumber call both the marc4j and the SQL implementation and compare the results. This comparison in the unit tests will be kept forever as explained on [MODINVSTOR-1090](https://issues.folio.org/browse/MODINVSTOR-1090).

Performance improvements:

PUT /holdings-storage/holdings/$id
* When 0 item needs to be updated: From 14 ms to 7 ms
* When 1 item needs to be updated: From 14 ms to 11 ms
* When 10 items need to be updated: From 20 ms to 17 ms

POST /holdings-storage/batch/synchronous?upsert=true
* When 0 item needs to be updated: From 5 ms to 5 ms
* When 1 item needs to be updated: From 5 ms to 8 ms
* When 10 items need to be updated: From 5 ms to 16 ms

Note that the old implementation of the batch API incorrectly doesn't update items, therefore this performance degregation is caused by fixing the bug. And the performance is still better than the PUT single record API.

- [x] Adjust existing tests and add new ones for the the new implementation


src/main/resources/templates/db_scripts/instance-hr-item/write.sql
```diff
@@ -0,0 +1,379 @@
...
+      WHEN '03dd64d0-5626-4ecd-8ece-4531e0069f35' THEN call_number = dewey_call_number(call_number);
+      WHEN '95467209-6d7b-468b-94df-0f5d7ad2747d' THEN call_number = lc_nlm_call_number(call_number);  -- LC
+      WHEN '054d460d-d6b9-4469-9e37-7a78a2266655' THEN call_number = lc_nlm_call_number(call_number);  -- NLM
+      WHEN 'fc388041-6cd0-4806-8a74-ebe3b9ab4c6e' THEN call_number = su_doc_call_number(call_number);
+      ELSE NULL;
+    END CASE;
+    RETURN item || jsonb_build_object('effectiveShelvingOrder',
+        concat_ws(' ', call_number, trim2null(item->'effectiveCallNumberComponents'->>'suffix')));
+  END;
+$$ LANGUAGE plpgsql IMMUTABLE PARALLEL SAFE STRICT;
```
psmagin, line 174: I am concerned about the prospect of duplicating the logic for creating shelving orders across different locations, specifically in database scripts and with the marc4j library. This approach could potentially lead to inconsistencies in shelving order processing, and maintaining consistency is crucial for ensuring accurate call-number search/browse functionality within the inventory app.

psmagin: It's not described, I can't find any solution for this. Maybe I described my concern not well. We could have inconsistency of shelving order processing. As now we are using marc4j for this processing in both mod-search and mod-inventory-storage after this change moving processing to the database level will require us to support the logic in 2 implementations (db and marc4j).

TarasSpashchenko: @julianladisch 
Since we have "creating shelving orders logic" implemented in marc4j and in stored functions in the DB, could you please develop tests that compare results produced by marc4j and stored functions for equality?

psmagin (approved):

PR #920, opened 2023-08-17
[MODINVSTOR-1096] Items with multiple circulation notes that do not have an id cannot be updated

bekhzodkurbonboev:
**Resolves**
[MODINVSTOR-1096](https://issues.folio.org/browse/MODINVSTOR-1096) Items with multiple circulation notes that do not have an id cannot be updated

**Approach**
See Analysis in [MODFISTO-215](https://issues.folio.org/browse/MODFISTO-215) ticket, and [PR](https://github.com/folio-org/mod-finance-storage/pull/214/files):
- Why we need _uuid-ossp extension_
- When we can use _uuid_generate_v4()_ or _uuid_generate_v5()_

### DIFF
--- a/src/main/resources/templates/db_scripts/populateCirculationNoteIds.sql
+++ b/src/main/resources/templates/db_scripts/populateCirculationNoteIds.sql
@@ -22,6 +22,6 @@ WITH item_circnotes as (
   GROUP BY itemId
 )
 UPDATE ${myuniversity}_${mymodule}.item
-SET jsonb = jsonb_set(jsonb, '{circulationNotes}', json_build_array(item_circnotes.circNotes)::jsonb)
+SET jsonb = jsonb_set(jsonb, '{circulationNotes}', to_jsonb(item_circnotes.circNotes))
 FROM item_circnotes
 WHERE item.id = item_circnotes.itemId;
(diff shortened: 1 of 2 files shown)


psmagin (commented):
What is the performance of the script? Do you expect that this script will be finished in 5-10 minutes for 1M, 10M records?
If not then we can proceed with such script.

KaterynaSenchenko:
> What is the performance of the script? Do you expect that this script will be finished in 5-10 minutes for 1M, 10M records? If not then we can proceed with such script.

Hi @psmagin, we'll run the script on perf-rancher and share the results once we have the figures

Aliaksandr-Fedasiuk (approved):

bekhzodkurbonboev:
@psmagin could you please take a look at result of the script on _perf-rancher_:
_NOTICE:  extension "uuid-ossp" already exists, skipping
UPDATE 1_

_Query returned successfully in **2 min 19 secs**._

_SELECT count(id)
FROM fs09000000_mod_inventory_storage.item; -> **10550828**_

MicrosoftTeams-image

psmagin (approved):

yaroslav-epam (approved):

PR #930, opened 2023-09-14
MODINVSTOR-1097 Ad Hoc Shadow Instance Creation During Batch API Operations

tatsiana-tarhonskaya:
## Purpose
[MODINVSTOR-1097](https://issues.folio.org/browse/MODINVSTOR-1097)
Ad Hoc Shadow Instance Creation During Batch API Operations

## Approach
Create Shadow Instance via the consortium sharing process when needing to create a holding on a Shadow Instance that is not present in the local tenant: 

1. Call /user-tenant?limit=1 to get consortiumId and centralTenantId. If configuration is present, it means that tenant belongs to consortium. 
2. If tenant does not belong to consortium, proceed with further steps. If tenant belongs to consortium, check if instance ids exist in the local tenant, if not - invoke sharing API .

### Changes checklist
- [ ] API paths, methods, request or response bodies changed, added, or removed
- [] Database schema changes
- [ ] Interface versions changes
- [ ] Interface dependencies added, or removed
- [ ] Permissions changed, added, or removed
- [ ] Check logging.

### Learning


src/main/java/org/folio/services/holding/HoldingsService.java
```diff
@@ -127,15 +143,20 @@ public Future<Response> deleteHoldings(String cql) {
...
+    Map<String, Future<SharingInstance>> instanceFuturesMap = new HashMap<>();
     for (HoldingsRecord holdingsRecord : holdings) {
       holdingsRecord.setEffectiveLocationId(calculateEffectiveLocation(holdingsRecord));
+      instanceFuturesMap.computeIfAbsent(holdingsRecord.getInstanceId(), this::createShadowInstanceIfNeeded);
+    }
+
+    CompositeFuture instancesFuture = CompositeFuture.all(new ArrayList<>(instanceFuturesMap.values()));
+
+    if (upsert) {
+      return upsertHoldings(holdings, instancesFuture, optimisticLocking);
```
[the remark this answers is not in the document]
tatsiana-tarhonskaya: Thanks, updated

src/main/java/org/folio/services/holding/HoldingsService.java
```diff
@@ -216,6 +234,30 @@ private String calculateEffectiveLocation(HoldingsRecord holdingsRecord) {
     }
   }
 
+  private Future<SharingInstance> createShadowInstanceIfNeeded(String instanceId) {
```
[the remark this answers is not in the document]
tatsiana-tarhonskaya: Moved this check to be done prior to instance retrieving

okolawole-ebsco (approved):

src/main/java/org/folio/services/holding/HoldingsService.java
```diff
@@ -127,30 +143,34 @@ public Future<Response> deleteHoldings(String cql) {
   }
 
   public Future<Response> createHoldings(List<HoldingsRecord> holdings, boolean upsert, boolean optimisticLocking) {
-    if (upsert) {
-      return upsertHoldings(holdings, optimisticLocking);
-    }
-
     for (HoldingsRecord holdingsRecord : holdings) {
```
okolawole-ebsco, line 146: In the previous version of the code, effective location is not set when upsert is involved. In the current version, effective location is always set. Will this break something?

tatsiana-tarhonskaya: It was set inside of upsertHoldings: https://github.com/folio-org/mod-inventory-storage/pull/930/files#diff-78dff0242682c91208ab3b3ce644115ee6d3cb522c0697639db65927241547feL150
Now it's just set in a single place for both upsert and non upsert

okolawole-ebsco: Ah. good catch!

okolawole-ebsco: Actually a thought occurred to me. upsertHoldings is a public method so it may be called by itself. If it is called by itself, effective location will not be set. What do you think?

tatsiana-tarhonskaya: Good point. As I can see upsertHoldings  was added recently and is called only from createHoldings for now. Not sure if there were plans to use it somewhere else. Probably it could be just made private? 

Or if it should remain public, then it should contain not only setting effective locations, but also creating shadow instances, as in createHoldings, right?  

@julianladisch , what are your thoughts on this?

RomanChernetskyi (approved):

yaroslav-epam (approved):

src/main/java/org/folio/services/holding/HoldingsService.java
okolawole-ebsco: the name upsertHoldings tells me that holdings can be created as well as updated. I would be okay to leave as is if only updates were done; that means the holding is already under an instance. So lets keep the original intent of author of this method and keep it public and add setting effective locations and creating shadow instances.

PR #933, opened 2023-09-21
MODINVSTOR-1076 - Shadow Instance Synchronization

RuslanLavrov:
## Purpose
When shared instances are updated, it is necessary to update/synchronize the corresponding shadow instances on the member tenants as well, so that they reflect the same changes as the shared instances

## Approach
* add a consumer subscribed to the event about instance update
* implement event handler that performs shadow instance update/synchronization on the member tenants upon receiving an event about shared instance update on the central tenant
* add tests

## Learning
[MODINVSTOR-1076](https://issues.folio.org/browse/MODINVSTOR-1076)
requires: https://github.com/folio-org/folio-kafka-wrapper/pull/45


src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private Future<Void> updateShadowInstance(Instance instance, String tenantId, Map<String, String> okapiHeaders) {
+    LOG.debug("updateShadowInstance:: Trying to update shadow instance, tenantId: '{}', instanceId: '{}'",
+      tenantId, instance.getId());
+    HashMap<String, String> headers = new HashMap<>(okapiHeaders);
+    headers.put(TENANT, tenantId);
+    InstanceRepository instanceRepository = new InstanceRepository(vertx.getOrCreateContext(), headers);
+
+    return instanceRepository.getById(instance.getId())
```
okolawole-ebsco, line 156: Can we have retries when the update is blocked by optimistic locking? or do you think that is unnecessary and is not a pattern used in other parts of inventory?

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  private Future<Void> updateShadowInstances(DomainEvent<Instance> event, List<String> tenantIds,
+                                             Map<String, String> headers) {
+    LOG.info("updateShadowInstances:: Trying to update shadow instances in the following tenants: {} ", tenantIds);
+    ArrayList<Future<Void>> updateFutures = new ArrayList<>();
+    Instance instance = JsonObject.mapFrom(event.getNewEntity()).mapTo(Instance.class);
+    prepareInstanceForUpdate(instance);
+
+    for (String tenantId : tenantIds) {
```
okolawole-ebsco, line 139: This has the potential to launch numerous call to update instances at the same time. Is it possible to constraint it?

okolawole-ebsco: Maybe updating instances sequentially on a worker thread.

src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
```diff
@@ -179,6 +180,10 @@ private Future<Void> publish(String key, Object value) {
 
     KafkaProducer<String, String> producer = getOrCreateProducer();
 
+    if (okapiHeaders.get(XOkapiHeaders.TOKEN.toLowerCase()) != null) {
```
okolawole-ebsco, line 183: Does this mean shadow instances are updated with the same token that updated the shared instance? If so, what if the user that updated the shared instance does not have an affiliation to a shadow instance's tenant?

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
RuslanLavrov: it might be unnecessary because updating shadow instances through the UI or data import is not expected. There is a probability of encountering optimistic locking error during a shared instance update using ui, but in this case, the shared instance update will fail and this handler will not be called.

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
RuslanLavrov: ShadowInstanceSynchronizationVerticle is deployed as a worker verticle so it’s already executed using a thread from vertx worker thread pool (options.setWorker(true) deployment option is set in the InitApi class).

src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
RuslanLavrov: yes correct, this token will be received by this handler. It will work fine because the handler using this token performs http requests to central tenant. Just double-checked this case on the rancher env that it works as expected.

okolawole-ebsco: can the token perform http requests to member tenants if the user of the token does not have an affiliation?

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
okolawole-ebsco: I see. Makes it better that my original thought. There is still a gap in that the database connections can be exhausted when large manipulations occur with instances. Firing off 61 requests would mean potentially a maximum of 61 requests for a database connection. More than likely mod-inventory-storage is working while a data import job is running so the connection pool will be exhausted and might see error. Do you see my scenario?

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
okolawole-ebsco: Sounds good!

src/main/java/org/folio/services/domainevent/CommonDomainEventPublisher.java
okolawole-ebsco: nevermind! I was thinking we use the token to update shadow instance. But this is inventory-storage, a token is only used to get the latest central tenant instance.

src/main/java/org/folio/services/migration/async/AbstractAsyncMigrationJobRunner.java
```diff
@@ -79,10 +80,10 @@ private Future<Long> processStream(StreamingContext context) {
   }
 
   private KafkaProducerRecordBuilder<String, Object> rowToProducerRecord(Row row, StreamingContext context) {
-    return new KafkaProducerRecordBuilder<String, Object>()
+    String tenantId = tenantId(context.getMigrationContext().getOkapiHeaders());
+    return new KafkaProducerRecordBuilder<String, Object>(tenantId)
       .key(row.getUUID("id").toString())
-      .value(
-        asyncMigrationEvent(context.getJob(), TenantTool.tenantId(context.getMigrationContext().getOkapiHeaders())))
+      .value(asyncMigrationEvent(context.getJob(), tenantId))
```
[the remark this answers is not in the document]
RuslanLavrov: yes sure, thanks for pointing that @julianladisch!

src/main/java/org/folio/rest/impl/InitApiImpl.java
```diff
@@ -19,16 +20,11 @@ public class InitApiImpl implements InitAPI {
...
-      })
-      .onFailure(th -> {
-        handler.handle(Future.failedFuture(th));
-        log.error("Consumer Verticles were not started", th);
-      });
     initConsortiumDataCache(vertx, context);
+    initAsyncMigrationVerticle(vertx)
+      .compose(v -> initShadowInstanceSynchronizationVerticle(vertx, getConsortiumDataCache(context)))
+      .onComplete(v -> handler.handle(Future.succeededFuture()))
+      .onFailure(th -> handler.handle(Future.failedFuture(th)));
```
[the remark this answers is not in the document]
RuslanLavrov: onComplete() good catch! Saw it as well but forgot to fix and it left as was.

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
RuslanLavrov: got it, I see. Was thinking about it. Each request for a shadow instance update uses connection from a separate connection pool of the particular member tenant. So each shadow instance update will use one connection from pool of particular member tenant and because of that pool for a particular member tenant should not be exhausted.

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationVerticle.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Override
+  public void start(Promise<Void> startPromise) throws Exception {
+    HttpClient httpClient = vertx.createHttpClient();
+    ShadowInstanceSynchronizationHandler handler =
+      new ShadowInstanceSynchronizationHandler(consortiumDataCache, httpClient, vertx);
+
+    createKafkaConsumerWrapper(handler)
+      .onFailure(startPromise::fail)
+      .onSuccess(v -> startPromise.complete());
```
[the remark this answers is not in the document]
RuslanLavrov: can you say what the significant benefits of such refactoring?

src/test/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandlerTest.java
```diff
@@ -0,0 +1,260 @@
...
+    Future<Instance> future = synchronizationHandler.handle(kafkaRecord)
+      .onComplete(ar -> context.assertTrue(ar.succeeded()))
+      .compose(v -> getInstanceById(sharedInstance.getId(), TENANT_ID));
+
+    future.onComplete(ar -> {
+      context.assertTrue(ar.succeeded());
+      Instance updatedShadowInstance = ar.result();
+      context.assertEquals(sharedInstance.getTitle(), updatedShadowInstance.getTitle());
+      async.complete();
+    });
```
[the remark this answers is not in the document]
RuslanLavrov: yes, I know this testContext api

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
okolawole-ebsco: For MOBIUS, we are planning to shared db pools. does that change anything?

RuslanLavrov: pls clarify does it mean one common shared pool for tenants or multiple pools (1 per tenant) that share connections between tenants?

KaterynaSenchenko (approved):

src/main/java/org/folio/services/consortia/ShadowInstanceSynchronizationHandler.java
okolawole-ebsco: One common pool for all tenants in a module.

RuslanLavrov: got it, added changes that allow updating not more than 10 instances (by default) in parallel.
What max pool size is planned to be set for such pool?

okolawole-ebsco: I don't know what the max pool size will be. Concrete values should be gotten from performance testing

src/main/java/org/folio/services/consortium/ShadowInstanceSynchronizationHandler.java
```diff
@@ -0,0 +1,185 @@
...
+  private Future<Void> updateShadowInstances(DomainEvent<Instance> event, List<String> tenantIds,
+                                             Map<String, String> headers) {
+    LOG.info("updateShadowInstances:: Trying to update shadow instances in the following tenants: {} ", tenantIds);
+    Instance instance = JsonObject.mapFrom(event.getNewEntity()).mapTo(Instance.class);
+    prepareInstanceForUpdate(instance);
+    List<List<String>> tenantsChunks = Lists.partition(tenantIds, instancesParallelUpdatesLimit);
+
+    Future<CompositeFuture> future = Future.succeededFuture();
+    for (List<String> tenantsChunk : tenantsChunks) {
+      future = future.compose(v -> updateShadowInstances(tenantsChunk, instance, headers));
```
okolawole-ebsco, line 148: if one update to a shadow instance fails, it will abort the chain. Update `updateShadowInstances` to prevent this.

src/main/java/org/folio/services/consortium/ShadowInstanceSynchronizationHandler.java
```diff
@@ -0,0 +1,185 @@
...
+      return Future.succeededFuture(affiliationsTenantIds);
+    });
+  }
+
+  private Future<Void> updateShadowInstances(DomainEvent<Instance> event, List<String> tenantIds,
+                                             Map<String, String> headers) {
+    LOG.info("updateShadowInstances:: Trying to update shadow instances in the following tenants: {} ", tenantIds);
+    Instance instance = JsonObject.mapFrom(event.getNewEntity()).mapTo(Instance.class);
+    prepareInstanceForUpdate(instance);
+    List<List<String>> tenantsChunks = Lists.partition(tenantIds, instancesParallelUpdatesLimit);
```
okolawole-ebsco, line 144: I like how this is done!

src/main/java/org/folio/services/consortium/ShadowInstanceSynchronizationHandler.java
RuslanLavrov: updated

okolawole-ebsco (approved):

src/main/java/org/folio/services/consortium/ShadowInstanceSynchronizationHandler.java
RuslanLavrov: it will work in this way as well because .join in the updateShadowInstances waits for the completion of all single instance updates

PR #937, opened 2023-09-29
MODINVSTOR-1103 - Ad hoc Shadow Instance creation when adding a Holdings to a Shared Instance

RuslanLavrov:
## Purpose
to provide the creation of corresponding shadow instance upon creation of holdings record on a shared instance

## Approach
* create a shadow instance on the local tenant through the consortium sharing process if such instance doesn't exist with the instance id specified in the received holding to be created
* add and update tests

## Learning
[MODINVSTOR-1103](https://issues.folio.org/browse/MODINVSTOR-1103)

### DIFF
--- a/descriptors/ModuleDescriptor-template.json
+++ b/descriptors/ModuleDescriptor-template.json
@@ -117,7 +117,11 @@
         {
           "methods": ["POST"],
           "pathPattern": "/holdings-storage/batch/synchronous",
-          "permissionsRequired": ["inventory-storage.holdings.batch.post"]
+          "permissionsRequired": ["inventory-storage.holdings.batch.post"],
+          "modulePermissions": [
+            "user-tenants.collection.get",
+            "consortia.sharing-instances.item.post"
+          ]
         }
       ]
     },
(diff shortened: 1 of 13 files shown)


KaterynaSenchenko (approved):

descriptors/ModuleDescriptor-template.json
```diff
@@ -90,7 +90,11 @@
         }, {
           "methods": ["POST"],
           "pathPattern": "/holdings-storage/holdings",
-          "permissionsRequired": ["inventory-storage.holdings.item.post"]
+          "permissionsRequired": ["inventory-storage.holdings.item.post"],
+          "modulePermissions": [
+            "user-tenants.collection.get",
+            "consortia.sharing-instances.item.post"
```
KaterynaSenchenko, line 96: I assume, there won't be any problems in case this permission does not exist, right?

RuslanLavrov: yes

okolawole-ebsco (approved):

PR #946, opened 2023-10-20
MODINVSTOR-1116, MODINVSTOR-1115: Fix reference and sample data

julianladisch:
Fix call number types reference data and instance relationships sample data so that InstallUpgradeIT succeeds.

This also fixes MODINVSTOR-1115:
Check "error" in GET /_/tenant/<tenantid> in InstallUpgradeIT

https://issues.folio.org/browse/MODINVSTOR-1116
https://issues.folio.org/browse/MODINVSTOR-1115

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,6 +10,7 @@
 * Description ([ISSUE_NUMBER](https://issues.folio.org/browse/ISSUE_NUMBER))
 
 ### Bug fixes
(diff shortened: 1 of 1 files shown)


psmagin (approved):
Thanks for the PR @julianladisch. Please update NEWS file

mukhiddin-yusuf (approved):

PR #954, opened 2024-01-15
add logger

GeloPakDev1:


src/main/java/org/folio/services/ResponseHandlerUtil.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public static Response handleResponse(Response response) {
+    var errorMessage = response.getEntity().toString();
+    if (errorMessage.contains(HRID_ERROR_MESSAGE) && (errorMessage.contains("instance"))) {
+      return createResponse(errorMessage);
+    } else {
+      return response;
+    }
+  }
```
viacheslavkol, line 21:
```suggestion
  public static Response handleInstanceHridError(Response response) {
    if (response.getStatus() != 400) {
      return response;
    }

    var errorMessage = response.getEntity().toString();
    if (errorMessage.contains(HRID_ERROR_MESSAGE) && errorMessage.contains("instance")) {
      errorMessage = errorMessage.replace(HRID_ERROR_MESSAGE, HRID);
      return textPlainResponse(400, errorMessage);
    }

    return response;
  }
```

src/main/java/org/folio/services/ResponseHandlerUtil.java
```diff
@@ -0,0 +1,30 @@
...
+import javax.ws.rs.core.Response;
+
+public final class ResponseHandlerUtil {
+  private static final String HRID_ERROR_MESSAGE = "lower(f_unaccent(jsonb ->> 'hrid'::text))";
+  private static final String HRID = "HRID";
+
+  private ResponseHandlerUtil() {
+  }
+
+  public static Response handleResponse(Response response) {
```
viacheslavkol, line 14: It'd be better also to rename this method because it specifically designed to modify istance hrid error response message. See previous messages for a name suggestion

psmagin:
I suggest also to have this logic in holdings and items. And also check bulk endpoints.

src/main/java/org/folio/services/ResponseHandlerUtil.java
```diff
@@ -0,0 +1,60 @@
...
+      return response;
+    }
+
+    var errorMessage = getErrorMessage(response.getEntity());
+    if (errorMessage.contains(HRID_ERROR_MESSAGE)
+      && errorMessage.contains("instance") && statusCode == 400) {
+      return createResponse(400, errorMessage);
+    } else if (errorMessage.contains(HRID_ERROR_MESSAGE)
+      && (errorMessage.contains("item") || errorMessage.contains("holdings_record")) && statusCode == 422) {
+      return createResponse(422, errorMessage);
```
viacheslavkol, line 28: You could pass original Response object, to avoid duplicate/pass statuses and additional conditions for content type.
It'd be good also to keep original object for 422 (with error parameters) and replace only desired message

src/main/java/org/folio/services/ResponseHandlerUtil.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  public static Response failedValidationResponse(Response response) {
+    var entity = (Errors) response.getEntity();
+    var errors = entity.getErrors();
+    errors.get(0).setMessage(errors.get(0).getMessage().replace(HRID_ERROR_MESSAGE, HRID));
+    entity.setErrors(errors);
+    return Response.fromResponse(response)
+      .entity(entity)
+      .header(CONTENT_TYPE, "application/json")
```
viacheslavkol, line 55: Doesn't original response have this header? Do you really need to specify it directly?

src/main/java/org/folio/services/ResponseHandlerUtil.java
```diff
@@ -0,0 +1,58 @@
...
+      errorMessage = errors.getErrors().get(0).getMessage();
+    } else if (responseEntity instanceof String message) {
+      errorMessage = message;
+    }
+    return errorMessage;
+  }
+
+  private static Response createResponse(Response response) {
+    if (response.getStatus() == 400) {
+      return Response.fromResponse(response).header(CONTENT_TYPE, "text/plain")
```
viacheslavkol, line 41: Doesn't original response have this header? Do you really need to specify it directly?

viacheslavkol (approved):

TsaghikKhachatryan (approved):

PR #1029, opened 2024-05-29
[MODINVSTOR-1216] - Implement Domain Event Production for Library Create/Update/Delete

azizbekxm:
### Purpose
https://folio-org.atlassian.net/browse/MODINVSTOR-1216
### Approach
Implemented Domain Event Production for Library Create/Update/Delete

### Result 

telegram-cloud-photo-size-2-5384489242618618046-y
telegram-cloud-photo-size-2-5384489242618618035-y
telegram-cloud-photo-size-2-5384489242618618047-y
telegram-cloud-photo-size-2-5384489242618618048-y
telegram-cloud-photo-size-2-5384489242618618054-y
telegram-cloud-photo-size-2-5384489242618618116-y
telegram-cloud-photo-size-2-5384489242618618119-y
telegram-cloud-photo-size-2-5384489242618618124-y

#### Messages data
telegram-cloud-photo-size-2-5384489242618618125-y


checkstyle/checkstyle.xml
```diff
@@ -269,12 +269,6 @@
     </module>
     <module name="OverloadMethodsDeclarationOrder"/>
     <module name="VariableDeclarationUsageDistance"/>
-    <module name="CustomImportOrder">
-      <property name="sortImportsInGroupAlphabetically" value="true"/>
-      <property name="separateLineBetweenGroups" value="true"/>
-      <property name="customImportOrderRules" value="STATIC###THIRD_PARTY_PACKAGE"/>
-      <property name="tokens" value="IMPORT, STATIC_IMPORT, PACKAGE_DEF"/>
-    </module>
```
azizbekxm, line 277: removing import orders check style configuration. 
Reason: 
 1. It is taking extra effort to fix every single file import of orders
 2. Import of order is not big deal to understand code, what matter is code itself
 3. This checkstyle is not even compatible with official folio documentation
 4. In these days, it is better to trust IDEA, so more focus one code itself.
 
If you have strong reason to keep this, please give, or it will be removed.

Note: Only order of imports is being deleted. not deleting unused import check and wildcard style

psmagin: Please revert this change as it verifies the code style of the repository.

azizbekxm: it is not about code style, removing part is for import order style. Is there any reason to keep order of imports?

SerhiiNosko: We should follow order of imports that owners of this module define

psmagin: Reasons to have it:

- Merge Conflict Reduction: Consistent import ordering reduces the likelihood of merge conflicts. When multiple developers work on the same file, inconsistent import orders can lead to unnecessary conflicts that could have been avoided.

- Tool Compatibility: While modern IDEs like IntelliJ IDEA offer tools to manage imports automatically, not all team members might use the same IDE or have the same configuration. A standardized import order ensures consistency across different development environments.

- Code Quality and Maintenance: Maintaining a structured import order is part of adhering to coding standards that promote high-quality code. It ensures that the codebase remains maintainable and comprehensible, especially as the project grows.

azizbekxm: 1. Solving Merge conflict for imports are way easier than fix imports one by one based on checkstyle.
2. I don't think many people who contribute this repo in this year, are not using Intelij IDEA. even if you use other tool still they can configure their ide instead of many intellij idea user configure.
3. :) 

@psmagin basically, are you totally okay, if someone come from other team and try to contribute and spend extra 10-20 minut to fix import checkstyle issue.

Why I am saying is that, I saw this same checkstyle in other module and spend extra time on to fix just import orders which turn out be different from other modules. We have similar two ticket to contribute this module, I am trying to avoid other people to spend useless time to fix import-order style. 

If you still thinking to revert this based on these informations, please update your checkstyle based on documentation, for us easier to use same imports style for all modules.
https://folio-org.atlassian.net/wiki/spaces/FOLIJET/pages/1378531/Java+code+formatter+for+IDEs

BKadirkhodjaev: A small side note, you need to install CheckStyle-IDEA plugin into the IDEA first

azizbekxm: @psmagin thanks for sharing I will use definetly

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,153 @@
...
+    return PgUtil.post(LIBRARY_TABLE, loclib, okapiHeaders, context, PostLocationUnitsLibrariesResponse.class)
+      .onSuccess(domainEventService.publishCreated());
+  }
+
+  public Future<Response> update(String id, Loclib loclib) {
+    if (loclib.getId() == null) {
+      loclib.setId(id);
+    }
+    var checks = List.of(checkIdChange(id, loclib));
+    return doLibraryCheck(checks)
```
psmagin, line 66: I don't understand why, for a simple check in the service, you need to:

- Create a method doLibraryCheck with a List as a parameter.
- Create a special exception that you then convert into an Errors class object.

It would be better to perform the check and, if it fails, immediately return a completed future with an Errors response

psmagin (changes requested):
Also please update NEWS file

src/main/java/org/folio/InventoryKafkaTopic.java
```diff
@@ -27,7 +28,8 @@ public enum InventoryKafkaTopic implements KafkaTopic {
    */
   private static final Map<InventoryKafkaTopic, Pair<String, String>> TOPIC_PARTITION_MAP = Map.of(
     CLASSIFICATION_TYPE, Pair.of("KAFKA_CLASSIFICATION_TYPE_TOPIC_NUM_PARTITIONS", "1"),
-    LOCATION, Pair.of("KAFKA_LOCATION_TOPIC_NUM_PARTITIONS", "1")
+    LOCATION, Pair.of("KAFKA_LOCATION_TOPIC_NUM_PARTITIONS", "1"),
+    LIBRARY, Pair.of("KAFKA_LIBRARY_TOPIC_NUM_PARTITIONS", "1")
```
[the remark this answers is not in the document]
azizbekxm: There is no this section. And now env variables for KAFKA_LOCATION_TOPIC_NUM_PARTITIONS, KAFKA_CLASSIFICATION_TYPE_TOPIC_NUM_PARTITIONS
image
please clarify, this comment

src/main/java/org/folio/services/locationunit/LibraryService.java
azizbekxm: First I did based on LocationApi pr, removed this part and do much improvements based on you feedback. Thanks.

src/main/java/org/folio/rest/impl/LocationUnitApi.java
```diff
@@ -302,132 +303,64 @@ public void getLocationUnitsLibraries(String query, String totalRecords, int off
...
-            asyncResultHandler.handle(Future.succeededFuture(
-              PostLocationUnitsLibrariesResponse.respond500WithTextPlain(message)));
-          }
-        } else {
-          String responseObject = reply.result();
-          entity.setId(responseObject);
-          asyncResultHandler.handle(Future.succeededFuture(
-            PostLocationUnitsLibrariesResponse
-              .respond201WithApplicationJson(entity,
-                PostLocationUnitsLibrariesResponse.headersFor201().withLocation(URL_PREFIX + responseObject))));
```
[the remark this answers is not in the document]
azizbekxm: I checked the logic in there. it turned out to be not correct approach for example we will get 422 when campus Id is not valid and object already exists. This logic putting just checking error already exists or not. Added many test cases and it is working great.

src/main/java/org/folio/rest/impl/LocationUnitApi.java
```diff
@@ -302,132 +303,64 @@ public void getLocationUnitsLibraries(String query, String totalRecords, int off
...
-      .save(LIBRARY_TABLE, id, entity, reply -> {
-        if (reply.failed()) {
-          String message = StorageHelper.logAndSaveError(reply.cause());
-          if (StorageHelper.isDuplicate(message)) {
-            asyncResultHandler.handle(Future.succeededFuture(
-              PostLocationUnitsLibrariesResponse
-                .respond422WithApplicationJson(
-                  ValidationHelper.createValidationErrorMessage(
-                    "loclib", entity.getId(),
-                    "Library already exists"))));
```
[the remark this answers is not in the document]
azizbekxm: I added integration tests cases for possible cases, PgUtil correctly handling exception cases

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,153 @@
...
+import org.folio.rest.jaxrs.resource.LocationUnits.DeleteLocationUnitsLibrariesByIdResponse;
+import org.folio.rest.jaxrs.resource.LocationUnits.DeleteLocationUnitsLibrariesResponse;
+import org.folio.rest.jaxrs.resource.LocationUnits.GetLocationUnitsLibrariesByIdResponse;
+import org.folio.rest.jaxrs.resource.LocationUnits.GetLocationUnitsLibrariesResponse;
+import org.folio.rest.jaxrs.resource.LocationUnits.PostLocationUnitsLibrariesResponse;
+import org.folio.rest.jaxrs.resource.LocationUnits.PutLocationUnitsLibrariesByIdResponse;
+import org.folio.rest.persist.PgUtil;
+import org.folio.services.domainevent.LibraryDomainEventPublisher;
+
+public class LibraryService {
```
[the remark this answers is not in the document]
azizbekxm: Already added many integration test, reaching good code coverage. I don't see any reason to write unit tests that can cover places that integration test didn't cover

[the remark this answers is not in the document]
psmagin: The reason for having more unit tests and fewer integration tests is that integration tests are heavy and increase build time a lot as they include the setup of Kafka and PostgreSQL, enabling tenants in the module, and so on. On the other hand unit tests are lightweight.
So, the main purpose of integration tests is to cover critical paths, and unit tests to cover all edge cases.

src/main/java/org/folio/rest/impl/LocationUnitApi.java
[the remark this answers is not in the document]
psmagin: I'm asking not about error handling exceptions or validation but about setting the Location header on POST as it was implemented in the highlighted code.

src/main/java/org/folio/InventoryKafkaTopic.java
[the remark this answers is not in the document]
psmagin: It's okay. Please also add it to the module descriptor.

src/test/java/org/folio/rest/impl/LocationUnitLibraryIT.java
```diff
@@ -0,0 +1,205 @@
...
+  void post_shouldReturn422_whenObjectIsDuplicate(Vertx vertx,
+                                                  VertxTestContext ctx) {
+    HttpClient client = vertx.createHttpClient();
+    String libraryId = UUID.randomUUID().toString();
+    var library = sampleRecord().withId(libraryId);
+
+    doPost(client, resourceUrl(), pojo2JsonObject(library))
+      .onComplete(verifyStatus(ctx, HTTP_CREATED));
+
+    doPost(client, resourceUrl(), pojo2JsonObject(library))
```
psmagin, line 135: There you are doing 2 posts, that will work in parallel. So, there could be a case when second post will be completed before 1st one and test will fail.
You should do 2nd post after 1st is completed.

src/main/java/org/folio/InventoryKafkaTopic.java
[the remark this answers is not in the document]
azizbekxm: Thanks, added

src/test/java/org/folio/rest/impl/LocationUnitLibraryIT.java
azizbekxm: Good point, Fixed

src/main/java/org/folio/services/locationunit/LibraryService.java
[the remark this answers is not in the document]
azizbekxm: Fair point, unit tests + karate are good for mod-inventory-storage. In addition, unit tests should test logic of service, but LibraryService doesn't contain logic, only interaction with DB, playing DAO role in here except for one edge checking id mistmach in put operation. It is already covered in LocationUnitLibraryIT test. I also couldn't find any example unit test for service class that contains PgUtil methods like LibraryService

src/main/java/org/folio/InventoryKafkaTopic.java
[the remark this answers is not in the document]
psmagin: Still cannot see updated module descriptor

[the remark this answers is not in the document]
azizbekxm: Both README and Module desciprtor update, thanks

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  public Future<Response> create(Loclib loclib) {
+    return PgUtil.post(LIBRARY_TABLE, loclib, okapiHeaders, context,
+        PostLocationUnitsLibrariesResponse.class)
+      .onSuccess(response ->
+        PostLocationUnitsLibrariesResponse
+          .respond201WithApplicationJson(response,
+            PostLocationUnitsLibrariesResponse.headersFor201()
+              .withLocation(URL_PREFIX + response)))
```
SerhiiNosko, line 63: should we invoke domainEventService.publishCreated() here?

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,116 @@
...
+          .onSuccess(domainEventService.publishUpdated(oldLoclib))
+          .otherwise(throwable ->
+            PostLocationUnitsLibrariesResponse.respond500WithTextPlain(
+              throwable.getMessage()))
+      );
+  }
+
+  public Future<Response> delete(String id) {
+    return repository.getById(id)
+      .compose(oldLocation ->
```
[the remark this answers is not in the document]
azizbekxm: Thanks, fixed

src/main/java/org/folio/services/locationunit/LibraryService.java
azizbekxm: Good catch Thanks, fixed

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,117 @@
...
+
+  public Future<Response> getById(String id) {
+    return PgUtil.getById(LIBRARY_TABLE, Loclib.class, id, okapiHeaders, context,
+      GetLocationUnitsLibrariesByIdResponse.class);
+  }
+
+  public Future<Response> create(Loclib loclib) {
+    return PgUtil.post(LIBRARY_TABLE, loclib, okapiHeaders, context,
+        PostLocationUnitsLibrariesResponse.class)
+      .onSuccess(response -> {
```
SerhiiNosko, line 59: map() would be more suitable here instead of onSuccess(), because in our current case we just return future with PgUtil.post result and onSuccess is executed independently. For sending message to kafka it is okay to use onSuccess because we should send to kafka asyncronusly.
See ItemService as example.

src/main/java/org/folio/services/locationunit/LibraryService.java
```diff
@@ -0,0 +1,117 @@
...
+        PostLocationUnitsLibrariesResponse.respond201WithApplicationJson(response,
+            PostLocationUnitsLibrariesResponse.headersFor201()
+              .withLocation(URL_PREFIX + response));
```
azizbekxm, line 62: @psmagin Could you verfiy this
Previously location setting this value 
image

Without configuration it will be just UUID. 

/location-units879d4440-1aea-470c-94e9-34f7ae34f631 this value in header is being used in anywhere on purpose. It is confusing string to me

azizbekxm: I am removing this custom location set. do you agree

SerhiiNosko (approved):

BKadirkhodjaev (approved):

psmagin (changes requested):
NEWS still not updated

SerhiiNosko:
Updated NEWS

Saba-Zedginidze-EPAM (approved):

psmagin (approved):

PR #1062, opened 2024-08-16
MODINVSTOR-1225 - Implement Endpoint for Bulk Instances Upsert from External File

RuslanLavrov:
## Purpose
provide endpoint to save re-mapped instances from external file during MARC migration flow initiated for instances

## Approach
* the `POST /instance-storage/instances/bulk` endpoint has been implemented to upsert instances in bulk from a file located on S3-compatible storage. The endpoint's response includes the number of errors that occurred during processing, along with the names of the files containing the errors and failed instances, if any errors occurred. In case of occurrence processing errors the files with errors and failed instances are uploaded to external S3-compatible storage.
* added tests

## Learning
[MODINVSTOR-1225](https://issues.folio.org/browse/MODINVSTOR-1225)


src/main/java/org/folio/services/instance/InstanceS3Service.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  private Future<Void> ensureInstancesWithNonMarcControlledFields(List<Instance> instances) {
+    return instanceRepository.getById(instances, Instance::getId).map(existingInstances -> {
+      instances.forEach(instance -> {
+        if (existingInstances.get(instance.getId()) != null) {
+          InstanceUtil.copyNonMarcControlledFields(instance, existingInstances.get(instance.getId()));
+        }
+      });
+      return null;
```
Aliaksandr-Fedasiuk, line 148: Perhaps it would be better if we returned Future.succeededFuture() instead of null?

src/main/java/org/folio/rest/support/BulkProcessingErrorFileWriter.java
```diff
@@ -0,0 +1,91 @@
...
+import io.vertx.core.file.AsyncFile;
+import io.vertx.core.file.OpenOptions;
+import io.vertx.core.json.Json;
+import java.nio.file.Path;
+import java.util.function.Function;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.folio.services.BulkProcessingContext;
+
+public class BulkProcessingErrorFileWriter {
```
VRohach, line 15: @RuslanLavrov  Let's add documentation for this class for a better understanding of the main purpose of this entity.

src/main/java/org/folio/rest/support/InstanceUtil.java
```diff
@@ -0,0 +1,265 @@
...
+import org.folio.rest.jaxrs.model.Instance;
+import org.folio.rest.jaxrs.model.Instance.SourceRecordFormat;
+import org.folio.rest.jaxrs.model.InstanceNote;
+import org.folio.rest.jaxrs.model.Publication;
+import org.folio.rest.jaxrs.model.PublicationPeriod;
+import org.folio.rest.jaxrs.model.Series;
+import org.folio.rest.jaxrs.model.Subject;
+import org.folio.rest.jaxrs.model.Tags;
+
+public final class InstanceUtil {
```
VRohach, line 26: @RuslanLavrov Let's add documentation for this class for a better understanding of the main purpose of this entity.

src/main/java/org/folio/services/BulkProcessingContext.java
```diff
@@ -0,0 +1,41 @@
+package org.folio.services;
+
+import org.apache.commons.lang3.StringUtils;
+
+public class BulkProcessingContext {
```
VRohach, line 5: @RuslanLavrov Let's add documentation for this class for a better understanding of the main purpose of this entity.

src/main/java/org/folio/services/instance/InstanceS3Service.java
```diff
@@ -0,0 +1,246 @@
...
+
+      try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream))) {
+        return reader.lines()
+          .map(JsonObject::new)
+          .toList();
+      }
+    });
+  }
+
+  private Future<InstanceBulkResponse> upsert(List<JsonObject> instances, InstanceBulkRequest bulkRequest) {
```
VRohach, line 109: @RuslanLavrov  I think it would be better to make a different method name, to not overload them. For example, here rename to "upsertInstances". And the next method - "createInstanceAndUpdatePrecedingSucceedingTitles". Or smth like this. What do you think?

VRohach (approved):

KaterynaSenchenko (approved):

Aliaksandr-Fedasiuk (approved):

src/main/java/org/folio/rest/support/InstanceUtil.java
```diff
@@ -0,0 +1,269 @@
...
+
+  /**
+   * Maps instance json representation that corresponds to the mod-inventory module instance schema to instance object
+   * corresponding to the inventory storage module schema.
+   *
+   * @param instanceDtoJson - mod-inventory instance representation as json object
+   * @return {@link Instance} object
+   */
+  public static Instance mapInstanceDtoJsonToInstance(JsonObject instanceDtoJson) {
+    return new Instance()
```
psmagin, line 82: I'm very concerned about this. mod-inventory-storage must not know anything about mod-inventory schema.
What is the reason for that?

src/main/java/org/folio/services/instance/InstanceS3Service.java
```diff
@@ -0,0 +1,262 @@
...
+    log.debug("processInstances:: Processing bulk instances request, filename: '{}'", bulkRequest.getRecordsFileName());
+    return loadInstances(bulkRequest)
+      .compose(instances -> upsert(instances, bulkRequest))
+      .onFailure(e -> log.warn("processInstances:: Failed to process instances bulk request, filename: '{}'",
+        bulkRequest.getRecordsFileName(), e));
+  }
+
+  private Future<List<JsonObject>> loadInstances(InstanceBulkRequest bulkRequest) {
+    return vertx.executeBlocking(() -> {
+      FolioS3Client s3Client = folioS3ClientFactory.getFolioS3Client();
```
psmagin, line 103: There is a defined `folioS3Client` field in the class, that could be used there

psmagin (changes requested):

src/main/java/org/folio/rest/support/InstanceUtil.java
RuslanLavrov: understand your concern, this is necessary to convert the mapped instances retrieved from a file on S3 to the instance representation used by the mod-inventory-storage module. The mapping logic used in the mod-marc-migration to remap instances produces instances that correspond to the mod-inventory instance schema. In order to provide a clearer understanding about the incoming instance structure, the schema that the incoming instance representation corresponds to is described/mentioned here.
Do you prefer to describe it here smth like: "methods maps bulk instance record representation retrieved from external bulk instances file to inventory storage instance", or do you have another proposal?

src/main/java/org/folio/services/instance/InstanceS3Service.java
RuslanLavrov: rewritten

src/main/java/org/folio/rest/support/BulkProcessingErrorFileWriter.java
RuslanLavrov: added

src/main/java/org/folio/rest/support/InstanceUtil.java
RuslanLavrov: added, however, objectively the class semantic is obvious, and methods here had already been commented

src/main/java/org/folio/services/BulkProcessingContext.java
RuslanLavrov: added

src/main/java/org/folio/services/instance/InstanceS3Service.java
RuslanLavrov: changed

dmytrokrutii (approved):

src/main/java/org/folio/rest/support/InstanceUtil.java
psmagin: Hello @RuslanLavrov,

I understand. This class will likely be challenging to maintain. I recommend creating a schema for this instance representation that extends the current module instance schema. This would enable you to use JSON-to-object deserialization, which I believe would greatly streamline the process.

Additionally, looking ahead, you could consider refactoring to introduce a new endpoint that utilizes this extended schema. In mod-inventory, this would simplify operations by allowing you to call this new endpoint directly, eliminating the need for additional calls (for preceding/succeeding titles).

Thoughts?

RuslanLavrov: Hello @psmagin, could you elaborate on how adding a new schema would allow to use json-to-object deserialization to inventory-storage instance object? It’s not quite clear because this method maps instance json (that correspond to the mod-inventory schema) to the inventory-storage instance representation. Additionally, during this mapping some fields (e.g. preceding/succeeding titles) need to be ignored. Or do you mean that the method could transfer data to the existing Instance object from instance object representation generated by the new proposed schema instead of from a JsonObject?

src/main/java/org/folio/services/s3storage/FolioS3ClientFactory.java
```diff
@@ -0,0 +1,36 @@
...
+import org.folio.s3.client.S3ClientProperties;
+
+public class FolioS3ClientFactory {
+
+  public static final String AWS_URL_CONFIG = "AWS_URL";
+  public static final String AWS_REGION_CONFIG = "AWS_REGION";
+  public static final String AWS_BUCKET_CONFIG = "AWS_BUCKET";
+  public static final String AWS_ACCESS_KEY_ID_CONFIG = "AWS_ACCESS_KEY_ID";
+  public static final String AWS_SECRET_ACCESS_KEY_CONFIG = "AWS_SECRET_ACCESS_KEY";
+  public static final String AWS_SDK_CONFIG = "AWS_SDK";
```
psmagin, line 14: I suggest to use these var names: https://github.com/folio-org/mod-bulk-operations?tab=readme-ov-file#environment-variables
I believe that all new implementations with S3 use this naming.

RuslanLavrov: renamed variables

PR #1087, opened 2024-09-27
[MODINVSTOR-1264] Add codes to Subject sources

RomanChernetskyi:
### Purpose
Add codes to Subject sources
### Approach

- Extend schema of Subject source to contain code (code should be unique)
- Remove reference data “specified in subfiled $2”
- Add codes to existing as ref data Subject sources (find corresponding values in [Subject Heading and Term Source Codes: Source Codes for Vocabularies, Rules, and Schemes (Network Development and MARC Standards Office, Library of Congress)](https://www.loc.gov/standards/sourcelist/subject.html) )

### Jira 
https://folio-org.atlassian.net/browse/MODINVSTOR-1264

### DIFF
--- a/reference-data/subject-types/Topicalterm.json
+++ b/reference-data/subject-types/Topicalterm.json
@@ -1,0 +1,5 @@
+{
+  "id": "d6488f88-1e74-40ce-81b5-b19a928ff5b7",
+  "name": "Topical term",
+  "source": "folio"
+}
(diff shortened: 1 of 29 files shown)


psmagin (approved):
Please update NEWS

mukhiddin-yusuf (approved):

dmytrokrutii (approved):

reference-data/subject-types/Topicalterm.json
RuslanLavrov, line 1: should we fill in this file or we don't need it?

RomanChernetskyi: Filled file, thank you, regarding adding ref-data jsons: it is done to keep ref-data consistent and visible

KaterynaSenchenko (approved):

reference-data/subject-types/Topicalterm.json
RuslanLavrov: thank you. However, the corresponding sql scripts already ensure consistency and visibility of source and type ref-data as well. Just thought that mixing two approaches for ref-data creation could lead to confusion about how this data is being created. Maybe we should use only one of them to maintain clarity on how the subjects source and type ref-data is created, what do you think?

RomanChernetskyi: @psmagin confirmed that we use this approach for now to keep ref data visible

PR #1098, opened 2024-10-15
[MODINVSTOR-1270] Allow Subject sources and Subject type to have consortium source

RomanChernetskyi:
### Purpose
Allow Subject sources and Subject type to have consortium source

### Jira
https://folio-org.atlassian.net/browse/MODINVSTOR-1270

### DIFF
--- a/src/main/java/org/folio/rest/support/SubjectUtil.java
+++ b/src/main/java/org/folio/rest/support/SubjectUtil.java
@@ -1,0 +1,70 @@
+package org.folio.rest.support;
+
+import static io.vertx.core.Future.succeededFuture;
+import static org.folio.rest.jaxrs.resource.SubjectSources.PostSubjectSourcesResponse.respond422WithApplicationJson;
+import static org.folio.rest.support.ResponseUtil.SOURCE_CANNOT_BE_FOLIO;
+import static org.folio.rest.support.ResponseUtil.SOURCE_CANNOT_BE_UPDATED_AT_NON_ECS;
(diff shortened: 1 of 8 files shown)


src/main/java/org/folio/rest/support/SubjectUtil.java
```diff
@@ -0,0 +1,70 @@
...
+      return consortiumService.getConsortiumData(okapiHeaders)
+        .map(consortiumDataOptional -> {
+          if (consortiumDataOptional.isEmpty()) {
+            return getValidationErrorMessage(incomingSubjectSource, SOURCE_CANNOT_BE_UPDATED_AT_NON_ECS);
+          }
+          return Optional.empty();
+        });
+    }
+    return Future.succeededFuture(Optional.empty());
+  }
```
psmagin, line 61: Could it be updated from consortium to local, from local to folio?

psmagin (changes requested):
Please cover it with unit tests.

src/main/java/org/folio/rest/support/SubjectUtil.java
RomanChernetskyi: It could be updated from consortium to local and vice versa only at ECS setup, updating to source folio is not allow

psmagin (approved):

mukhiddin-yusuf (approved):

RuslanLavrov (approved):

KaterynaSenchenko (approved):

PR #1119, opened 2024-11-27
MODINVSTOR-1284 User can delete local Subject types/sources when it linked to Instance

JavokhirAbdullayev:
### Purpose
[MODINVSTOR-1284](https://folio-org.atlassian.net/browse/MODINVSTOR-1284) User can delete local Subject types/sources when it linked to Instance

### Approach
add validation to check if subject source/type linked with instance

### Changes Checklist

### Learning and Resources (if applicable)
TODO: In the scope of another ticket [MODINVSTOR-1299](https://folio-org.atlassian.net/browse/MODINVSTOR-1299), we will update the rollback mechanism, for instance, when linking with subjects fails. 

### Screenshots (if applicable)
image


src/main/java/org/folio/persist/InstanceRepository.java
```diff
@@ -40,6 +40,24 @@ public InstanceRepository(Context context, Map<String, String> okapiHeaders) {
...
+
+    return postgresClient.execute(sql.toString())
+      .map(rowSet -> {
+        if (rowSet != null && rowSet.iterator().hasNext()) {
+          Row row = rowSet.iterator().next();
+          return row.getBoolean("record_exists");
+        }
+        return Boolean.FALSE;
+      });
+  }
```
psmagin, line 58: This approach could be highly inefficient; in the worst case, it would require searching through all records to find one with the subject field, potentially impacting other processes related to instances and subject types.

I suggest creating an intermediate many-to-many relationship table with foreign keys (FKs) linked to both subject types and instances. This would facilitate reliance on database validation and eliminate the need to search the instance table.

psmagin (changes requested):

src/main/java/org/folio/persist/InstanceRepository.java
```diff
@@ -35,11 +36,61 @@ public class InstanceRepository extends AbstractRepository<Instance> {
...
   private static final String INVENTORY_VIEW_JSONB_FIELD = "inventory_view.jsonb";
+  private static final String INSTANCE_SUBJECT_SOURCE_TABLE = "instance_subject_source";
+  private static final String INSTANCE_SUBJECT_TYPE_TABLE = "instance_subject_type";
+
 
   public InstanceRepository(Context context, Map<String, String> okapiHeaders) {
     super(postgresClient(context, okapiHeaders), INSTANCE_TABLE, Instance.class);
   }
 
+  public Future<Boolean> doesInstanceExistBySubjectField(String field, String value) {
```
psmagin, line 47: If you have FKs, then there is no need for such check. You could just catch database exception.

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -115,6 +115,13 @@ public Future<Response> createInstance(Instance entity) {
           // api client invoking this endpoint. The response is returned
           // a little earlier so the api client can continue its processing
           // while the domain event publish is satisfied.
+          .compose(response -> {
+            if (response.getEntity() instanceof Instance instanceResponse) {
+              instanceRepository.linkInstanceWithSubjectSourceAndType(instanceResponse);
```
RuslanLavrov, line 120: could you say please, should we insert/delete data in the "instance_subject_source", "instance_subject_type" tables upon instance update as well? For example, if during instance update some of subject is added with source/type that has not been specified yet in these new tables, or when instance sent for update does not contain subject with type/source that is already specified in the new tables.

VRohach (approved):

src/main/java/org/folio/services/instance/InstanceService.java
```diff
@@ -150,14 +157,39 @@ public Future<Response> updateInstance(String id, Instance newInstance) {
...
+
+    if (newInstance.getSubjects().isEmpty()) {
+      instanceRepository.unlinkInstanceFromSubjectSource(instanceId);
+      instanceRepository.unlinkInstanceFromSubjectType(instanceId);
+      return;
+    }
+
+    for (var subject : newInstance.getSubjects()) {
+      if (subject.getSourceId() == null) {
+        instanceRepository.unlinkInstanceFromSubjectSource(instanceId);
```
psmagin, line 179: Here you unlink all subjects from the instance, but should unlink only one based on your logic.

psmagin (changes requested):

src/main/java/org/folio/services/instance/InstanceService.java
JavokhirAbdullayev: here, i'm iterating over subjects and unlink if some of them does not exist in newInstance json

src/main/java/org/folio/persist/InstanceRepository.java
```diff
@@ -35,11 +36,69 @@ public class InstanceRepository extends AbstractRepository<Instance> {
...
+        instanceId, sourceId);
+    postgresClient.execute(sql);
+  }
+
+  public void unlinkSubjectType(String instanceId, String typeId) {
+    var sql = "DELETE FROM " +
+      postgresClientFuturized.getFullTableName(INSTANCE_SUBJECT_TYPE_TABLE) +
+      String.format(" WHERE instance_id = '%s' AND type_id = '%s';", instanceId,
+        typeId);
+    postgresClient.execute(sql);
```
psmagin, line 98: Basicaly the logic should be like following:
1. Get all source/type ids from new instance subjects.
2. Get all source/type ids from old instance subjects.
3. Identify distinction: what should be created, what should be deleted.
4. Do queries for creating and deleting these relations in batch and not one by one.

JavokhirAbdullayev:
> How do you ensure consistency? Specifically, what happens if an instance is saved successfully but the related data fails to save? Do you have any rollback mechanisms for the instance? Are you considering the use of transactions?
For me, it feels a bit over-engineering, a small bug become a huge technical issue, I will do that if you want

JavokhirAbdullayev:
> How do you ensure consistency? Specifically, what happens if an instance is saved successfully but the related data fails to save? Do you have any rollback mechanisms for the instance? Are you considering the use of transactions?

JavokhirAbdullayev:
> How do you ensure consistency? Specifically, what happens if an instance is saved successfully but the related data fails to save? Do you have any rollback mechanisms for the instance? Are you considering the use of transactions?

Currently because of transactional behaviour of post Instance, most of the tests were failing because of DB Timeouts

KaterynaSenchenko (approved):

psmagin (approved):

PR #1128, opened 2024-12-20
MODINVSTOR-1343: Sort holdings by location name in instanceId query

julianladisch:
### Purpose
The UI has the instance view that shows the holdings that belong to that instance. There are many libraries with many holdings per instance, they need this sorted by effectiveLocation.name, callNumberPrefix, callNumber, callNumberSuffix.

RMB doesn’t support sorting by effectiveLocation.name because that’s a foreign table field. It only supports sorting by effectiveLocationId.

Implement this special case that the UI will use:

`instanceId==5b966ed6-3cd6-4dda-a0c7-f9b5e2826ccd sortBy effectiveLocation.name callNumberPrefix callNumber callNumberSuffix`

Allow any sub-set and order for the sortBy fields.

### Approach
Create a LEFT JOIN query.
This builds a pair of the holdings jsonb and the location name.
Put this into a WITH clause.
Then do the sorting.
Create a JSON list with the result.
Add a totalRecords value.

### Changes Checklist

### Related Issues
https://folio-org.atlassian.net/browse/MODINVSTOR-1343


src/main/java/org/folio/persist/HoldingsRepository.java
```diff
@@ -19,6 +21,53 @@ public HoldingsRepository(Context context, Map<String, String> okapiHeaders) {
...
+      }
+      orderBy.append(s);
+    }
+    var sql = "WITH data AS ("
+        + " SELECT h.jsonb AS jsonb, l.jsonb->>'name' AS name"
+        + " FROM " + HOLDINGS_RECORD_TABLE + " h"
+        + " LEFT JOIN " + LOCATION_TABLE + " l"
+        + "   ON h.effectiveLocationId = l.id"
+        + " WHERE h.instanceId=$1"
+        + " )"
```
mweaver-ebsco, line 55: Personal opinion: if this CTE is dealing with enough data to justify the `OFFSET` and `LIMIT`, then it should probably be set to not get materialized.

```suggestion
    var sql = "WITH data AS NOT MATERIALIZED ("
        + " SELECT h.jsonb AS jsonb, l.jsonb->>'name' AS name"
        + " FROM " + HOLDINGS_RECORD_TABLE + " h"
        + " LEFT JOIN " + LOCATION_TABLE + " l"
        + "   ON h.effectiveLocationId = l.id"
        + " WHERE h.instanceId=$1"
        + " )"
```

src/main/java/org/folio/persist/HoldingsRepository.java
```diff
@@ -19,6 +21,53 @@ public HoldingsRepository(Context context, Map<String, String> okapiHeaders) {
     super(postgresClient(context, okapiHeaders), HOLDINGS_RECORD_TABLE, HoldingsRecord.class);
   }
 
+  /**
+   * Row where the first value is an array with all holdings records; the second value is
+   * the exact totalRecords count.
+   */
```
mweaver-ebsco, line 27: Nit: The query results should be returned with usable names rather than having to be referred to by ordinal position. It's MUCH more usable and maintainable.

psmagin (approved):

viacheslavkol (approved):

PR #1141, opened 2025-01-24
MODINVSTOR-1354 Optimize check for existing shadow Instance before creating it

KaterynaSenchenko:
[MODINVSTOR-1354](https://folio-org.atlassian.net/browse/MODINVSTOR-1354)
In ECS when Holdings is created on a member tenant associated with a shared Instance that exists on a central tenant, a shadow copy of the Instance should be created on the member to link the Holdings, unless it already exists. To check if the Instance exists it is unnecessary to retrieve the whole object, the operation could be optimized by merely checking that object with such id exists.


JavokhirAbdullayev (approved):

psmagin (commented):
Please update NEWS.
Please update PR description with reasoning for the change.

psmagin (approved):

PR #1181, opened 2025-04-07
MODINVSTOR-1389 Fix error serialization when writing to files uploaded to S3

JavokhirAbdullayev:
### Purpose
[MODINVSTOR-1389](https://folio-org.atlassian.net/browse/MODINVSTOR-1389) Fix error serialization when writing to files uploaded to S3
 

### Approach
Fix the serialization of Errors when writing errors into a file from the bulk api

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,7 +12,7 @@
 
 ### Bug fixes
 * Fix ordering of electronic access items for inventory-hierarchy, oai-pmh-view ([MODINVSTOR-1224](https://folio-org.atlassian.net/browse/MODINVSTOR-1224))
-* Errors during save phase of marc-migrations (instances) ([MODINVSTOR-1389](https://folio-org.atlassian.net/browse/MODINVSTOR-1389))
+* Fix error serialization when writing to files uploaded to S3 ([MODINVSTOR-1389](https://folio-org.atlassian.net/browse/MODINVSTOR-1389))
 
 ### Tech Dept
 * Description ([ISSUE](https://folio-org.atlassian.net/browse/ISSUE))


src/main/java/org/folio/services/bulkprocessing/AbstractEntityS3Service.java
```diff
@@ -122,10 +123,13 @@ private Future<Void> handleUpsertFailure(AtomicInteger errorsCounter, BulkProces
...
-    log.warn("handleUpsertFailure:: Failed to process single entity upsert operation, entityId: '{}'",
-      entityId, e);
+    var errorDetails = (e instanceof ValidationException validationException)
+      ? validationException.getErrors().getErrors().getFirst().getMessage() :
+      e.getMessage();
+    log.warn("handleUpsertFailure:: Failed to process single entity upsert operation, entityId: '{}', error: {}",
+      entityId, errorDetails);
     errorsCounter.incrementAndGet();
-    return errorsWriter.write(entityToWrite, this::extractEntityId, e);
+    return errorsWriter.write(entityToWrite, this::extractEntityId, errorDetails);
```
mukhiddin-yusuf, line 132: maybe it makes sense to just pass the ```validationException.getErrors().getErrors().getFirst()``` instead of modifying the ```write(...)``` method and what if ```.getErrors()``` contains more errors - don't we need to aggregate them?

also other minor point: if it is validationException then would be nice to check ```validationException.getErrors() != null``` as I think it is not protected from being null

psmagin (dismissed):

psmagin (changes requested):
Please create mod-inventory-storage Jira ticket and use it in the PR and commit message for reference.
Update NEWS

NEWS.md
```diff
@@ -12,6 +12,7 @@
 
 ### Bug fixes
 * Fix ordering of electronic access items for inventory-hierarchy, oai-pmh-view ([MODINVSTOR-1224](https://folio-org.atlassian.net/browse/MODINVSTOR-1224))
+* Errors during save phase of marc-migrations (instances) ([MODINVSTOR-1389](https://folio-org.atlassian.net/browse/MODINVSTOR-1389))
```
psmagin, line 15: The mod-inventory-storage micro service doesn't know anything about marc migration. Please specify in news (and in following squash merge commit) what exactly changed in this PR (what endpoint logic was modified and how)

psmagin (changes requested):

psmagin:
Please update PR description and title to include the created jira ticket. Would also be appreciated if you will update the description of the ticket with details that make sense for the module

PR #1217, opened 2025-06-24
test: stabilize tests by fixing  cache issue during create/update holding

SvitlanaKovalova1:
### Purpose
[MODINVSTOR-1405](https://folio-org.atlassian.net/browse/MODINVSTOR-1405) Stabilize tests by fixing  cache issue during create/update holding

### Approach
Add `X-Okapi-Url` header with mock url to fix the cache issue `Failed to request method='GET' uri='/user-tenants?limit=1'`

### Changes Checklist
- [ ] **Unit Testing**: Confirm that changed classes were covered by unit tests.
- [ ] **Manual Testing**: Confirm that changes were tested on local or dev environment.


src/test/java/org/folio/rest/support/http/ResourceClient.java
```diff
@@ -1,5 +1,6 @@
 package org.folio.rest.support.http;
 
+import static org.folio.rest.api.TestBaseWithInventoryUtil.mockServer;
```
psmagin, line 3: I think it's a bad solution to use a variable that is initialized by Wiremock framework in the client, which could be used when Wiremock and TestBaseWithInventoryUtil are not used in the test.

I would suggest to set header somewhere in tests implicitly or create some special resource client for holdings that will override basic client methods with logic we need to add to it. 
Wiremock url could be passed in constructor for such client or it could be null in case there is no wiremock

PR #1224, opened 2025-06-27
feat(holding-validation) Holdings record statistical codes validation enhancement

SvitlanaKovalova1:
### Purpose
[MODINVSTOR-1185](https://folio-org.atlassian.net/browse/MODINVSTOR-1185) Holdings record statistical codes validation enhancement. 
A Holding Create request with an invalid `statisticalCodeIds` value returns a `500 Internal Server Error` with the message: `ERROR: invalid input syntax for type uuid: "incorrect value" (22P02)`. The returned status and message are incorrect. 
The expected response is `HTTP 400 Bad Request` with the message: `invalid input syntax for type uuid: "incorrect value"`.

### Approach
Updated failure response handler for Holding Create API when handling invalid fields.

### Changes Checklist
- [ ] **Unit Testing**: Confirm that changed classes were covered by unit tests.

### Screenshots
Before:
  <kbd>
  key</kbd>

After:
  <kbd>
  key</kbd>

### DIFF
--- a/src/main/java/org/folio/persist/AbstractRepository.java
+++ b/src/main/java/org/folio/persist/AbstractRepository.java
@@ -4,6 +4,7 @@ import static io.vertx.core.Future.succeededFuture;
 
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Future;
+import io.vertx.core.Handler;
 import io.vertx.sqlclient.Row;
 import io.vertx.sqlclient.RowSet;
 import io.vertx.sqlclient.Tuple;
@@ -36,6 +37,10 @@ public abstract class AbstractRepository<T> {
 
   public Future<String> save(String id, T entity) {
     return postgresClientFuturized.save(tableName, id, entity);
+  }
+
+  public void save(String id, T entity, Handler<AsyncResult<T>> replyHandler) {
+    postgresClientFuturized.save(tableName, id, entity, replyHandler);
   }
 
   public Future<List<T>> get(Criterion criterion) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/folio/persist/AbstractRepository.java
```diff
@@ -39,8 +38,8 @@ public Future<String> save(String id, T entity) {
     return postgresClient.save(tableName, id, entity);
   }
 
-  public void save(String id, T entity, Handler<AsyncResult<T>> replyHandler) {
-    postgresClientFuturized.save(tableName, id, entity, replyHandler);
+  public Future<T> save(String table, String id, T entity) {
```
psmagin, line 41: There is a `tableName` field in the class, so we don't need it in the method args.

src/main/java/org/folio/services/holding/HoldingsService.java
```diff
@@ -146,17 +160,62 @@ public Future<Response> createHolding(HoldingsRecord entity) {
...
+                postResponse.handle(Future.succeededFuture(PostHoldingsStorageHoldingsResponse
+                  .respond201WithApplicationJson(reply.result(),
+                    PostHoldingsStorageHoldingsResponse.headersFor201().withLocation(LOCATION_PREFIX))));
+              } else {
+                handleFailedResponse(hr.getId(), reply.cause(), postResponse);
+              }
+            } catch (Exception e) {
+              internalServerErrorDuringPost(e, postResponse);
+            }
+          });
```
psmagin, line 181: I think it would be better to perform validation before sending the record to the database, instead of duplicating code from PgUtil.

SvitlanaKovalova1: updated the code

psmagin (commented):
Have you checked instances and items, they also have same field

psmagin:
By the way you could use UuidUtil.isUuid method to validate

SvitlanaKovalova1:
> Have you checked instances and items, they also have same field

The same validation is added for Instance/Item records

psmagin (approved):

PR #1284, opened 2025-12-09
Modinvstor 1496

mhamann-ubl:
### Purpose
This PR makes use of a dedicated additionalCallNumberComponent, which helps differentiate its properties from those of the EffectiveCallNumberComponents. This helps solving an issue with naming in the version history in inventory, where callNumber, effectiveCallNumber and additionalCallNumber properties naming leads to the user being unable to differentiate between them.

### Approach
This change renames the additionalCallNumber components by prefixing them with "additionalCallNumber".

### Changes Checklist
- [ ] **Manual Testing**: Confirm that changes were tested on local or dev environment.
- [ ] **NEWS**: Confirm that the NEWS file is updated with relevant information about the changes made in this pull request.

### Related Issues
https://folio-org.atlassian.net/browse/UIIN-3548

### DIFF
--- a/ramls/common/additionalCallNumberComponents.json
+++ b/ramls/common/additionalCallNumberComponents.json
@@ -3,27 +3,27 @@
   "description": "A call number",
   "type": "object",
   "properties": {
-    "typeId": {
+    "additionalCallNumberTypeId": {
       "type": "string",
       "description": "unique ID for the type of call number on a holdings record, a UUID",
       "$ref": "../uuid.json"
     },
-    "prefix": {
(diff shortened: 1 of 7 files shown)


psmagin (changes requested):
I don’t think we should rename it. If there is a collision in the UI, then the UI should be adjusted. The audit response contains enough information to determine whether the call number components are from the additional call numbers list, the effective call number, or the item-level call number.

elsenhans:
Naming issue will be solved in UI:
https://github.com/folio-org/ui-inventory/pull/2951

PR #1309, opened 2026-02-16
MODINVSTOR-1520 Add indexes to the instance table

mweaver-ebsco:
### Purpose
This PR adds indexes for several instance fields. This will enable FQM to efficiently run queries based on these fields. They were identified as commonly used fields in FQM, but none of them have (useful) indexes available right now.

### Approach
FQM works by accessing live data via DB views. This means that indexes in the source schema can be used directly by FQM to improve its query performance.

### Changes Checklist
- [X] **Database Schema Changes**: This adds several DB functions and indexes via SQL scripts. In an upgrade, sysops will probably want to manually run these scripts. If they don't, then the DB objects will still be created, but the upgrade process will hang for a potentially very long time as the indexes are created.
- [X] **Manual Testing**: This has been tested in the Corsair eperf rancher environment. I'll be doing further testing before merging, too.
- [X] **NEWS**: Confirm that the NEWS file is updated with relevant information about the changes made in this pull request.

### Related Issues
MODINVSTOR-1520
MODFQMMGR-1056

### Learning and Resources (if applicable)
I noticed that the `instance_identifiers_idx_ft` is not particularly useful: it indexes a significant amount of unneeded data, which makes the index significantly larger and slower than it should be, and it's also tricky to actually use. Unless there's something that actually uses it, that index should probably be dropped. The alternative version added in this PR performs significantly better.

### DIFF
--- a/mod-inventory-storage-server/src/main/resources/templates/db_scripts/createToTimestampFunction.sql
+++ b/mod-inventory-storage-server/src/main/resources/templates/db_scripts/createToTimestampFunction.sql
@@ -1,0 +1,12 @@
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.to_timestamp(text)
+  RETURNS timestamp AS $$
+BEGIN
+  -- We assume the input text is always in a format (like ISO 8601)
+  -- that casts consistently to timestamp. This function turns the
+  -- ::timestamp cast into an IMMUTABLE function, so that it can be
+  -- used in indexes
+  RETURN $1::timestamp;
+EXCEPTION WHEN others THEN
+  RETURN NULL;
+END;
+$$ LANGUAGE plpgsql IMMUTABLE;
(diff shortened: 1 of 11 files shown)


mod-inventory-storage-server/src/main/resources/templates/db_scripts/createToTimestampFunction.sql
```diff
@@ -0,0 +1,12 @@
+CREATE OR REPLACE FUNCTION ${myuniversity}_${mymodule}.to_timestamp(text)
```
mweaver-ebsco: I kind of hate this, but you can't use `::timestamp` directly in an index because that's not considered an immutable function (it depends on server settings). By treating the cataloged date as a date instead of text, the index is faster, smaller, and supports more query operators.

bvsharp (approved):

psmagin (approved):

PR #1316, opened 2026-02-24
MODINVSTOR-1528 Add indexes for instance note fields

mweaver-ebsco:
Similar to #1309, this commit adds new indexes for the sub-fields within instance notes and for instance administrative notes. These indexes will be used for querying by FQM (MODFQMMGR-1057).

### Purpose
This PR adds indexes for several instance fields. This will enable FQM to efficiently run queries based on these fields. They were identified as commonly used fields in FQM, but none of them have (useful) indexes available right now.

### Approach
FQM works by accessing live data via DB views. This means that indexes in the source schema can be used directly by FQM to improve its query performance.

### Changes Checklist
- [X] **Database Schema Changes**: This adds several DB functions and indexes via SQL scripts. In an upgrade, sysops will probably want to manually run these scripts. If they don't, then the DB objects will still be created, but the upgrade process will hang for a potentially very long time as the indexes are created.
- [X] **Manual Testing**: This has been tested in the Corsair eperf rancher environment. I'll be doing further testing before merging, too.
- [X] **NEWS**: Confirm that the NEWS file is updated with relevant information about the changes made in this pull request.

### Related Issues
MODINVSTOR-1528
MODFQMMGR-1057

### DIFF
--- a/mod-inventory-storage-server/src/main/resources/templates/db_scripts/instance/createInstanceNotesStaffOnlyIndex.sql
+++ b/mod-inventory-storage-server/src/main/resources/templates/db_scripts/instance/createInstanceNotesStaffOnlyIndex.sql
@@ -1,0 +1,6 @@
+CREATE INDEX IF NOT EXISTS instance_notes_staffonly_idx ON ${myuniversity}_${mymodule}.instance
+  USING gin (
+             ${myuniversity}_${mymodule}.normalize_jsonb_array(
+               ${myuniversity}_${mymodule}.extract_values(jsonb, 'notes', 'staffOnly'))
+             jsonb_path_ops
+    );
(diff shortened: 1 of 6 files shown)


bvsharp (approved):

psmagin (commented):
Hey @mweaver-ebsco , How many more indexes do you plan to create? I understand that this improves fetch performance, but it also decreases insert/update performance.

mod-inventory-storage-server/src/main/resources/templates/db_scripts/instance/createInstanceNotesStaffOnlyIndex.sql
```diff
@@ -0,0 +1,6 @@
+CREATE INDEX IF NOT EXISTS instance_notes_staffonly_idx ON ${myuniversity}_${mymodule}.instance
+  USING gin (
+             ${myuniversity}_${mymodule}.normalize_jsonb_array(
```
mweaver-ebsco: Technically, this doesn't need to be normalized, since this is actually stored as a true `boolean` in the JSONB, but it ends up being easier to work with as a normalized string, and there's basically no overhead in the index

psmagin (approved):

PR #1378, opened 2026-07-03
feat(inventory): add feature flag to prevent redundant updates

SvitlanaKovalova1:
### Purpose
Create feature flag for enabling optimization to prevent redundant updates in Inventory 

#### Database Schema
Created a new `settings` table with the following columns:

| Column | Type | Description |
|--------|------|-------------|
| **id** | UUID | Primary key, auto-generated unique identifier for each setting record |
| **key** | VARCHAR(255) | Unique key identifying the setting (e.g., `inventory.optimize-updates.enabled`). Enforced by `unique_settings_key` constraint |
| **value** | TEXT | Current value of the setting, stored as a string regardless of the logical type |
| **type** | VARCHAR(50) | Data type of the setting value: `BOOLEAN`, `INTEGER`, or `STRING`. Used for validation on updates |
| **central_managed** | BOOLEAN | Indicates if the setting is centrally managed in ECS environments. When `true`, only the central tenant can update the value; member tenants receive propagated updates |
| **description** | TEXT | Human-readable description explaining the setting's purpose and effect |
| **created_date** | TIMESTAMP | Timestamp when the setting record was created |
| **created_by_user_id** | UUID | ID of the user who created the setting |
| **updated_date** | TIMESTAMP | Timestamp when the setting was last updated |
| **updated_by_user_id** | UUID | ID of the user who last updated the setting |

The table is seeded with the **`inventory.optimize-updates.enabled`** setting (key) — a feature flag that controls whether the module skips redundant update operations for Instances, Holdings, and Items when the incoming record is identical to the stored one (ignoring metadata). 
Default `value: false`, `type: BOOLEAN`, `central_managed: true`.
image
#### Caching
Added a `SettingCache` — an async cache (keyed `tenantId:key`) for setting values to avoid repeated DB lookups . The cache is initialized in `InitApiImpl` and stored on the Vert.x context, with a configurable expiration time (default: 24 hours). Whenever the feature flag is updated via the PATCH endpoint, the corresponding cache entry is also updated with the new value.

#### API Endpoints
Added new Settings Storage API endpoints:
- **`GET /settings/{key}`** — Retrieve a setting by key (available for both non-ECS and ECS tenants)
- **`PATCH /settings/{key}`** — Update a setting value by key (available for non-ECS tenants and ECS central tenants only; member tenants in ECS receive automatic propagation when the central tenant updates a setting)

### Changes Checklist

### Related Issues
[MODINVSTOR-1577](https://folio-org.atlassian.net/browse/MODINVSTOR-1577)

### Screenshots (if applicable)

<table>
 <tr>
<td>
PATCH  {{host}}/settings/inventory.optimize-updates.enabled
 </td>
  </tr>
  <tr>
    <td>
      image
    </td>
  </tr>
</table>

***

<table>
 <tr>
<td>
GET {{host}}/settings/inventory.optimize-updates.enabled
 </td>
  </tr>
  <tr>
    <td>
image
    </td>
  </tr>
</table>

***

Validation:
<table>
  <tr>
    <td>
image
    </td>
  </tr>
  <tr>
    <td>
image
    </td>
  </tr>
  <tr>
    <td>
image
    </td>
  </tr>
</table>


descriptors/ModuleDescriptor-template.json
```diff
@@ -1436,6 +1436,22 @@
           "permissionsRequired": ["inventory-storage.migration.job.item.delete"]
         }
       ]
+    },
+    {
+      "id": "settings",
```
psmagin:
```suggestion
      "id": "inventory-settings",
```
Just 'settings' is to general. It could overlap with other modules.

mod-inventory-storage-server/src/main/java/org/folio/services/holding/HoldingsService.java
```diff
@@ -286,7 +290,8 @@ private Future<Response> updateHolding(HoldingsRecord oldHoldings, HoldingsRecor
   private Future<Response> checkAndPerformUpdate(HoldingsRecord oldHoldings, HoldingsRecord newHoldings) {
     try {
       var noChanges = equalsIgnoringMetadata(oldHoldings, newHoldings);
-      if (noChanges) {
+      var isOptimizeUpdatesEnabled = settingsService.isOptimizeUpdatesEnabled(okapiHeaders.get(TENANT));
+      if (isOptimizeUpdatesEnabled && noChanges) {
```
psmagin:
```suggestion
      var isOptimizeUpdatesEnabled = settingsService.isOptimizeUpdatesEnabled(okapiHeaders.get(TENANT));
      if (isOptimizeUpdatesEnabled && equalsIgnoringMetadata(oldHoldings, newHoldings)) {
```
To calculate `noChanges` only when seeting is enabled.

mod-inventory-storage-server/src/main/java/org/folio/services/setting/SettingsService.java
```diff
@@ -0,0 +1,175 @@
...
+        logger.debug("Setting updated: {} with value: {}", key, value);
+      }).mapEmpty();
+  }
+
+  public boolean isOptimizeUpdatesEnabled(String tenantId) {
+    return Boolean.parseBoolean(
+      getCachedSettingValue(tenantId, INVENTORY_OPTIMIZE_UPDATES_ENABLED.getValue())
+        .result()
+    );
+  }
```
psmagin: It should return Future. Getting result in this way is against Vert.X best practice.

mod-inventory-storage-server/src/main/java/org/folio/services/setting/SettingsService.java
```diff
@@ -0,0 +1,175 @@
...
+      );
+  }
+
+  private Future<Void> updateSettingAcrossConsortium(String key, Object value, Map<String, String> okapiHeaders,
+                                                     Setting entity, List<String> memberTenants) {
+    return updateSettingByKey(key, value, entity, okapiHeaders)
+      .compose(updatedSetting -> {
+        var cacheKey = okapiHeaders.get(TENANT) + ":" + key;
+        cache.put(cacheKey, CompletableFuture.completedFuture(value.toString()));
+        ArrayList<Future<Void>> updateFutures = new ArrayList<>();
```
psmagin:
```suggestion
        var updateFutures = new ArrayList<Future<Void>>();
```

mod-inventory-storage-server/src/main/java/org/folio/services/setting/SettingsService.java
```diff
@@ -0,0 +1,175 @@
...
+          .onFailure(t -> logger.error("Error updating setting across consortium for key {}", key, t))
+          .mapEmpty();
+      });
+  }
+
+  private void buildMemberTenantUpdateFutures(String key, Map<String, String> okapiHeaders, Setting updatedSetting,
+                                              List<String> memberTenants, List<Future<Void>> futures) {
+    var value = updatedSetting.getValue();
+    for (String memberTenantId : memberTenants) {
+      HashMap<String, String> headers = new HashMap<>(okapiHeaders);
```
psmagin:
```suggestion
     var  headers = new HashMap<>(okapiHeaders);
```

mod-inventory-storage-server/src/main/java/org/folio/services/setting/SettingsService.java
psmagin: public method should be before private ones.

mod-inventory-storage-server/src/main/java/org/folio/services/setting/SettingsService.java
```diff
@@ -0,0 +1,175 @@
...
+        if (entity == null) {
+          logger.warn("Setting not found: {}", key);
+          return Future.failedFuture(new NotFoundException("Setting not found: " + key));
+        }
+        logger.info("Setting found: {}, value: {}", key, entity.getValue());
+        return Future.succeededFuture(entity.getValue());
+      });
+  }
+
+  private Future<Setting> updateSettingByKey(String key, Object value, Setting entity,
```
psmagin:
```suggestion
  private Future<Setting> updateSettingByKey(String key, Object value, Setting existingSetting,
```
To make it clear that we passing existing setting

psmagin: Also, a key param probably reduntandt as we could get it from `Setting` object

mod-inventory-storage-server/src/main/java/org/folio/validator/SettingsValidator.java
```diff
@@ -0,0 +1,34 @@
...
+      return;
+    }
+    throw new SettingsValidationException("Setting value should be a boolean");
+  }
+
+  private <T> void validateType(Object value, Class<T> type, String errorMessage) {
+    if (!type.isInstance(value)) {
+      throw new SettingsValidationException(errorMessage);
+    }
+  }
```
psmagin: I would prefer to have one method in a way:

```suggestion
  private void validateBoolean(Object value) {
    if (value instanceof Boolean) {
      return;
    }
    if (value instanceof String string
      && ("true".equalsIgnoreCase(string) || "false".equalsIgnoreCase(string))) {
      return;
    }
    throw new SettingsValidationException("Setting value should be a boolean");
  }

  private <T> void validateType(Object value, Class<T> type, String errorMessage) {
  switch (type) {
       case Boolean.class -> validateBoolean(value);
       default -> {
           if (!type.isInstance(value)) {
            throw new SettingsValidationException(errorMessage);
          }
       }
    }
  }
```

psmagin (changes requested):

viacheslavkol (approved):

psmagin (commented):
How the cache will evict on another instances of the module?

SvitlanaKovalova1:
> How the cache will evict on another instances of the module?

Currently, caches in different instances are independent and only evict entries based on the configured expiration period, which is 24 hours by default. So yes, that behavior is not correct for our use case because other instances may continue serving stale feature flag values until the cached entry expires.
To address this, I added a publisher for setting update events and a consumer in each module instance. When a setting is updated, an event is published, and all instances receive it and immediately update the corresponding cache entry by key.

psmagin (approved):

PR #1405, opened 2026-08-28
MODINVSTOR-1600: Add instance custom link setting

zepheiryan:
### Purpose

Add a custom link setting intended for use and display with instance actions.

### Approach

- Add RAML and supporting schemas for `/instance-custom-links`
- Update database schema.json
- Implement API with additional field validation beyond required and maximum field size
- Add repository with ability to limit the number of custom links per tenant to 10

### Changes Checklist

### Related Issues

https://folio-org.atlassian.net/browse/MODINVSTOR-1600

### DIFF
--- a/mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
+++ b/mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
@@ -4,9 +4,6 @@ import io.vertx.core.AsyncResult;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.Handler;
-import jakarta.validation.constraints.Max;
-import jakarta.validation.constraints.Min;
-import jakarta.validation.constraints.Pattern;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
@@ -34,9 +31,8 @@ public class InstanceCustomLinkApi extends BaseApi<InstanceCustomLink, InstanceC
 
   @Validate
   @Override
-  public void getInstanceCustomLinks(String query, @Pattern(regexp = "exact|estimated|none|auto") String totalRecords,
-      @Min(0) @Max(2147483647) int offset, @Min(0) @Max(2147483647) int limit, Map<String, String> okapiHeaders,
-      Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
+  public void getInstanceCustomLinks(String query, String totalRecords, int offset, int limit, Map<String,
+      String> okapiHeaders, Handler<AsyncResult<Response>> asyncResultHandler, Context vertxContext) {
     getEntities(query, totalRecords, offset, limit, okapiHeaders, asyncResultHandler, vertxContext,
       GetInstanceCustomLinksResponse.class);
   }
--- a/mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
+++ b/mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
@@ -28,7 +28,7 @@ public class InstanceCustomLinkApi extends BaseApi<InstanceCustomLink, InstanceC
 
   public static final String INSTANCE_CUSTOM_LINK_TYPE_TABLE = "instance_custom_link";
 
-  private static final String FIELD_BASE_URL = "baseUrl";
+  private static final String FIELD_LINK = "link";
   private static final List<String> QUERY_TOKENS = List.of("{{UUID}}", "{{HRID}}", "{{indexTitle}}");
 
   @Validate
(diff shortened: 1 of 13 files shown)


mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
```diff
@@ -0,0 +1,172 @@
...
+
+  public static final String INSTANCE_CUSTOM_LINK_TYPE_TABLE = "instance_custom_link";
+
+  private static final String FIELD_BASE_URL = "baseUrl";
+  private static final List<String> QUERY_TOKENS = List.of("{{UUID}}", "{{HRID}}", "{{indexTitle}}");
+
+  @Validate
+  @Override
+  public void getInstanceCustomLinks(String query, @Pattern(regexp = "exact|estimated|none|auto") String totalRecords,
+      @Min(0) @Max(2147483647) int offset, @Min(0) @Max(2147483647) int limit, Map<String, String> okapiHeaders,
```
psmagin: I believe that these annotation are part of `InstanceCustomLinks`, so there is no need to duplicate them here.

mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
```diff
@@ -0,0 +1,172 @@
...
+    errors.addAll(validateQueryString(entity));
+    errors.addAll(validateBaseUrl(entity));
+
+    return errors;
+  }
+
+  private List<Error> validateQueryString(InstanceCustomLink entity) {
+    List<Error> errors = new ArrayList<>();
+
+    if (entity.getQueryString() != null) {
```
psmagin: Shouldn't the query string be required?

mod-inventory-storage-server/src/main/resources/templates/db_scripts/schema.json
```diff
@@ -228,6 +228,25 @@
...
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "linkText",
+          "tOps": "ADD"
+        },
+        {
+          "fieldName": "baseUrl",
+          "tOps": "ADD"
+        }
```
psmagin, line 247: Maybe we need to have unique combination of baseUrl + queryString instead of just baseUrl. Should be validated with business requirements.

mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
psmagin: Does it make sense to validate query string to start with one of `/` or `?` ?

psmagin (changes requested):
ModuleDescriptor must be updated  with new API.

mod-inventory-storage-server/src/main/java/org/folio/rest/impl/InstanceCustomLinkApi.java
zepheiryan: It's called out as not required, I assumed it was to allow for links without a connection to the instance; I'll check and make sure. That might explain why the base URL and query string are separate as well - the query string needs to contain a token only if provided, but a base URL does not require further processing or validation to interpolate. Otherwise we would combine them into one field.

A user could terminate the base URL with `/` so that requiring the query string to start with it would be redundant, or they may have included `/query?somethingfixed=1&id=` in the base URL. Without being very strict about what's allowed where, I think It's probably best to simply concatenate them after interpolating without additional rules or processing. It will be more straightforward to diagnose for users if they initially get the split wrong.

psmagin (commented):
What about length limitation for name, baseUrl, queryString? I think it's pretty important to have at least some limitation, like 1000 or so, because in other case it could contain a huge texts.

zepheiryan:
> What about length limitation for name, baseUrl, queryString? I think it's pretty important to have at least some limitation, like 1000 or so, because in other case it could contain a huge texts.

There were no business requirements for it, and I also didn't see any similar limitations in other schemas, but I'll add a high limit here.

psmagin (approved):

zepheiryan:
After working a bit on the front end, I found I needed to be able to cleanly associate the uniqueness errors with the fields they occurred on, and the built-in errors based on database constraints were obscured behind methods and JSON notation.

mod-inventory-storage-server/src/main/resources/templates/db_scripts/schema.json
psmagin: @zepheiryan, what do you think about this?

zepheiryan: After some further discussion with POs, we'll combine these two fields into one and update the validation to fit the combination - a unique field requiring a URI protocol and optionally one or more of the three tokens. Any other use of `{{` / `}}` delimited substrings will prompt an error - users who for some reason really need curly braces outside of this should URL encode them instead.

psmagin (approved):

Aliaksandr-Fedasiuk (approved):

PR #1415, opened 2026-09-18
MODINREACH-607 update inventory-view

wafschneider:
**Note that these changes are required for mod-inn-reach [PR#536](https://github.com/folio-org/mod-inn-reach/pull/536)**

### Purpose
mod-innreach uses the `inventory-view` API to search for instances by instance HRID. This API provides a CQL search of the `instance-holdings-item-view` view in the tenant's mod-inventory-storage schema.

The view is constructed mostly by creating a JSONB column on the fly using `JSONB_BUILD_OBJECT`. The only value available for searching outside the JSONB column is the instance `id`. That means that a search for the instance HRID cannot use existing index for HRID on the `instance` table, and so must perform a scan of the constructed JSONB column to satisfy a search for HRID. With a large collection (11 million instances), this results in very high CPU usage on the database and effectively takes down the system if performed in batch, and even a single query may time out before results are returned. Note that this query is initiated by the INN-Reach API server via edge-inn-reach/mod-inn-reach and so the library has little control over the load placed on the system.

### Approach
This patch adds a column to the `instance-holdings-item-view` view with the instance HRID as it is indexed in the `instance` table to allow for search supported by a database index. It also adds a special code path to support an exact match search against the `instanceHrid` CQL index that will use the `instance_hrid` column added to the view.

### Changes Checklist
- [x] **API Changes**: A CQL index `instanceHrid` is added to the `inventory-view` interface implementation, requiring an interface version bump. The shape of the response is not changed, and the change is backwards compatible (all previously supported queries remain unchanged).
- [x] **Database Schema Changes**: The `instance-holdings-item-view` definition is updated, with a change to the view creation script 
- [x] **Interface Version Changes**: `inventory-view` interface bumped to v.3.3. This is not ideal, but it appears that the interface version was bumped to v3.2 as part of the Trillium release, although there were not changes to the interface.
- [ ] **Interface Dependencies**: No changes to dependencies.
- [ ] **Permissions**: No changes to permissions.
- [ ] **Logging**: No changes to logging.
- [X] **Unit Testing**: Tests added to exercise new CQL index.
- [x] **Manual Testing**: Change has been tested locally in combination with update to mod-inn-reach.

### Related Issues
[MODINREACH-607](https://folio-org.atlassian.net/browse/MODINREACH-607)


wafschneider:
Closing this PR to open new one on local branch

